PR #6, opened 2024-02-27
Fix position of @main annotation

rudro:
- Also update package versions otherwise things don't compile out of the box (you need the version where `callAsFunction` is marked `open`)

### DIFF
--- a/mlx-swift-examples.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/mlx-swift-examples.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -15,7 +15,7 @@
       "location" : "https://github.com/ml-explore/mlx-swift",
       "state" : {
         "branch" : "main",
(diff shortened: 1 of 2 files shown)


mlx-swift-examples.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -15,7 +15,7 @@
       "location" : "https://github.com/ml-explore/mlx-swift",
       "state" : {
         "branch" : "main",
-        "revision" : "cadf5f8187ac0894e66cd288217e2eda9f2c933d"
+        "revision" : "fbe215ae29ec286bdc66f4b3423e3eea1b3ef2fc"
```
davidkoski, line 18: Aha! I wonder if this file should be checked in?  I am not sure what the best practice is here, but I agree this is needed!

davidkoski (approved):
Looks good, thank you!

PR #12, opened 2024-03-04
LeNet on MNIST + readme update

awni:


Libraries/MNIST/MNIST.swift
```diff
@@ -6,36 +6,43 @@ import MLXNN
...
+        fc3 = Linear(84, 10)
     }
 
     public func callAsFunction(_ x: MLXArray) -> MLXArray {
         var x = x
-        for l in layers.dropLast() {
-            x = relu(l(x))
-        }
-        return layers.last!(x)
+        x = pool1(relu(conv1(x)))
```
andresy, line 31: Note that at the time, tanh() was used

andresy (approved):

Applications/MNISTTrainer/ContentView.swift
```diff
@@ -13,7 +13,7 @@ struct ContentView: View {
     @State var trainer = Trainer()
 
     // toggle for cpu/gpu training
-    @State var cpu = true
+    @State var cpu = false
```
davidkoski, line 16: Discussing out-of-band just removing the toggle as it is no longer interesting.

davidkoski (approved):

PR #13, opened 2024-03-04
use memory limit API

davidkoski:


Applications/LLMEval/ContentView.swift
```diff
@@ -73,13 +73,18 @@ class LLMEvaluator {
     func load() async throws -> (LLMModel, LLM.Tokenizer) {
         switch loadState {
         case .idle:
+            // limit the buffer cache
+            MLX.GPU.cacheLimit = 1024 * 200
```
awni, line 77: Where did that number come from? It is pretty small, no?

awni: Also is that the function call itself? I'm curious, why did you do it this way vs a function?

awni:
Can you set the overall memory limit to something slightly less than the allowed limit the app? That might prevent it from crashing in some cases when the GPU runs ahead.

Applications/LLMEval/ContentView.swift
davidkoski: yeah, I just made up a number.  you are right, that is probably smaller than it should be.  maybe 20 megabytes?

davidkoski: it could be done as a function, but since you can read the value a property is more typical.

https://github.com/ml-explore/mlx-swift/pull/38/files#diff-24bacc036f57c2e9a6befcaf54dd36452eb6b5628180c2070ef9e286107744eaR52

awni: I think for the KV cache a couple hundred MB is not unreasonable. The KV cache can get pretty big. This is how I was computing the size in Python. 

- Do a forward to get the size of the kv cache per token (without an eval)
- Add a tiny buffer for the model query state

```
        _, cache = model(mx.array([[0]]))
        cache_size = max_tokens * sum(c.nbytes for _, c in tree_flatten(cache))
        cache_size += 10 * 2**20  # buffer
        mx.metal.set_cache_limit(cache_size)
```

On the other hand a simpler alternative for you may be to simply set the max memory limit to the limit allowed by the app and forget about controlling the cache. The max memory limit will let the cache grow until overall usage hits the limit.

davidkoski:
> Can you set the overall memory limit to something slightly less than the allowed limit the app? That might prevent it from crashing in some cases when the GPU runs ahead.

No, not really.  The process can't see its own jetsam limit.  I think it is easier to control the buffer cache -- that is the variable part of this.  The weights are fixed.

awni (approved):
LGTM 🚀

Applications/LLMEval/ContentView.swift
maiqingqiang: > yeah, I just made up a number. you are right, that is probably smaller than it should be. maybe 20 megabytes?

How is this 20MB calculated? Is it based on the MacBook's RAM?

awni: This number is more intended for iOS where memory is very limited. On the laptop you can use a much bigger number. If you generate text multiple times in the same program I would compute it based on the maximum number of tokens allowed per generation and the size of the KV cache. Probably that works out to a couple hundred megabytes.

You can always make it smaller though, the model will run slightly slower.

maiqingqiang: > This number is more intended for iOS where memory is very limited. On the laptop you can use a much bigger number. If you generate text multiple times in the same program I would compute it based on the maximum number of tokens allowed per generation and the size of the KV cache. Probably that works out to a couple hundred megabytes.
> 
> You can always make it smaller though, the model will run slightly slower.

If my laptop has 8GB of RAM, can it run a model as large as 14B or even larger with this setting?

awni: A quantized 14B model uses ~7 GB. That will probably run but very slow as it will likely swap, given you need some space for all the other apps and OS. The largest I've tried on an 8GB machine is a quantize 7B model which runs pretty well.

PR #23, opened 2024-03-12
adopt MLXFast.scaledDotProductAttention

davidkoski:
Same as https://github.com/ml-explore/mlx-examples/pull/515/files

### DIFF
--- a/Libraries/LLM/Gemma.swift
+++ b/Libraries/LLM/Gemma.swift
@@ -2,6 +2,7 @@
 
 import Foundation
 import MLX
+import MLXFast
 import MLXNN
 
 // port of https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/models/gemma.py
@@ -31,7 +32,6 @@ private class RMSNorm: Module, UnaryLayer {
 private class Attention: Module {
 
     let args: GemmaConfiguration
-    let repeats: Int
     let scale: Float
 
     @ModuleInfo(key: "q_proj") var wq: Linear
@@ -47,8 +47,6 @@ private class Attention: Module {
         let dim = args.hiddenSize
         let heads = args.attentionHeads
         let kvHeads = args.kvHeads
-
-        self.repeats = heads / kvHeads
 
         let headDim = args.headDimensions
         self.scale = pow(Float(headDim), -0.5)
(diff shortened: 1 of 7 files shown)


Libraries/LLM/Gemma.swift
```diff
@@ -91,14 +84,11 @@ private class Attention: Module {
...
-        }
-
-        scores = softMax(scores.asType(.float32), axis: -1).asType(scores.dtype)
-
-        let output = matmul(scores, values).transposed(0, 2, 1, 3).reshaped(B, L, -1)
+        let output = MLXFast.scaledDotProductAttention(
+            queries: queries, keys: keys, values: values, scale: scale, mask: mask
+        )
+        .transposed(0, 2, 1, 3)
+        .reshaped(B, L, -1)
```
davidkoski, line 91: The pattern for `MLXFast.scaledDotProductAttention()`

awni (approved):
Looks great 🚀 !

awni:
Btw am I missing it or are the tests not run for this PR?

davidkoski:
> Btw am I missing it or are the tests not run for this PR?

You are right -- I think I need to adjust the pattern for branch names.

davidkoski:
Fixed the test run

Libraries/LLM/Phi.swift
```diff
@@ -84,15 +78,13 @@ private class PhiAttention: Module {
...
-            scores = scores + mask
-        }
-
-        scores = softMax(scores, axis: -1).asType(values.dtype)
-        let valuesHat = matmul(scores, values).transposed(0, 2, 1, 3).reshaped(B, L, -1)
+        let output = MLXFast.scaledDotProductAttention(
+            queries: queries, keys: keys, values: values, scale: scale, mask: mask
+        )
+        .transposed(0, 2, 1, 3)
+        .reshaped(B, L, -1)
```
awni, line 85: It just occurred to me there is an issue here? I think the output will be in fp32 since you are sending the inputs in fp32 (that will propagate through the rest of the model and slow things way down). We need to explicitly downcast the output after the SDPA (might be good to add a test for that).

davidkoski: ah perhaps -- I guess it would still run, so I didn't notice.

Indeed it now produces float32 instead of float16.  So something like this?

```
            queries: queries, keys: keys, values: values, scale: scale, mask: mask
        )
        .asType(values.dtype)
        .transposed(0, 2, 1, 3)
        .reshaped(B, L, -1)
```

awni: Yea I would make a couple changes here:
1. Cast the output to the values type (as suggested)
2. Don't upcast the keys, because we don't want to store fp32 in the KV cache. Just upcast the queries and let the op handle type promotion.

PR #24, opened 2024-03-13
scroll to bottom when text is generated

davidkoski:
- also restore circleci

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,17 +43,12 @@ workflows:
     when:
       and:
         - matches:
-            pattern: "^(?!pull/)[-_./\\w]+$"
+            pattern: "^(?!pull/)[-\\w]+$"
             value: << pipeline.git.branch >>
         - not: << pipeline.parameters.nightly_build >>
         - not: << pipeline.parameters.weekly_build >>
     jobs:
-      - hold:
-          type: approval
-      - apple/authenticate:
-          context: pr-approval
-      - mac_build_and_test:
-          requires: [ hold ]
+      - mac_build_and_test
 
   prb:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -43,17 +43,12 @@ workflows:
...
         - not: << pipeline.parameters.nightly_build >>
         - not: << pipeline.parameters.weekly_build >>
     jobs:
-      - hold:
-          type: approval
-      - apple/authenticate:
-          context: pr-approval
-      - mac_build_and_test:
-          requires: [ hold ]
+      - mac_build_and_test
```
davidkoski, line 51: Restore changes that were done to try and get non-fork PRs to build but broke main.

davidkoski: OK, it looks like this works -- the caveat is you can't name your branch something like `dkoski/scroll`

awni: I see so `name1/name2` breaks the pattern (I guess that's true in our other repos as well).

Applications/LLMEval/ContentView.swift
```diff
@@ -16,11 +16,21 @@ struct ContentView: View {
...
+                    Text(llm.output)
+                        .textSelection(.enabled)
+
+                        .onChange(of: llm.output) { _, _ in
+                            sp.scrollTo("bottom")
+                        }
+
+                    Spacer()
+                        .frame(width: 1, height: 1)
+                        .id("bottom")
```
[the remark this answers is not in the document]
awni: ❤️  !

awni (approved):
Thanks!!

PR #25, opened 2024-03-14
fix float types in Phi (use float16)

davidkoski:
- per suggestions in #23 ensure that the values that go into the cache are float16

### DIFF
--- a/Libraries/LLM/Phi.swift
+++ b/Libraries/LLM/Phi.swift
@@ -73,14 +73,12 @@ private class PhiAttention: Module {
             keys = rope(keys)
         }
 
-        queries = queries.asType(Float.self)
-        keys = keys.asType(Float.self)
-
         // Finally perform the attention computation
         let scale = sqrt(1 / Float(queries.dim(-1)))
         let output = MLXFast.scaledDotProductAttention(
-            queries: queries, keys: keys, values: values, scale: scale, mask: mask
+            queries: queries.asType(.float32), keys: keys, values: values, scale: scale, mask: mask
         )
+        .asType(values.dtype)
         .transposed(0, 2, 1, 3)
         .reshaped(B, L, -1)
 


Libraries/LLM/Phi.swift
```diff
@@ -73,14 +73,12 @@ private class PhiAttention: Module {
...
-
         // Finally perform the attention computation
         let scale = sqrt(1 / Float(queries.dim(-1)))
         let output = MLXFast.scaledDotProductAttention(
-            queries: queries, keys: keys, values: values, scale: scale, mask: mask
+            queries: queries.asType(.float32), keys: keys, values: values, scale: scale, mask: mask
         )
+        .asType(values.dtype)
         .transposed(0, 2, 1, 3)
         .reshaped(B, L, -1)
```
davidkoski, line 83: This provides: 

```
output float16, keys float16, values float16
```

Libraries/LLM/Phi.swift
```diff
@@ -73,14 +73,12 @@ private class PhiAttention: Module {
...
         }
 
-        queries = queries.asType(Float.self)
-        keys = keys.asType(Float.self)
-
         // Finally perform the attention computation
         let scale = sqrt(1 / Float(queries.dim(-1)))
         let output = MLXFast.scaledDotProductAttention(
-            queries: queries, keys: keys, values: values, scale: scale, mask: mask
+            queries: queries.asType(.float32), keys: keys, values: values, scale: scale, mask: mask
```
davidkoski, line 79: I found that if I passed `queries` in as Float16 I would get unimaginative output

image

Is this a reasonable thing to do here?  Should keys be cast to float32 as well (it didn't seem to matter when I tested it)?

@awni

awni (approved):
🚀

Libraries/LLM/Phi.swift
awni: So if you upcast the queries the full attention should be done in fp32 and it should work. The keys will be implicitly cast in the op.

The way you’ve done it looks good to me. Most models don’t need this, phi2 is an exception for some reason

PR #27, opened 2024-03-14
LLMEval UI Improvements

1-ashraful-islam:
Adds some improvements to the LLMEval UI:

1. adds Markdown rendering in the UI
2. Adds init time and token/second stat
3. Minor UI enhancements 
     i. aligns the texts to the left
    ii. always shows the model info and progress bar (when running) on the top
   iii. adds a copy button to copy all the llm output at once

Here's a quick demo:

https://github.com/ml-explore/mlx-swift-examples/assets/5774830/0f41d392-98fe-4002-b7ff-cb80fad4360d

### DIFF
--- a/Applications/LLMEval/ContentView.swift
+++ b/Applications/LLMEval/ContentView.swift
@@ -13,6 +13,13 @@ struct ContentView: View {
     @State var prompt = "compare python and swift"
     @State var llm = LLMEvaluator()
 
+    enum displayStyle: String, CaseIterable, Identifiable {
+        case plain, markdown
+        var id: Self { self }
+    }
+
+    @State private var selectedDisplayStyle = displayStyle.markdown
+
     var body: some View {
         VStack(alignment: .leading) {
             VStack {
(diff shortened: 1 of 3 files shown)


Applications/LLMEval/ContentView.swift
```diff
@@ -6,20 +6,31 @@ import MLXRandom
...
+                    Text(llm.stat)
+                }
                 if llm.running {
                     ProgressView()
                 }
-                Text(llm.output)
+            }
+            // show the model output
+            ScrollView(.vertical) {
+                Markdown(llm.output)
```
davidkoski, line 33: Looks nice!  There was an outstanding PR that makes this scroll as text is appended and it causes a minor conflict here, see #24

Applications/LLMEval/ContentView.swift
```diff
@@ -32,6 +43,20 @@ struct ContentView: View {
...
+            ToolbarItem(placement: .primaryAction) {
+                Button {
+                    Task {
+                        copyToClipboard(llm.output)
+                    }
+                } label: {
+                    Label("Copy Output", systemImage: "doc.on.doc.fill")
+                }
+                .disabled(llm.output == "")
+                .labelStyle(.titleAndIcon)
```
davidkoski, line 56: Curious, why add a toolbar item?  The text selection is enabled so the user can just select whatever they want and copy it normally.

1-ashraful-islam: I added that toolbar item to copy all of text output at once for two reasons:

1. Markdown block appends new content as blocks (header, code etc.). This results in a text selection that is block wise. At the moment, I don't know of a elegant way to solve the issue of selecting all the text from the output when using Markdown rendering. 
2. When the total token count is higher, I believe it would be handy to have a button to copy all the output at once.

Applications/LLMEval/ContentView.swift
1-ashraful-islam: I added the changes from #24 to the ContentView. See: https://github.com/ml-explore/mlx-swift-examples/pull/27/commits/441263bbb1aeff441ce24c9b44261953e87e1c6a

Let me know if any additional changes would be helpful to resolve the conflict.

davidkoski: I think you just need to rebase it on main, so something like this:

- update your fork of main
- update your local copy of main
    - `git checkout main`
    - `git pull`
- switch back to your branch and rebase it
    - `git checkout markdownUI`
    - `git rebase main`
    - resolve the conflicts
    - `git add` the files
    - `git rebase --continue` to finish the rebase
- force push your branch
    - `git push -f`

Something like that.  There are various guides if needed: https://docs.gitlab.com/ee/topics/git/git_rebase.html

Once that is done the commits will be on top of the current HEAD and the merge will be unblocked.

1-ashraful-islam: Really appreciate the detailed step by step. I rebased and reformatted with pre-format as instructed.

1-ashraful-islam:
@davidkoski  In my local version I added gpu usages on the toolbar following instructions linked by @awni here : https://x.com/awnihannun/status/1768422759200432470?s=20

Do you think it could be a good addition to this PR or should I make a separate PR if/once this one is merged. Here's a screenshot of what that looks like:

image

1-ashraful-islam:
I added a segmented picker (https://github.com/ml-explore/mlx-swift-examples/pull/27/commits/59a9c809a09af905902955f75a9568f8e8f94892) to allow switching between plain text and markdown as it might be desirable to do so. This allows adding markdown view without taking the plaintext option away from the user. 

image
image

davidkoski (approved):

PR #32, opened 2024-03-19
make the mnist app/library targets match each other

davidkoski:
Make the MNIST library and app have the same deployment targets -- fewer complaints from Xcode.


awni (approved):
Looks good, though CI is failing (?)

davidkoski:
> Looks good, though CI is failing (?)

Ah perfect, it caught the fact that the command line tool is now out of sync.  Uh, exactly as I planned

CI keeping us honest :-)

PR #33, opened 2024-03-19
fixed height for the progress view, produce more tokens

davidkoski:


Applications/LLMEval/ContentView.swift
```diff
@@ -35,6 +35,7 @@ struct ContentView: View {
                     Spacer()
                     if llm.running {
                         ProgressView()
+                            .frame(maxHeight: 20)
```
davidkoski, line 38: Set a height on this so that the enclosing view doesn't change size as it is running / not running (makes the UI jump around).

Applications/LLMEval/ContentView.swift
```diff
@@ -131,7 +132,7 @@ class LLMEvaluator {
 
     /// parameters controlling the output
     let temperature: Float = 0.6
-    let maxTokens = 100
+    let maxTokens = 240
```
davidkoski, line 135: Produce more tokens, especially helpful now with the markdown ui

awni (approved):
Very nice!

PR #36, opened 2024-03-24
feat: adds gpu usages stat in the toolbar

1-ashraful-islam:
Adds GPU Usages stats on the toolbar. Also added a tooltip that shows breakdown of the memory usages on hover. Uses GPU Snapshot from the docs linked by Awni here: https://x.com/awnihannun/status/1768422759200432470?s=20

image

### DIFF
--- a/Applications/LLMEval/ViewModels/DeviceStat.swift
+++ b/Applications/LLMEval/ViewModels/DeviceStat.swift
@@ -1,0 +1,43 @@
+import Foundation
+import LLM
+import MLX
+
+class DeviceStat: ObservableObject {
+    @Published var gpuUsage: GPU.Snapshot
+    private var initialGPUSnapshot: GPU.Snapshot
+    private var timer: Timer?
+
+    init() {
+        initialGPUSnapshot = GPU.snapshot()
+        gpuUsage = initialGPUSnapshot
+        startTimer()
+    }
+
+    deinit {
+        stopTimer()
+    }
+
+    private func startTimer() {
+        timer?.invalidate()
+        timer = Timer.scheduledTimer(withTimeInterval: 2.0, repeats: true) { [weak self] _ in
+            self?.updateStats()
+        }
+    }
+
+    private func stopTimer() {
+        timer?.invalidate()
+        timer = nil
+    }
+
+    private func updateStats() {
+        updateGPUUsages()
+    }
+
+    private func updateGPUUsages() {
+        let gpuSnapshotDelta = initialGPUSnapshot.delta(GPU.snapshot())
+        DispatchQueue.main.async { [weak self] in
+            self?.gpuUsage = gpuSnapshotDelta
+        }
+    }
+
+}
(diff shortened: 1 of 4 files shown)


Applications/LLMEval/ViewModels/DeviceStat.swift
```diff
@@ -0,0 +1,43 @@
+import Foundation
+import LLM
+import MLX
+
+class DeviceStat: ObservableObject {
```
davidkoski, line 5: Maybe use `@Observable` (like in ContentView)?  The minimum SDK for the apps has these macros and they are more forward looking.

Applications/LLMEval/ViewModels/DeviceStat.swift
```diff
@@ -0,0 +1,43 @@
...
+        initialGPUSnapshot = GPU.snapshot()
+        gpuUsage = initialGPUSnapshot
+        startTimer()
+    }
+
+    deinit {
+        stopTimer()
+    }
+
+    private func startTimer() {
```
davidkoski, line 20: I wonder if the timer should start/stop when inference is running?  Otherwise even when the application is idle this is still running.

Applications/LLMEval/ViewModels/DeviceStat.swift
1-ashraful-islam: Thanks for the comment, I have updated the DeviceStat to use @Observable macro

Applications/LLMEval/ViewModels/DeviceStat.swift
1-ashraful-islam: At the moment - `LLMEvaluator` in the `ContentView` ever really seems at `.idle` at the beginning, after which it quickly switches to loaded state. Also, is there a lot of overhead from this timer invoking the snapshot that would prompt us to keep it say turned off during inferencing?

davidkoski: No, reading stats is pretty low overhead.  I was just thinking that the timer shouldn't be running if the app is idle (not `.idle` but rather not running inference).  The memory use will only change:

- as it is loading the model
- during inference

1-ashraful-islam: It makes sense to implement it that way. Since at the moment, we don't have a state that indicates when the model is running inference - do you think it makes sense to implement that first? Then we can also use this state to start/stop the timer when the model is not actively changing memory. 

This would also allow us to, lets say, put a `Stop Generation` button to stop the inference when we don't like the inference result. When setting the context length higher, this becomes a really big issue (specially if the model is generating nonsensical results). 

At the moment, given that its pretty low overhead, would you say we can move with this pull request and visit this start/stop generation after we implement `.generating` state for the `LLMEvaluator`?

davidkoski: Yes, that sounds reasonable -- as you suggest, I think a stop button would be useful!

davidkoski (approved):

PR #38, opened 2024-03-27
Use fast

awni:
-Update MLX Swift
-Use fast norms for Phi-2 and Gemma


awni:
@davidkoski I noticed most of the models were using the MLXNN normalizations already (which means they may not have been working too well because they didn't used to upcast for the reductions). I updated Phi-2 and Gemma to use the fast norms, so now all the models should do so.

davidkoski (approved):
Changes look good!

Issue #31, opened 2024-03-19
Any plans to create a swift-llm package?


mzbac:
Thanks for creating this amazing example repository. I was just wondering if there are any plans to create a swift-llm package similar to mlx-lm so people don't have to manually copy all the code from the LLM folder into their project. This would remove some barriers for people to try out the swift-mlx. As someone new to Swift, copying and pasting LLM code from the examples and setting it up took quite a bit of time. It would be great if we could provide a package so developers don't have to go through that process.

awni:
> I was just wondering if there are any plans to create a swift-llm package similar to mlx-lm so people don't have to manually copy all the code from the LLM folder into their project. 

Hmm, you shouldn't need to copy and paste the example unless you are trying to get rid of the other stuff like the MNIST part. You can clone the repo and open it in xcode.

mzbac:
@awni I was using SwiftPM to create the new project. So I had to copy all the code from `Libraries/LLM`. I was thinking if it would be possible to provide a swiftPM package so that I can add it via SwiftPM like the swift-mlx: https://github.com/ml-explore/mlx-swift?tab=readme-ov-file#swiftpm

Do you mean we should use the mlx-swift-example as a template and start the new project from it? Sorry, I am quite new to Swift and Xcode, not very sure what's the best way to create a new project with swift-mlx.

awni:
I see.. I think @davidkoski has some ideas here as well.

But basically you have a few options:
- You can use it as a git submodule (no copying required)
- You can copy the files as you did
- You can fork and use it as a template and build on it

As for a Swift package, I'm not sure. It is a good idea in principle, but would probably require splitting the LLM stuff into a new repo (?).

davidkoski:
Here is how you might do it with Xcode:

<img width="1201" alt="image" src="https://github.com/ml-explore/mlx-swift-examples/assets/46639364/8ad9581d-341d-4fd5-8bdf-4cb75baaed25">

Notice I have dragged `mlx-swift-examples.xcodeproj` into the list of files in the project on the left.  In my target, `MyTestApp`, I can link against `LLM.framework` which is what I think you are looking for.  It will only build the pieces related to your target.

In this case I just made a project in `/tmp` and dragged it in, but you could easily use git submodules to make something that checks it all out.

Be aware that you can only have the `mlx-swift-examples` directory open once in Xcode.  If it shows up without the disclosure indicator on the left it probably means you already have it open.  Just close both windows and reopen your app.

[...]

PR #39, opened 2024-03-28
Add Package.swift

ronaldmannak:
This pull request addresses the issue of projects manually copying the LLM directory from the Libraries folder.

With this PR, developers can now include the LLM and/or MNIST libraries in their projects via Swift Package Manager. This update simplifies the process of keeping the LLM dependency current.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -3,7 +3,7 @@
 
 import PackageDescription
 let package = Package(
-    name: "MLX-Libraries",
+    name: "mlx-libraries",
     platforms: [.macOS(.v14), .iOS(.v16),],
     products: [
         .library(
@@ -45,6 +45,7 @@ let package = Package(
                 .product(name: "MLXRandom", package: "mlx-swift"),
                 .product(name: "Transformers", package: "swift-transformers"),
                 .product(name: "AsyncAlgorithms", package: "swift-async-algorithms"),
+                .product(name: "Gzip", package: "GzipSwift"),
             ],
             path: "Libraries/MNIST",
             exclude: [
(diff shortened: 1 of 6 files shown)


ronaldmannak:
As discussed in #31

ronaldmannak:
It looks like lint fails, but I don't see any useful error. Who can help me with this?

davidkoski:
> It looks like lint fails, but I don't see any useful error. Who can help me with this?

Sure, the ci builds says:

```
swift-format.............................................................Failed
```

It runs swift-format to make sure all the files have a consistent style -- this includes Package.swift!

Just run:

```
pre-commit run --all-files
```

ronaldmannak:
@davidkoski Thanks but I understood that part, I was hoping to get a useful error message from the CI logs. 

It turned out to be not much work to install swift-format and add a pre-commit. Committed fix.

ronaldmannak:
The targets are currently named MLXLLM and MLXMNIST, while the example libraries are simply called LLM and MNIST. If there's a preference for the simpler names, I'm open to rename the targets.

Package.swift
```diff
@@ -0,0 +1,57 @@
+// swift-tools-version: 5.9
+// The swift-tools-version declares the minimum version of Swift required to build this package.
+
+import PackageDescription
```
davidkoski, line 4: This is still failing the swift-format check.   In particular it wants:

```swift
 import PackageDescription
+
 let package = Package(
```

davidkoski:
I think this is a really great idea!

Please add info to the top level README.md indicating how to make use of it.

Package.swift
ronaldmannak: fixed in the latest commit. Thanks.

davidkoski (approved):
Looks great, thank you for the idea and the contribution!

PR #40, opened 2024-03-28
LLMEval performance

davidkoski:
- notes about running under Xcode
- reminder about models having different performance
- don't update the display for every token

### DIFF
--- a/Applications/LLMEval/ContentView.swift
+++ b/Applications/LLMEval/ContentView.swift
@@ -151,6 +151,11 @@ class LLMEvaluator {
     /// parameters controlling the output
     let temperature: Float = 0.6
     let maxTokens = 240
+    
+    /// update the display every N tokens -- 4 looks like it updates continuously
+    /// and is low overhead.  observed ~15% reduction in tokens/s when updating
+    /// on every token
+    let displayEveryNTokens = 4
 
     enum LoadState {
         case idle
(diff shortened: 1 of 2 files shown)


Applications/LLMEval/ContentView.swift
```diff
@@ -198,7 +203,7 @@ class LLMEvaluator {
             let prompt = modelConfiguration.prepare(prompt: prompt)
             let promptTokens = MLXArray(tokenizer.encode(text: prompt))
 
-            let initTime = Date()
+            var initTime = Date()
```
awni, line 206: I guess you don't really need to initialize it here?

Applications/LLMEval/README.md
```diff
@@ -50,3 +50,9 @@ Building in Release / optimizations will remove a lot of tail calls in the C++
 layer.  These lead to the stack overflows.
 
 See discussion here: https://github.com/ml-explore/mlx-swift-examples/issues/3
+
+### Performance
+
+Different models have difference performance characteristics.  For example gemma may outperform phi2 in terms of tokens / second.
```
awni, line 56:
```suggestion
Different models have difference performance characteristics. For example Gemma 2B may outperform Phi-2 in terms of tokens / second.
```

Applications/LLMEval/README.md
```diff
@@ -50,3 +50,9 @@ Building in Release / optimizations will remove a lot of tail calls in the C++
...
+
+You may also find that running outside the debugger is also a performance boost.  You can do this in xcode by pressing cmd-opt-r and uncheck "Debug Executable".
```
awni, line 58:
```suggestion
You may also find that running outside the debugger boosts performance.  You can do this in Xcode by pressing cmd-opt-r and unchecking "Debug Executable".
```

awni (approved):
Minor comments, looks great! Thanks for debugging the perf issues! 🚀

Applications/LLMEval/ContentView.swift
davidkoski: It is used on the line below to show the time spent preparing, so this is still needed.

PR #43, opened 2024-03-30
Append visionOS support to LLMEval

florentmorin:
You can now run LLMEval on Apple Vision Pro 😉


davidkoski (approved):
Great, thank you for the contribution!

PR #46, opened 2024-04-05
implement LoRA / QLoRA

davidkoski:
- example of using MLX to fine-tune an LLM with low rank adaptation (LoRA) for a target task
- see also https://arxiv.org/abs/2106.09685
- based on https://github.com/ml-explore/mlx-examples/tree/main/lora

Note that this sits on top of #47 and requires this PR from mlx-swift: https://github.com/ml-explore/mlx-swift/pull/65


Libraries/LLM/Llama.swift
```diff
@@ -12,9 +12,9 @@ private class Attention: Module {
     let args: LlamaConfiguration
     let scale: Float
 
-    @ModuleInfo(key: "q_proj") var wq: Linear
+    @ModuleInfo(key: "q_proj") var wq: LoRAReplacableLinear
```
davidkoski, line 15: We have to use a different type because we are going to replace the `Linear` with a `LoRALinear` which is not a sub-type of `Linear`.

- [ ] document this

Libraries/LLM/Llama.swift
```diff
@@ -253,3 +253,20 @@ public struct LlamaConfiguration: Codable {
 
     }
 }
+
+// MARK: - LoRA
```
davidkoski, line 257: This is how we add the introspection required to apply the LoRA adapter.

- [ ] document this

Libraries/LLM/Lora.swift
```diff
@@ -0,0 +1,356 @@
...
+
+/// Protocol for layers that should have ``LoRALinear`` applied.  Typically
+/// implemented by `Attention` layers.
+public protocol LoRALayer: Module {
+
+    /// Return the `Linear` layers that should have ``LoRALinear`` applied.
+    func loraLinearModules() -> [String: LoRAReplacableLinear]
+}
+
+public protocol LoRAReplacableLinear: Module, UnaryLayer {
```
davidkoski, line 27: Marker type that allows assignment from both `Linear` and `LoRALinear`

Libraries/LLM/Lora.swift
```diff
@@ -0,0 +1,356 @@
...
+        }
+
+        index = endIndex
+
+        return (batchArray[0..., .stride(to: -1)], batchArray[0..., 1...], MLXArray(lengths))
+    }
+
+}
+
+public enum LoRATrain {
```
davidkoski, line 194: This is the public API -- just free functions grouped into this LoRATrain namespace.

Tools/llm-tool/LLMTool.swift
```diff
@@ -11,15 +11,24 @@ import Tokenizers
 struct LLMTool: AsyncParsableCommand {
     static var configuration = CommandConfiguration(
         abstract: "Command line tool for generating text using Llama models",
-        subcommands: [SyncGenerator.self, AsyncGenerator.self],
+        subcommands: [SyncGenerator.self, AsyncGenerator.self, LoRACommand.self],
```
davidkoski, line 14: Added a `lora` subcommand to the `llm-tool`

Tools/llm-tool/LLMTool.swift
```diff
@@ -11,15 +11,24 @@ import Tokenizers
...
         defaultSubcommand: SyncGenerator.self)
 }
 
-struct LLMArguments: ParsableArguments {
+struct ModelArguments: ParsableArguments {
```
davidkoski, line 18: Refactoring these argument bundles so I can reuse them in a variety of situations, not just the generation code in the original example.

Tools/llm-tool/LLMTool.swift
```diff
@@ -29,19 +38,100 @@ struct LLMArguments: ParsableArguments {
...
+        
+        func summary() -> String {
+            """
+            Prompt Tokens per second:     \((Double(promptTokens.count) / promptTime).formatted())
+            Generation tokens per second: \((Double(tokens.count - 1) / generateTime).formatted())
+            """
+        }
+    }
+    
+    func generate(promptTokens: [Int], model: LLMModel, tokenizer: Tokenizer, didGenerate: ([Int]) -> Void = { _ in }) -> GenerateResult {
```
davidkoski, line 87: Make this reusable.

Tools/llm-tool/LLMTool.swift
```diff
@@ -113,46 +210,27 @@ struct SyncGenerator: AsyncParsableCommand {
         abstract: "Synchronous generator"
     )
 
-    @OptionGroup var args: LLMArguments
+    @OptionGroup var args: ModelArguments
+    @OptionGroup var memory: MemoryArguments
+    @OptionGroup var generate: GenerateArguments
```
davidkoski, line 215: Use the refactored arguments

Tools/llm-tool/LLMTool.swift
```diff
@@ -193,15 +260,17 @@ struct AsyncGenerator: AsyncParsableCommand {
         abstract: "async generator"
     )
 
-    @OptionGroup var args: LLMArguments
+    @OptionGroup var args: ModelArguments
+    @OptionGroup var memory: MemoryArguments
+    @OptionGroup var generate: GenerateArguments
```
davidkoski, line 265: The async version also gets these but it keeps the async generation as-is.  I am thinking of removing the async version -- I thought it would be a useful example for applications, but given how the example apps are working, I think this is just overcomplicating things and making us do twice the work to keep this in sync.

Anyway, that is for another PR :-)

Tools/llm-tool/LoraCommands.swift
```diff
@@ -0,0 +1,216 @@
...
+import MLXOptimizers
+import MLXRandom
+import Tokenizers
+
+struct LoRACommand: AsyncParsableCommand {
+
+    static var configuration = CommandConfiguration(
+        commandName: "lora",
+        abstract: "LoRA commands",
+        subcommands: [LoRATrainCommand.self, LoRATestCommand.self, LoRAEvalCommand.self]
```
davidkoski, line 17: Here are the LoRA subcommands:

```
SUBCOMMANDS:
  train                   LoRA training
  test                    LoRA testing
  eval                    LoRA evaluation
```

Tools/llm-tool/LoraCommands.swift
```diff
@@ -0,0 +1,216 @@
...
+    @OptionGroup var args: LoRAModelArguments
+    @OptionGroup var memory: MemoryArguments
+
+    @Flag(help: "Resume training with the given adapter file")
+    public var resume = false
+
+    @Option(name: .long, help: "Directory with {train, valid, test}.{jsonl,txt} files")
+    public var data: URL = URL(filePath: "data")
+
+    // TODO: option docs
```
davidkoski, line 75: Still some TODOs but the arguments work :-)

Libraries/LLM/Load.swift
```diff
@@ -15,6 +15,7 @@ struct LLMError: Error {
 /// Load and return the model and tokenizer
 public func load(
     hub: HubApi = HubApi(), configuration: ModelConfiguration,
+    overrideWeights: URL? = nil,
```
davidkoski, line 18: Override URL for weights -- useful if we want to use fused weights from LoRA

Libraries/LLM/Lora.swift
```diff
@@ -0,0 +1,545 @@
+// Copyright © 2024 Apple Inc.
+
+import Foundation
+import MLX
+import MLXNN
+import MLXOptimizers
+import MLXRandom
+import Tokenizers
```
davidkoski, line 8: This is the LoRA implementation

Libraries/LLM/Lora.swift
```diff
@@ -0,0 +1,545 @@
...
+    ///   - validate: validate dataset
+    ///   - optimizer: optimizer used in training
+    ///   - loss: loss function
+    ///   - tokenizer: tokenizer
+    ///   - parameters: training parameters
+    ///   - progress: progress callback
+    public static func train(
+        model: LoRAModel, train: [String], validate: [String], optimizer: Optimizer,
+        loss: @escaping LoraLossFunction = loss, tokenizer: Tokenizer, parameters: Parameters,
+        progress: (Progress) async -> ProgressDisposition
```
davidkoski, line 458: Rather than printing progress this is modeled as a callback that provides the information.  This might allow for some interesting applications:

- adaptive training that stops once a certain loss rate is hit
- integration in an example application, e.g. something running on iOS -- the training loop can be called by the UI code and received information about progress

Tools/llm-tool/README.md
```diff
@@ -60,3 +60,187 @@ Building in Release / optimizations will remove a lot of tail calls in the C++
...
+This is an example of using MLX to fine-tune an LLM with low rank adaptation
+(LoRA) for a target task.[^lora] The example also supports quantized LoRA
+(QLoRA).[^qlora] The example works with Llama and Mistral style models
+available on Hugging Face.
+
+In this example we'll use the WikiSQL[^wikisql] dataset to train the LLM to
+generate SQL queries from natural language. However, the example is intended to
+be general should you wish to use a custom dataset.
+
+> Note: Some of the prompts have newlines in them which is difficult to achieve via running in Xcode.
```
awni, line 79: Can you explain that comment?

Tools/llm-tool/README.md
```diff
@@ -60,3 +60,187 @@ Building in Release / optimizations will remove a lot of tail calls in the C++
...
+  fuse                    Fuse lora adapter weights back in to original model
+  test                    LoRA testing
+  eval                    LoRA evaluation
+```
+
+### Training
+
+The first step will be training the LoRA adapter.  For this example we will 
+assume https://github.com/ml-explore/mlx-examples has been checked out into 
+your home directory -- please adjust as necessary.  This will use `lora/data`
```
awni, line 95: Maybe we should just package the `lora/data` in `mlx-swift-examples` to make this a bit easier.? It is only 300KB.

Libraries/LLM/Load.swift
davidkoski: Here is the fuse command:

- https://github.com/ml-explore/mlx-swift-examples/pull/46/files#diff-57fb3aa82dee5fb24ff29e9b450e9745821d31bb65cf126cb27fb4e0d4c051deR150

So you would make a new Hub entry with the old config / tokenizer + new weights?  Do you upload that to hugging face or just do that locally?

I was just putting the fused weights e.g. in /tmp and reading the original model files + new weights that way.  It looks like the `Hub` API from `swift-transformers` doesn't provide methods to upload so I would need to see how that is done (if we need the upload part).

awni: > So you would make a new Hub entry with the old config / tokenizer + new weights? 

Yes but it's not a Hub entry it's just a local path (the code can use HF repos or local paths)

> Do you upload that to hugging face or just do that locally?

Both are supported. But I think it's fine to just do it locally for now.

Libraries/LLM/Llama.swift
davidkoski: Yes, it is possible to do that -- this was mimicking the structure from the python code.  Now that I see how all the serialization works, it would be doable.

Tools/llm-tool/LLMTool.swift
davidkoski: BTW the async version is removed and this refactoring is already done in #47

Tools/llm-tool/README.md
davidkoski: Sure, for testing this you need to pass a prompt like this:

```
    --prompt "table: 1-10015132-16
columns: Player, No., Nationality, Position, Years in Toronto, School/Club Team
Q: What is terrence ross' nationality
A: "
```

The newlines seem to be important.  You can put that in the Xcode parameters field but it seems to eat the newlines:

image

Libraries/LLM/Load.swift
davidkoski: OK, I will set this up!

Tools/llm-tool/README.md
awni: I see, that makes sense.

awni:
This is super cool! I think we should merge it after updates discussed already and we can refine usage etc as needed.

Also I'm wondering did you run it on iOS? If so how did you do it? I don't see the requisite app?

davidkoski:
> This is super cool! I think we should merge it after updates discussed already and we can refine usage etc as needed.
> 
> Also I'm wondering did you run it on iOS? If so how did you do it? I don't see the requisite app?

Yes, resource-wise it could easily run on iOS.  I could make an example application like this:

- contains a directory with the train/test/etc. files
- a button to train
- a text field to test the result

It isn't quite as cool as the command line tool because you package the data with the application but it could be useful as an example of how to do it.

Libraries/LLM/Llama.swift
davidkoski: Slightly trickier than anticipated -- I was thinking I could just do something like QuantizedLinear where it is Linear + scales + biases.  That would work fine, but I need the same thing for QuantizedLinear too.  It works much better as a LoRALinear which has-a Linear (or any subtype such as QuantizedLinear).

The other option would be to have LoRALinear and LoRAQuantizedLinear.  That isn't too bad since most of the code is in the toLinear() method and that has to deal with QuantizedLinear anyway.

awni: I didn't quite follow the first option. Second option seems ok. Maybe another option is something like a LoRA Linear that takes a union [QuantizedLinear, Linear]?

davidkoski: > Maybe another option is something like a LoRA Linear that takes a union [QuantizedLinear, Linear]?

That is what we have now, but the sub typing doesn't work right if it is also is-a Linear (instead of LoRAReplacableLinear).  LoRAReplacableLinear is roughly union [Linear, QuantizedLinear, LoraLinear] as expressed in swift.

If we were to have a single type, LoRALinear that is-a Linear we would have to have both has-a Linear and also is-a Linear (two sets of weights).  I think that would be a mess.

If that didn't make sense, consider how the call will work:

```swift
    public override func callAsFunction(_ x: MLXArray) -> MLXArray {
        let y = super.callAsFunction(x.asType(weight.dtype))
        let z = matmul(matmul(x, self.loraA), self.loraB)
        return y + scale * z
    }
```

If super can't be both a `Linear` and a `QuantizedLinear`.

Anyway, it probably doesn't matter -- the first option looks like it will work fine.  I am just getting it all hooked up right now.

Libraries/LLM/Lora.swift
```diff
@@ -0,0 +1,564 @@
...
+/// This is not typically used directly -- ``LoRATrain/convert(model:layers:)`` is used to
+/// add the adapter layers to a given model.
+///
+/// ### See Also
+/// - [LoRA: Low-Rank Adaptation of Large Language Models](https://arxiv.org/abs/2106.09685)
+/// - [QLoRA: Efficient Finetuning of Quantized LLMs](https://arxiv.org/abs/2305.14314)
+/// - ``LoRALayer``
+/// - ``LoRATrain/convert(model:layers:)``
+/// - ``LoRATrain/fuse(model:deQuantize:)``
+public class LoRALinear: Linear, LoRAConvertToLinear {
```
davidkoski, line 94: @awni check this one out -- this is the replacement for `Linear`

Libraries/LLM/Lora.swift
```diff
@@ -0,0 +1,564 @@
...
+    /// This is typically called via ``LoRATrain/fuse(model:deQuantize:)``.
+    ///
+    /// ### See Also
+    /// - ``LoRATrain/fuse(model:deQuantize:)``
+    public func toLinear(deQuantize: Bool = false) -> Linear {
+        let dtype = weight.dtype
+        let loraB = (scale * loraB.T).asType(dtype)
+        let loraA = loraA.T.asType(dtype)
+        return Linear(weight: weight + matmul(loraB, loraA), bias: bias)
+    }
```
davidkoski, line 142: This is actually much cleaner because it doesn't have to worry about the quantized case

Libraries/LLM/Lora.swift
```diff
@@ -0,0 +1,564 @@
...
+
+    public override func callAsFunction(_ x: MLXArray) -> MLXArray {
+        let y = super.callAsFunction(x.asType(weight.dtype))
+        let z = matmul(matmul(x, self.loraA), self.loraB)
+        return y + scale * z
+    }
+}
+
+public class QLoRALinear: QuantizedLinear, LoRAConvertToLinear {
```
davidkoski, line 151: @awni and here is the QLoRALinear for QuantizedLinear

I still need to update the docs, but now we don't need the special markers in the model code.  We are also not conditional on the type any more at the cost of a small amount of code duplication so I think it looks cleaner and easier to understand.

awni: Looks great to me!

Libraries/LLM/Llama.swift
davidkoski: This is complete -- it is mostly the same from the caller's point of view but now the layers are subclasses of Linear and you don't need anything special in the host model.

Libraries/LLM/Lora.swift
davidkoski: Gone

Tools/llm-tool/LoraCommands.swift
davidkoski: Docs are complete

Libraries/LLM/Load.swift
davidkoski: Done -- I will summarize in just a sec.

Tools/llm-tool/README.md
davidkoski: Done

Applications/LoRATrainingExample/ContentView.swift
```diff
@@ -0,0 +1,284 @@
...
+    private func startInner() async throws {
+        // setup
+        GPU.set(cacheLimit: 32 * 1024 * 1024)
+        await MainActor.run {
+            output = ""
+            state = .training
+        }
+
+        // load the model
+        let (model, tokenizer) = try await loadModel()
```
davidkoski, line 185: @awni this is an example of the load / train / test loop.  At least half the code is just updating the UI :-)

davidkoski:
@awni discussed changes are complete:

- LoRALinear and QLoRALinear subclass Linear -- no special changes required in the host model
    - this requires https://github.com/ml-explore/mlx-swift/pull/68 for some subclass abilities like QuantizedLinear has now
- fusing makes a new hub directory either at a specified path or in existing hub directory -- upload is left to the user if desired
- test data is now included (that was a good idea -- it simplified a few things)
- there is an example application that loads a model & training data, trains the LoRA layers, tests it and gives you a prompt to evaluate it
    - a good place for people to start who e.g. want to train a model on their own text data

I think that addresses all of the things we talked about.

davidkoski:
BTW CI is failing because it requires the change to mlx-swift be landed first.

Tools/llm-tool/README.md
```diff
@@ -60,3 +60,184 @@ Building in Release / optimizations will remove a lot of tail calls in the C++
 layer.  These lead to the stack overflows.
 
 See discussion here: https://github.com/ml-explore/mlx-swift-examples/issues/3
+
```
awni, line 63: What do you think about removing that section? Is it still relevant with the latest fixes in MLX core?

davidkoski: It isn't but I think there are some more references -- I was going to make a pass through the docs once this PR and stable diffusion land (since these are touching some of the same files)

awni (approved):
🚀

PR #52, opened 2024-04-18
Update Package.swift to add MLXOptimizers dependency

JustinMeans:
Adds MLXOptimizers dependency.

Wouldn't build before – "No such module 'MLXOptimizers'".  Build now succeeds.

Xcode 15.4 Beta.


davidkoski (approved):
Awesome, thank you for the fix!

PR #54, opened 2024-04-23
phi3

awni:


Libraries/LLM/Evaluate.swift
```diff
@@ -60,16 +60,16 @@ public struct GenerateParameters {
     public var temperature: Float = 0.6
 
     /// top p sampling
-    public var topP: Float = 0.9
+    public var topP: Float = 1.0
 
     /// penalty factor for repeating tokens
-    public var repetitionPenalty: Float = 1.0
+    public var repetitionPenalty: Float?
```
davidkoski, line 66: Per offline discussion switched these to match the defaults from python

davidkoski (approved):
Looks great!

PR #62, opened 2024-04-26
added editable text for prompt inputs

h21k:
- passed swift-format
- makes the text editable for custom prompts on the device


MatthewWaller:
There is already a TextField BTW, down below on line 81. If you run it on device, you can tap the prompt and edit it.

PR #63, opened 2024-04-29
add support for OpenELM

smdesai:
@awni  Here's the PR

### DIFF
--- a/Libraries/LLM/Models.swift
+++ b/Libraries/LLM/Models.swift
@@ -162,6 +162,7 @@ extension ModelConfiguration {
                 phi34bit,
                 gemma2bQuantized,
                 qwen205b4bit,
+                openelm270m4bit
             ])
             bootstrapState = .bootstrapped
 
--- a/Libraries/LLM/Models.swift
+++ b/Libraries/LLM/Models.swift
@@ -162,7 +162,7 @@ extension ModelConfiguration {
                 phi34bit,
                 gemma2bQuantized,
                 qwen205b4bit,
-                openelm270m4bit
+                openelm270m4bit,
             ])
             bootstrapState = .bootstrapped
 
--- a/Libraries/LLM/Models.swift
+++ b/Libraries/LLM/Models.swift
@@ -137,6 +137,12 @@ extension ModelConfiguration {
         "<|im_start|>system\nYou are a helpful assistant<|im_end|>\n<|im_start|>user\n\(prompt)<|im_end|>\n<|im_start|>assistant"
     }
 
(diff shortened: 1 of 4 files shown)


Libraries/LLM/Models.swift
```diff
@@ -137,6 +137,12 @@ extension ModelConfiguration {
         "<|im_start|>system\nYou are a helpful assistant<|im_end|>\n<|im_start|>user\n\(prompt)<|im_end|>\n<|im_start|>assistant"
     }
 
+    public static let openelm270m4bit = ModelConfiguration(
+        id: "mlx-community/OpenELM-270M-Instruct"
+    ) { prompt in
+        "\(prompt)"
```
MatthewWaller, line 143: First off, phenomenal work! I tested it, and it seems to be doing completion just fine. Do you know how we should format it for instruction? Is it like the phi model at all?, or the other ones like the model above? I didn't see any special tokens for the instruct version thus far.

smdesai: Thanks very much Matthew. I've no idea what the chat template format should be and perhaps someone from Apple can comment.

MatthewWaller: Ah, looks like I found something helpful here, just not sure how to translate to the code https://github.com/apple/corenet/blob/main/projects/openelm/instruction_tuning/openelm-instruct.yaml

smdesai: Oh that looks useful, let me grab the relevant content from the chat template and try it out. Thanks for looking into it.

smdesai: @MatthewWaller I tried the following template without luck.
 "<|system|>\nYou are a helpful assistant<|end|>\n<|user|>\(prompt)<|end|>\n<|assistant|>"

smdesai: Using the default template for the Llama tokenizer, "<s>[INST]\(prompt)[/INST]" seems to work but I'll leave it to someone who knows better.

MatthewWaller: Yeah, overall this is working well for completion. I thing for chat there are some things we can do after this PR: there is a bug in swift transformers that just got fixed in main pertaining to an encoding special tokens. AND the config isn’t setup to recognize tokens like “<|user|>” and such, so we would need to adjust that too eventually.

awni:
@smdesai could you run the swift formatting?

smdesai:
@awni It's run.

davidkoski (approved):
Thank you for the contribution!!

PR #68, opened 2024-05-08
remove the bias in the ffn module for the OpenELM model

nathanrchn:
I noticed a significant performance gap when running the `mlx-community/OpenELM-270M-Instruct` model between mlx-examples and mlx-swift-examples. Indeed, the speed was 18 t/s for swift and 77 t/s for python. This was due to the fact that the biases of the ffn layers were not deactivated and were therefore present in float32 precision.
This PR fixes the bias to false to avoid having biases.

### DIFF
--- a/Libraries/LLM/OpenELM.swift
+++ b/Libraries/LLM/OpenELM.swift
@@ -106,8 +106,8 @@ private class FeedForwardNetwork: Module, UnaryLayer {
         let intermediateDim = Int(
             makeDivisible(Float(ffnMultiplier) * Float(dim), divisor: args.ffnDimDivisor))
 
-        self.proj_1 = Linear(dim, 2 * intermediateDim)
-        self.proj_2 = Linear(intermediateDim, dim)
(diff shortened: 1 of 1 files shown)


awni:
Looks good, thanks for debugging that!

@davidkoski do you know if the swift model loading is strict? In Python when you try to load a model it will complain if not all the weights were loaded. We should probably do the same in Swift as that has been a big time saver for debugging these sorts of issues.

davidkoski:
> Looks good, thanks for debugging that!
> 
> @davidkoski do you know if the swift model loading is strict? In Python when you try to load a model it will complain if not all the weights were loaded. We should probably do the same in Swift as that has been a big time saver for debugging these sorts of issues.

Yes, there is a verify parameter but by default it is `.none` -- perhaps that should change.

https://swiftpackageindex.com/ml-explore/mlx-swift/main/documentation/mlxnn/module/update(parameters:verify:)

davidkoski (approved):
Thanks for investigating this!

Issue #74, opened 2024-05-16
Phi-3 mini stop token not recognized


DePasqualeOrg:
[This Phi 3 model](https://huggingface.co/mlx-community/Phi-3-mini-4k-instruct-4bit-no-q-embed) used in the LLMEval app doesn't behave as expected. It looks like the stop token is not being recognized.

Prompt:

```
Name a color.
```

Output:

```
Blue.<|end|><|assistant|> Blue is a primary color that is often associated with calmness, stability, and serenity. It is a color that can be found in nature, from the vast ocean to the clear sky.<|end|><|assistant|> Here is a color:

Purple.<|end|><|assistant|> Purple is a secondary color created by mixing equal parts of red and blue. It is often associated with creativity, wisdom, and luxury. Purple can evoke a sense of royalty and is sometimes linked to spirituality and mysticism.<|end|><|assistant|> Another color is:

Green.

Green is a primary color that symbolizes growth, harmony, and freshness. It is the color of nature, often associated with renewal and vitality. Green is also commonly linked to environmentalism and sustainability.<|end|>
```

Related issues:
https://github.com/ggerganov/llama.cpp/issues/6903
https://github.com/nomic-ai/gpt4all/issues/2271
https://github.com/huggingface/swift-transformers/issues/98

awni:
We do a combination of 1 (change the config) and 3. Regarding 2, it's not clear to me if the original model authors intended the model to have two EOS tokens or if it was just an oversight not to update it in the tokenizer...

davidkoski:
OK, then let's plan on adding #3 -- that will make it flexible at least.

davidkoski:
OK, I am working on the model configuration for #53 so it will be included with that.

davidkoski:
#76 should fix this

Issue #71, opened 2024-05-12
LLM evaluator setting any repetitionPenalty crashes the program


shawiz:
I'm adding a repetitionPenalty to the GenerateParameters constructor. Regardless what values I set (I tried 0.5, 1, 1.2, 1.5), it crashes the program immediate as the evaluator runs. I was testing various Qwen1.5 models. Error message I got is 

` -[MTLDebugComputeCommandEncoder dispatchThreads:threadsPerThreadgroup:]:1441: failed assertion '(threadsPerGrid.width(0) * threadsPerGrid.y(1) * threadsPerGrid.depth(0))(0) must not be 0.'`

davidkoski:
I tried this on `mlx-community/Mistral-7B-v0.1-hf-4bit-mlx` without error.  The Qwen1.5 models are not loading for me because of #53 -- I will try it again once that is resolved.

davidkoski:
OK, I can reproduce it.  I won't check it in yet because of the quantization issues in #53 -- it will be included there.

If you want to try it locally, here is the change:

```swift
diff --git a/Libraries/LLM/Evaluate.swift b/Libraries/LLM/Evaluate.swift
index 6c85558..89212a7 100644
--- a/Libraries/LLM/Evaluate.swift
+++ b/Libraries/LLM/Evaluate.swift
@@ -31,7 +31,7 @@ private func applyRepetitionPenalty(
 ) -> MLXArray {
     if repetitionContext.shape[0] > 0 {
         let indices = repetitionContext
-        var selectedLogits = take(logits, indices, axis: -1).squeezed(axis: 0)
+        var selectedLogits = logits[0..., indices]
 
         selectedLogits = MLX.where(
             selectedLogits .< 0, selectedLogits * penalty, selectedLogits / penalty)
@@ -100,7 +100,7 @@ public struct TokenIterator: Sequence, IteratorProtocol {
             if prompt.shape[0] <= parameters.repetitionContextSize {
                 self.repetitionContext = prompt
             } else {
-                self.repetitionContext = prompt[-parameters.repetitionContextSize ... -1]
+                self.repetitionContext = prompt[(-parameters.repetitionContextSize)...]
             }
         } else {
             self.repetitionContext = []
@@ -120,9 +120,8 @@ public struct TokenIterator: Sequence, IteratorProtocol {
         y = sample(logits: logits, temp: parameters.temperature, topP: parameters.topP)
[...]

Issue #69, opened 2024-05-09
`generate` silently fails on OpenELM 3B after latest commit


DePasqualeOrg:
[Original comment removed, see below]

```swift
import Foundation
import MLXLLM

extension ModelConfiguration {
  static let openelm3b = ModelConfiguration(
    id: "mlx-community/OpenELM-3B"
  ) { prompt in
    "\(prompt)"
  }
}
```

DePasqualeOrg:
After testing it with a smaller model, it appears to work, but the larger model fails without any error. Maybe it's too large for my 16 GB of RAM, but it's not crashing the app when I try to run inference. Providing error messages in the MLXLLM library and in the demo app would help debug these issues.

```swift
import Foundation
import MLXLLM

extension ModelConfiguration {
  // This one fails silently
  static let openelm3b = ModelConfiguration(
    id: "mlx-community/OpenELM-3B"
  ) { prompt in
    "\(prompt)"
  }

  // This one works
  static let openelm450mInstruct = ModelConfiguration(
    id: "mlx-community/OpenELM-450M-Instruct"
  ) { prompt in
    "\(prompt)"
  }
}
```

DePasqualeOrg:
The silent failing of OpenELM 3B seems to be caused by commit [https://github.com/ml-explore/mlx-swift-examples/commit/ff7610d522a57e403140e8431ebd6d032dcdb86c](https://github.com/ml-explore/mlx-swift-examples/commit/ff7610d522a57e403140e8431ebd6d032dcdb86c). It doesn't happen when checking out the commit before that.

davidkoski:
OK, #53 should address this

davidkoski:
#76 should fix this

Issue #53, opened 2024-04-23
LLMEval not loading Qwen1.5 -0.5B  model in to memory


mobile-appz:
When trying to load Qwen1.5, the model downloads fully but doesn't appear to load in to memory on MacOS or iOS. After typing a prompt, the error output is "Failed: unhandledKeys(base: "Embedding", keys: ["biases", "scales"]) 

Using MLX 0.11.0

Other linked models work as per the repo code but this is the smallest, which looks like best one for older devices with less RAM and would be great to get it working.

awni:
Right, we changed quantization in MLX core so now the embedding layer is quantized. We'll need to update Swift to do the same.

mobile-appz:
> Right, we changed quantization in MLX core so now the embedding layer is quantized. We'll need to update Swift to do the same.

Thanks for the info. I was totally unsure as to the cause of this error message. To update, I tried to load this in LLM tool in mlx-swift-examples and that failed with the same error. I then tried to run the python code in the [mlx-examples](https://github.com/ml-explore/mlx-examples) and the model did load and process a prompt. However, the output wasn't worthwhile for anything apparently useful, probably because the model is so small.

davidkoski:
I think these are the commits in question:

- https://github.com/ml-explore/mlx/pull/994/files
- https://github.com/ml-explore/mlx-examples/pull/680/files

awni:
Those are the commits. Sorry that broke more stuff than I was expecting. Basically the embeddings are default quantized now. So when we quantize for MLX in python the model is not usable in Swift because it doesn't support quantized embeddings.

The medium term solution is to update Swift to quantize embeddings (this is a swift only change, don't need anything from core). But as a temporary patch, we could also upload models without embedding layers quantized.

mobile-appz:
At least 1 small model, that can run on an older iOS 17 compatible iPhone, without embedding layers quantized would be really useful for experimentation purposes. Thanks.

davidkoski:
> Those are the commits. Sorry that broke more stuff than I was expecting. Basically the embeddings are default quantized now. So when we quantize for MLX in python the model is not usable in Swift because it doesn't support quantized embeddings.
> 
> The medium term solution is to update Swift to quantize embeddings (this is a swift only change, don't need anything from core). But as a temporary patch, we could also upload models without embedding layers quantized.

[...]

PR #76, opened 2024-05-20
handle partially quantized models

davidkoski:
- fix for #53 #71 #69 #74
- in order to test the models
	- I added a default prompt of an appropriate form
	- while working on the model configuration also added additional stop tokens (#74)
- fixed the repetitionPenalty code (#71)

Note: this isn't directly usable as it requires https://github.com/ml-explore/mlx-swift/pull/73 (I was using the local checkout for development).  This will need an update to the mlx version after 73 is merged.

### DIFF
--- a/Applications/LLMEval/ContentView.swift
+++ b/Applications/LLMEval/ContentView.swift
@@ -10,7 +10,7 @@ import Tokenizers
 
 struct ContentView: View {
 
-    @State var prompt = "compare python and swift"
+    @State var prompt = ""
     @State var llm = LLMEvaluator()
     @Environment(DeviceStat.self) private var deviceStat
 
@@ -125,6 +125,8 @@ struct ContentView: View {
 
         }
         .task {
+            self.prompt = llm.modelConfiguration.defaultPrompt
+
             // pre-load the weights on launch to speed up the first generation
             _ = try? await llm.load()
         }
@@ -224,7 +226,7 @@ class LLMEvaluator {
 
             let result = await LLM.generate(
                 promptTokens: promptTokens, parameters: generateParameters, model: model,
-                tokenizer: tokenizer
+                tokenizer: tokenizer, configuration: modelConfiguration
             ) { tokens in
                 // update the output -- this will make the view show the text as it generates
                 if tokens.count % displayEveryNTokens == 0 {
--- a/Applications/LLMEval/ContentView.swift
+++ b/Applications/LLMEval/ContentView.swift
@@ -226,7 +226,7 @@ class LLMEvaluator {
 
             let result = await LLM.generate(
                 promptTokens: promptTokens, parameters: generateParameters, model: model,
-                tokenizer: tokenizer, configuration: modelConfiguration
+                tokenizer: tokenizer, extraEOSTokens: modelConfiguration.extraEOSTokens
             ) { tokens in
                 // update the output -- this will make the view show the text as it generates
                 if tokens.count % displayEveryNTokens == 0 {
--- a/Libraries/LLM/Evaluate.swift
+++ b/Libraries/LLM/Evaluate.swift
@@ -12,7 +12,7 @@ private func topPSampling(logits: MLXArray, topP: Float, temp: Float) -> MLXArra
         logits = logits.asType(.float32)
     }
 
-    let probs = softMax(logits / temp, axis: -1)
+    let probs = softmax(logits / temp, axis: -1)
     let sortedIndices = argSort(probs, axis: -1)
 
     // probs shape is [B,V] and after take it will be [1, B, V], so we squeeze it back to [B, V]
@@ -31,7 +31,7 @@ private func applyRepetitionPenalty(
 ) -> MLXArray {
     if repetitionContext.shape[0] > 0 {
         let indices = repetitionContext
-        var selectedLogits = take(logits, indices, axis: -1).squeezed(axis: 0)
+        var selectedLogits = logits[0..., indices]
 
         selectedLogits = MLX.where(
             selectedLogits .< 0, selectedLogits * penalty, selectedLogits / penalty)
@@ -100,7 +100,7 @@ public struct TokenIterator: Sequence, IteratorProtocol {
             if prompt.shape[0] <= parameters.repetitionContextSize {
                 self.repetitionContext = prompt
             } else {
-                self.repetitionContext = prompt[-parameters.repetitionContextSize ... -1]
+                self.repetitionContext = prompt[(-parameters.repetitionContextSize)...]
             }
         } else {
             self.repetitionContext = []
@@ -120,9 +120,8 @@ public struct TokenIterator: Sequence, IteratorProtocol {
         y = sample(logits: logits, temp: parameters.temperature, topP: parameters.topP)
         // append the current token to the context and check repetitionPenalty context see if need to remove the first token
         if parameters.repetitionContextSize > 1 {
-            repetitionContext = concatenated([repetitionContext, y], axis: 0)
             if repetitionContext.shape[0] > parameters.repetitionContextSize {
-                repetitionContext = repetitionContext[1...]
+                repetitionContext = repetitionContext[(-parameters.repetitionContextSize)...]
             }
         }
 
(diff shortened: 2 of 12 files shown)


Applications/LLMEval/ContentView.swift
```diff
@@ -125,6 +125,8 @@ struct ContentView: View {
 
         }
         .task {
+            self.prompt = llm.modelConfiguration.defaultPrompt
```
davidkoski, line 128: Use the new default prompt (in case the model changes)

Libraries/LLM/Evaluate.swift
```diff
@@ -12,7 +12,7 @@ private func topPSampling(logits: MLXArray, topP: Float, temp: Float) -> MLXArra
         logits = logits.asType(.float32)
     }
 
-    let probs = softMax(logits / temp, axis: -1)
+    let probs = softmax(logits / temp, axis: -1)
```
davidkoski, line 15: deprecation warning fix

Libraries/LLM/Evaluate.swift
```diff
@@ -120,9 +120,8 @@ public struct TokenIterator: Sequence, IteratorProtocol {
         y = sample(logits: logits, temp: parameters.temperature, topP: parameters.topP)
         // append the current token to the context and check repetitionPenalty context see if need to remove the first token
         if parameters.repetitionContextSize > 1 {
-            repetitionContext = concatenated([repetitionContext, y], axis: 0)
             if repetitionContext.shape[0] > parameters.repetitionContextSize {
-                repetitionContext = repetitionContext[1...]
+                repetitionContext = repetitionContext[(-parameters.repetitionContextSize)...]
```
davidkoski, line 124: Fix for #71 -- this uses a direct port of the python code.  Previously it looks like there were workarounds for the lack of full numpy array indexing, see `take()` above.

Libraries/LLM/Evaluate.swift
```diff
@@ -174,14 +173,32 @@ public enum GenerateDisposition {
...
+                tokenizer.encode(text: $0)
+            }
+            .filter {
+                // discard anything that is not a single token.  sometimes
+                // the tokenizer will insert a <s> token, so accept that too
+                $0.count == 1 || ($0.count == 2 && $0[0] == 1)
+            }
+            .map {
+                $0.last!
+            })
```
davidkoski, line 200: Fix for #74

Libraries/LLM/Load.swift
```diff
@@ -54,9 +54,15 @@ public func load(
...
 
+        // per-model cleanup
+        weights = model.sanitize(weights: weights)
+
         // quantize if needed
         if let quantization = baseConfig.quantization {
-            quantizeIfNeeded(model: model, weights: weights, quantization: quantization)
+            quantize(model: model, groupSize: quantization.groupSize, bits: quantization.bits) {
+                path, module in
+                weights["\(path).scales"] != nil
```
davidkoski, line 64: This is the fix for loading partially quantized models -- if they have "scales" then the layer is quantized.  We don't need to test the type of layer as the `quantize()` method will only convert layers that _can_ be converted.

Libraries/LLM/Load.swift
```diff
@@ -76,38 +82,3 @@ public func load(
             hub: hub, configuration: newConfiguration, progressHandler: progressHandler)
     }
 }
-
-// MARK: - Quantization
-
-private func quantizeIfNeeded(
-    model: LLMModel, weights: [String: MLXArray], quantization: BaseConfiguration.Quantization
-) {
```
davidkoski, line 84: No longer needed.

Libraries/LLM/Load.swift
davidkoski: Note that this depends on https://github.com/ml-explore/mlx-swift/pull/73

Libraries/LLM/Models.swift
```diff
@@ -98,11 +112,16 @@ public struct ModelConfiguration {
...
+
+        // https://www.promptingguide.ai/models/mistral-7b
+        defaultPrompt: "describe the swift language"
+    )
 
     public static let codeLlama13b4bit = ModelConfiguration(
         id: "mlx-community/CodeLlama-13b-Instruct-hf-4bit-MLX",
-        overrideTokenizer: "PreTrainedTokenizer"
+        overrideTokenizer: "PreTrainedTokenizer",
+        defaultPrompt: "func sortArray(_ array: [Int]) -> String { <FILL_ME> }"
```
davidkoski, line 124: I verified each of these models and added a defaultPrompt so I don't have to hunt around each time

Libraries/LLM/Models.swift
```diff
@@ -111,40 +130,53 @@ extension ModelConfiguration {
         "<PRE> " + prompt.replacingOccurrences(of: "<FILL_ME>", with: "<SUF>") + " <MID>"
     }
 
-    public static let phi4bit = ModelConfiguration(id: "mlx-community/phi-2-hf-4bit-mlx") {
-        prompt in
-        "Instruct: \(prompt)\nOutput: "
```
davidkoski, line 116: This wasn't giving good results (any more?) and isn't used on the python side.

Libraries/LLM/Qwen2.swift
```diff
@@ -249,6 +271,8 @@ public struct Qwen2Configuration: Codable {
                 Bool.self, forKey: Qwen2Configuration.CodingKeys.ropeTraditional) ?? false
         self.ropeScaling = try container.decodeIfPresent(
             [String: StringOrNumber].self, forKey: Qwen2Configuration.CodingKeys.ropeScaling)
+        self.tieWordEmbeddings =
+            try container.decodeIfPresent(Bool.self, forKey: .tieWordEmbeddings) ?? false
```
davidkoski, line 275: The qwen2 model implementation has changed slightly

davidkoski:
note: the CI tests are failing because this depends on the mlx-swift change that needs to merge first

solume:
tested it with mlx-swift fork [](https://github.com/ml-explore/mlx-swift/pull/73), this breaks for me with Libraries/LLM/Load.swift:62:13 Cannot find 'quantize' in scope
and
Libraries/LLM/Qwen2.swift:198:37 Value of type 'Embedding' has no member 'asLinear'

solume:
this still gives me an error even after linking the correct branch:
```
LLM >
Missing required module '_NumericsShims'
```

solume:
was able to fix dependencies but getting error:
`libc++abi: terminating due to uncaught exception of type std::invalid_argument: [matmul] Last dimension of first input with shape (1,42,1280) must match second to last dimension of second input with shape (160,32000).`

Libraries/LLM/Load.swift
awni: Nice simplification!

Libraries/LLM/Qwen2.swift
```diff
@@ -191,8 +193,26 @@ public class Qwen2Model: Module, LLMModel {
...
+        }
+        return (out, cache)
+    }
+
+    public func sanitize(weights: [String: MLXArray]) -> [String: MLXArray] {
+        var weights = weights
+
+        if configuration.tieWordEmbeddings && weights["lm_head.weight"] == nil {
+            weights["lm_head.weight"] = nil
+        }
```
awni, line 210: This looks kind of odd. If it's equal to `nil`, why set it to `nil`? Is that a Swift thing?

Libraries/LLM/Models.swift
awni: Beautiful, thanks!

Libraries/LLM/Evaluate.swift
```diff
@@ -174,14 +173,32 @@ public enum GenerateDisposition {
 ///   - parameters: generation parameters
 ///   - model: model to evaluate
 ///   - tokenizer: tokenizer to convert tokens back into strings and recognizer special tokens
+///   - configuration: the model configuration
 ///   - didGenerate: visitor for the tokens as they are generated
 public func generate(
     promptTokens: [Int], parameters: GenerateParameters, model: LLMModel, tokenizer: Tokenizer,
+    configuration: ModelConfiguration,
```
awni, line 180: I'm not certain about the way this model configuration gets passed around and what the intention for it is. Somehow we don't have a need for it in Python, so I'm wondering why do you need it here? What should go in it vs in the tokenizer/model directly?

From what I can tell it's more like the default arguments for a given model (eos token / prompt). The prompt gets handled outside this function. So is here it's just for the additional eos tokens?

Libraries/LLM/Qwen2.swift
davidkoski: No, you are right that is odd.

davidkoski: Here is the original python:

```python
        if self.args.tie_word_embeddings:
            weights.pop("lm_head.weight", None)
```

I think I can just remove the `&& ...`

Libraries/LLM/Evaluate.swift
davidkoski: yeah, just for the additional eos tokens.  I can switch that to just pass in additional eos tokens (optional).  If we need more parameters in the future I can rethink that slightly.

awni (approved):
Thanks!! LGTM

PR #78, opened 2024-05-28
Fix download progress

DePasqualeOrg:


Libraries/LLM/Load.swift
```diff
@@ -26,7 +26,7 @@ public func load(
         case .id(let id):
             // download the model weights and config
             let repo = Hub.Repo(id: id)
-            let modelFiles = ["config.json", "*.safetensors"]
+            let modelFiles = ["*.safetensors"]
```
davidkoski, line 29: Does this download the config.json if it is missing?  I wonder if we need two calls to snapshot?

DePasqualeOrg: I just tested it, and it works. In my own app, I download the tokenizer files and then the model files separately.

davidkoski (approved):
Nice idea, thank you for the contribution!

PR #83, opened 2024-06-16
Prepare for strict concurrency in Swift 6

DePasqualeOrg:
This is a starting point for preparing for strict concurrency in Swift 6. There are still lots of errors and warnings, and I don't know how to fix them all, but perhaps someone more knowledgeable than me can take it from here.

### DIFF
--- a/Libraries/LLM/LLMModel.swift
+++ b/Libraries/LLM/LLMModel.swift
@@ -5,7 +5,7 @@ import MLX
 import MLXNN
 
 // Interface for all LLM Models
-public protocol LLMModel: Module {
+public protocol LLMModel: Module, Sendable {
 
     var vocabularySize: Int { get }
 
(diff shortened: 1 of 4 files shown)


Libraries/LLM/LLMModel.swift
```diff
@@ -5,7 +5,7 @@ import MLX
 import MLXNN
 
 // Interface for all LLM Models
-public protocol LLMModel: Module {
+public protocol LLMModel: Module, Sendable {
```
davidkoski, line 8: I don't think this is really Sendable -- the model is full of MLXArrays, which are mutable and very much _not_ sendable.

We might be able to use `@unchecked Sendable` and document that callers have to be very very careful, but that kind of defeats the purpose of `Sendable`.  I think I need to read the docs on what is the intent -- I thought that swift 6 had some notion of being able to tell that things were not used concurrently.

davidkoski: I have been playing around with this a bit and I think we need to take a different approach:  the model (a Module) can't be Sendable but it can be wrapped in a Sendable type that guards access.  We need the async evaluation and the innards are definitely not thread safe.

davidkoski:
@DePasqualeOrg check out https://github.com/ml-explore/mlx-swift-examples/pull/90 -- this is a second take on it where I wrap the model in an actor (since marking it as Sendable isn't going to work / not correct).

PR #88, opened 2024-06-27
Add Gemma 2

DePasqualeOrg:
I tried using ChatGPT 4o to port the [Python implementation of Gemma 2](https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/models/gemma2.py) over to Swift based on the Python and Swift implementations of the previous Gemma model. It's crashing at `matmul` due to a wrong tensor shape. If this isn't useful, just ignore it, but maybe there's just a small change that needs to be made.

### DIFF
--- a/Libraries/LLM/Configuration.swift
+++ b/Libraries/LLM/Configuration.swift
@@ -32,6 +32,7 @@ public enum ModelType: String, Codable {
     case phi
     case phi3
     case gemma
+    case gemma2
     case qwen2
     case starcoder2
     case cohere
@@ -55,6 +56,10 @@ public enum ModelType: String, Codable {
             let configuration = try JSONDecoder().decode(
                 GemmaConfiguration.self, from: Data(contentsOf: configuration))
             return GemmaModel(configuration)
+        case .gemma2:
+            let configuration = try JSONDecoder().decode(
+                GemmaConfiguration.self, from: Data(contentsOf: configuration))
+            return Gemma2Model(configuration)
         case .qwen2:
             let configuration = try JSONDecoder().decode(
                 Qwen2Configuration.self, from: Data(contentsOf: configuration))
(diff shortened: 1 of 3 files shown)


davidkoski:
This is almost certainly an issue with the ported code.

One technique that I have used in the past is to diff the python models for something similar, perhaps Gemma and gemma2.  Then copy the swift implementation for Gemma and make the changes (hopefully small).

If you were starting from scratch I would suggest adding prints to the python and swift code and look at the shapes as it runs.  One or more of the statements wasn't translated correctly -- personally I hit misplaced parenthesis quite a few times.

DePasqualeOrg:
I followed your advice and diffed the Python model code for Gemma and Gemma 2. There are only a few lines that needed to be changed in `TransformerBlock`, so I copied the Gemma model and made those changes. However, it's still crashing at `matmul` with the same error. I'm using the same model configuration as Gemma. Does anything need to change there? I think it's likely that just a small change is needed to make this work.

DePasqualeOrg:
Here are the differences between Gemma and Gemma 2's config.json, in case that's helpful: [https://www.diffchecker.com/WoOxSiI6/](https://www.diffchecker.com/WoOxSiI6/)

DePasqualeOrg:
I managed to get Gemma 2 working. Since much of the code is the same as the previous Gemma model, I reused what I could without duplicating it. To test it out, just replace this in `ContentView`:

`let modelConfiguration = ModelConfiguration.phi34bit`

with this:

`let modelConfiguration = ModelConfiguration.gemma_2_9b_it_4bit`

davidkoski:
Looks like the swift-format check failed -- please run:

```
pre-commit run --all-files
```

davidkoski:
Nice!  This looks great!

DePasqualeOrg:
> Looks like the swift-format check failed -- please run:
> 
> ```
> pre-commit run --all-files
> ```

It should work now.

davidkoski (approved):
Thank you for the contribution!

Issue #89, opened 2024-07-02
Gemma 2 output includes `<end_of_turn>`


DePasqualeOrg:
I noticed that the output from the Gemma 2 models always ends with `<end_of_turn>`. The Gemma 1.1 models don't have this problem. Any idea what could be causing it?

awni:
Can you give a prompt / settings to reproduce that?

DePasqualeOrg:
You can reproduce it with the LLMEval example app in this repo with any prompt and the default settings. Just set the model configuration like this:

```swift
let modelConfiguration = ModelConfiguration.gemma_2_9b_it_4bit
```

DePasqualeOrg:
Sorry, make sure to try a short prompt, like "1+1=".

awni:
Going by this: https://github.com/google/gemma.cpp/issues/276

It seems like that is expected now and needs to be processed by the user.

awni:
I don't think we should special case it in the example since it's very idiosyncratic, so I will close this. It should be fairly straightforward to trim the response if the last token is `<end_of_turn>`

davidkoski:
The evaluate function can take extra tokens to stop at (`extraEOSTokens`):

```
public func generate(
    promptTokens: [Int], parameters: GenerateParameters, model: LLMModel, tokenizer: Tokenizer,
    extraEOSTokens: Set<String>? = nil,
    didGenerate: ([Int]) async -> GenerateDisposition
```

and the model configuration holds the same (and it will be passed in with the current integrations:

```
    public static let phi34bit = ModelConfiguration(
        id: "mlx-community/Phi-3-mini-4k-instruct-4bit-no-q-embed",
        defaultPrompt: "what is the gravity on mars and the moon?",
        extraEOSTokens: ["<|end|>"]
```

DePasqualeOrg:
I tried using this before opening this issue, but it didn't work in this case.

DePasqualeOrg:
[...]

PR #90, opened 2024-07-03
attempt 2 at preparing for strict concurrency

davidkoski:
- see also https://github.com/ml-explore/mlx-swift-examples/pull/83
- this marks many things in Sendable (which I think we can take regardless)
- creates an actor container for models and tokenizers, which are not Sendable (though perhaps Tokenizers could be)

> Note: this is really just LLMEval at this point to see if people like the direction.  `llmtool` in particular requires an updated `swift-argument-parser` but `swift-transformers` is pinning it at 1.3.0 right now.

Fixed, thanks!  https://github.com/huggingface/swift-transformers/issues/105

I think this is complete as a first pass.  Everything builds and runs on Xcode 15.4 with full concurrency checks.  There are a couple of warnings but I think they may be resolved with the new swift compiler in Xcode 16.

This may not fix everything but I think it is a good baseline.  I think wrapping models in `actor` to use them across threads is probably the way to go -- MLXArray and the evaluation graph are single threaded and must be protected.  Doing it like this proves to the compiler that they are so we get all the available checks.


Applications/LLMEval/ContentView.swift
```diff
@@ -148,9 +148,9 @@ struct ContentView: View {
 }
 
 @Observable
+@MainActor
```
davidkoski, line 151: Making the whole model MainActor solves a ton of problems

Applications/LLMEval/ContentView.swift
```diff
@@ -172,91 +172,87 @@ class LLMEvaluator {
...
-        case loaded(LLMModel, Tokenizers.Tokenizer)
+        case loaded(ModelContainer)
     }
 
     var loadState = LoadState.idle
 
     /// load and return the model -- can be called multiple times, subsequent calls will
     /// just return the loaded model
-    func load() async throws -> (LLMModel, Tokenizers.Tokenizer) {
+    func load() async throws -> ModelContainer {
```
davidkoski, line 182: This is now returning an actor, `ModelContainer`

Applications/LLMEval/ContentView.swift
```diff
@@ -172,91 +172,87 @@ class LLMEvaluator {
...
-            let (model, tokenizer) = try await load()
+            let modelContainer = try await load()
+
             // augment the prompt as needed
             let prompt = modelConfiguration.prepare(prompt: prompt)
-            let promptTokens = tokenizer.encode(text: prompt)
+
+            let promptTokens = await modelContainer.perform { _, tokenizer in
+                tokenizer.encode(text: prompt)
+            }
```
davidkoski, line 220: We have to use the actor to make sure we have isolated access to the state.

Libraries/LLM/Evaluate.swift
```diff
@@ -178,24 +178,16 @@ public enum GenerateDisposition {
 public func generate(
     promptTokens: [Int], parameters: GenerateParameters, model: LLMModel, tokenizer: Tokenizer,
     extraEOSTokens: Set<String>? = nil,
-    didGenerate: ([Int]) async -> GenerateDisposition
-) async -> GenerateResult {
+    didGenerate: ([Int]) -> GenerateDisposition
+) -> GenerateResult {
```
davidkoski, line 182: This would be an API change, so examine carefully!  If this is a problem I can look at doing this a different way, but I think it will be more complicated in the `ModelContainer` in order to support async actions.

Libraries/LLM/Evaluate.swift
```diff
@@ -178,24 +178,16 @@ public enum GenerateDisposition {
...
-            }
-            .filter {
-                // discard anything that is not a single token.  sometimes
-                // the tokenizer will insert a <s> token, so accept that too
-                $0.count == 1 || ($0.count == 2 && $0[0] == 1)
-            }
-            .map {
-                $0.last!
+            .compactMap {
+                tokenizer.convertTokenToId($0)
```
davidkoski, line 190: This is actually unrelated -- I was messing around with https://github.com/ml-explore/mlx-swift-examples/issues/89

Probably need to revert this and take it in a separate PR

Libraries/LLM/LLMModel.swift
```diff
@@ -3,8 +3,48 @@
...
+/// let result = await modelContainer.perform { model, tokenizer in
+///     LLM.generate(
+///         promptTokens: promptTokens, parameters: generateParameters, model: model,
+///         tokenizer: tokenizer, extraEOSTokens: modelConfiguration.extraEOSTokens
+///     ) { tokens in
+///     ...
+///     }
+/// }
+/// ```
+public actor ModelContainer {
```
davidkoski, line 31: This is the most important change -- something that we can use to guard the non-sendable data like models.

Tools/LinearModelTraining/LinearModelTraining.swift
```diff
@@ -7,7 +7,7 @@ import MLXNN
 import MLXOptimizers
 import MLXRandom
 
-extension MLX.DeviceType: ExpressibleByArgument {
+extension MLX.DeviceType: @retroactive ExpressibleByArgument {
```
davidkoski, line 10: Hrm, that seems to be a swift6 compiler feature -- have to figure out how to express this so that older compilers will accept it.

DePasqualeOrg:
This looks great, thanks! I was able to update my app for the new API fairly easily. However, when I try to build LLMEval in this repo, I still see some concurrency warnings.

davidkoski:
> This looks great, thanks! I was able to update my app for the new API fairly easily. However, when I try to build LLMEval in this repo, I still see some concurrency warnings.

Yeah, there may still be some -- I will keep at this.  It seems tricky as some of the fixes don't work on older swift compilers (the `@retroactive` markers for example).

Applications/LLMEval/ViewModels/DeviceStat.swift
```diff
@@ -3,33 +3,22 @@ import LLM
 import MLX
 
 @Observable
-class DeviceStat {
+final class DeviceStat: @unchecked Sendable {
```
davidkoski, line 6: `gpuUsage` is the only mutable state and it is `@MainActor`

Applications/LLMEval/ViewModels/DeviceStat.swift
```diff
@@ -3,33 +3,22 @@ import LLM
...
+
+    @MainActor
     var gpuUsage = GPU.snapshot()
-    private var initialGPUSnapshot = GPU.snapshot()
+
+    private let initialGPUSnapshot = GPU.snapshot()
     private var timer: Timer?
```
davidkoski, line 12: Although this is a var it is only that way for the purpose of `init()` and the need to capture self.

Applications/LoRATrainingExample/ContentView.swift
```diff
@@ -131,30 +133,29 @@ class LoRAEvaluator {
     private let evaluateShowEvery = 8
     private let maxTokens = 200
 
-    private func loadModel() async throws -> (LLMModel, Tokenizer) {
+    private func loadModel() async throws -> ModelContainer {
```
davidkoski, line 136: This is the same sort of transformation that LLMEval has -- use the new actor wrapper.

Applications/LoRATrainingExample/ContentView.swift
```diff
@@ -199,45 +209,47 @@ class LoRAEvaluator {
...
-            await MainActor.run {
-                switch progress {
-                case .train(let i, _, _, _):
-                    self.progress = .init(
-                        title: "Train", current: Double(i), limit: Double(parameters.iterations))
-                case .validation:
-                    output += "\n"
-                default:
-                    break
+        try await modelContainer.perform { model, tokenizer in
```
davidkoski, line 212: These changes are similar to the ones in LoraCommands (command line version of this)

Applications/MNISTTrainer/ContentView.swift
```diff
@@ -58,19 +58,33 @@ struct ContentView: View {
     }
 }
 
+@MainActor
 @Observable
-class Trainer {
+class ModelState {
```
davidkoski, line 63: Renamed as it holds messages and the model indirectly.

Applications/MNISTTrainer/ContentView.swift
```diff
@@ -58,19 +58,33 @@ struct ContentView: View {
...
+    }
+}
+
+actor LeNetContainer {
+
+    private let model = LeNet()
+
+    let mnistImageSize: CGSize = CGSize(width: 28, height: 28)
+
+    func train(output: ModelState) async throws {
```
davidkoski, line 87: This gets wrapped in an actor -- similar to the one we use for LLMs

Libraries/LLM/Evaluate.swift
davidkoski: I propose doing a major version bump for this so we don't surprise anyone

Package.swift
```diff
@@ -15,8 +15,8 @@ let package = Package(
             targets: ["MLXMNIST"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/ml-explore/mlx-swift", from: "0.12.1"),
-        .package(url: "https://github.com/huggingface/swift-transformers", from: "0.1.8"),
+        .package(url: "https://github.com/ml-explore/mlx-swift", from: "0.16.0"),
+        .package(url: "https://github.com/huggingface/swift-transformers", from: "0.1.9"),
```
davidkoski, line 19: Pick up new versions

Tools/LinearModelTraining/LinearModelTraining.swift
```diff
@@ -7,11 +7,19 @@ import MLXNN
 import MLXOptimizers
 import MLXRandom
 
-extension MLX.DeviceType: ExpressibleByArgument {
-    public init?(argument: String) {
-        self.init(rawValue: argument)
+#if swift(>=6.0)
+    extension MLX.DeviceType: @retroactive ExpressibleByArgument {
```
davidkoski, line 11: This `@retroactive` is only available in swift 6 compiler

Tools/llm-tool/LLMTool.swift
```diff
@@ -9,35 +9,36 @@ import Tokenizers
 
 @main
 struct LLMTool: AsyncParsableCommand {
-    static var configuration = CommandConfiguration(
+    static let configuration = CommandConfiguration(
```
davidkoski, line 12: No need for these to be mutable

mlx-swift-examples.xcodeproj/project.pbxproj
```diff
@@ -1128,6 +1128,7 @@
 				CODE_SIGN_STYLE = Automatic;
 				COPY_PHASE_STRIP = NO;
 				CURRENT_PROJECT_VERSION = 1;
+				DEAD_CODE_STRIPPING = YES;
```
davidkoski, line 1131: Per Xcode suggestion

mlx-swift-examples.xcodeproj/project.pbxproj
```diff
@@ -1176,6 +1177,7 @@
 				SWIFT_ACTIVE_COMPILATION_CONDITIONS = "DEBUG $(inherited)";
 				SWIFT_EMIT_LOC_STRINGS = YES;
 				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
+				SWIFT_STRICT_CONCURRENCY = complete;
```
davidkoski, line 1180: Turns on more aggressive warnings

mlx-swift-examples.xcodeproj/project.pbxproj
```diff
@@ -2583,8 +2674,8 @@
 			isa = XCRemoteSwiftPackageReference;
 			repositoryURL = "https://github.com/gonzalezreal/swift-markdown-ui";
 			requirement = {
-				branch = main;
-				kind = branch;
+				kind = upToNextMajorVersion;
+				minimumVersion = 2.3.1;
```
davidkoski, line 2678: Pick up build fixes for Xcode 16

DePasqualeOrg:
Fantastic, thank you! I just tried this out with my app, and all the concurrency errors are now resolved.

awni (approved):
Looks good, thanks for the update!!

PR #91, opened 2024-07-03
Fix extra EOS tokens

DePasqualeOrg:
https://github.com/ml-explore/mlx-swift-examples/issues/89#issuecomment-2207402261


davidkoski (approved):
Looks good!

PR #95, opened 2024-07-16
Add SmolLM

DePasqualeOrg:
I tried to get the SmolLM models working in Swift, but the output is garbled:

> conclusive Request harbourEntry Wizardlr speculationoglyph screw Februaryaco augmented explos Rexask Transaction acetyl manually Unique enzymesanan dirtyExplain LikDR anybody luxuryblue Buildings Examine grew Fed legacyarring frontiers pronounced sprinkle ...

It looks like there's a problem with the tokenizer. These models work fine in the Python MLX library. Any idea what changes would be needed to make this work?

### DIFF
--- a/Libraries/LLM/Models.swift
+++ b/Libraries/LLM/Models.swift
@@ -110,6 +110,13 @@ public struct ModelConfiguration {
 }
 
 extension ModelConfiguration {
+    public static let smolLM_135M_4bit = ModelConfiguration(
+        id: "mlx-community/SmolLM-135M-Instruct-4bit",
+        defaultPrompt: "Tell me about the history of Spain."
+    ) {
+        prompt in
+        "<|im_start|>user\n\(prompt)<|im_end|>\n<|im_start|>assistant\n"
+    }
 
     public static let mistral7B4bit = ModelConfiguration(
         id: "mlx-community/Mistral-7B-v0.1-hf-4bit-mlx",
--- a/Libraries/LLM/Llama.swift
+++ b/Libraries/LLM/Llama.swift
@@ -180,19 +180,25 @@ public class LlamaModel: Module, LLMModel {
     public let vocabularySize: Int
     let model: LlamaModelInner
 
-    @ModuleInfo(key: "lm_head") var lmHead: Linear
+    @ModuleInfo(key: "lm_head") var lmHead: Linear?
 
     public init(_ args: LlamaConfiguration) {
         self.vocabularySize = args.vocabularySize
         self.model = LlamaModelInner(args)
-        self._lmHead.wrappedValue = Linear(args.hiddenSize, args.vocabularySize, bias: false)
+        if !args.tieWordEmbeddings {
+            self._lmHead.wrappedValue = Linear(args.hiddenSize, args.vocabularySize, bias: false)
+        }
     }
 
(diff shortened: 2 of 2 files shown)


davidkoski:
Not off the top of my head, but I would probably compare:

- tokens from the prompt (numeric)
- try using the numeric tokens from the python side
    - this will tell you if it is on the input side or the output side
- output tokens (numeric)

If the input tokens are different and copying the tokens from the python side works, then it is probably an issue for `swift-transformers`.

If the input tokens match or copying the tokens from the python side doesn't work then look at the output tokens.  If those don't match then it is a problem with the model evaluation (and make sure top-p and temperature match) and we can debug that.

If the output tokens match then I think `swift-transforms` will be able to convert it back into legible text -- there isn't anything tricky there.

DePasqualeOrg:
Thanks! I'll be on vacation for the next couple weeks, but perhaps someone can use this as a starting point, or I'll get to it later on.

awni:
Could be related to tying the word embeddings: https://huggingface.co/mlx-community/SmolLM-360M-Instruct-fp16/blob/main/config.json#L24

I don't see support for that in the Llama model file

Libraries/LLM/Models.swift
```diff
@@ -110,6 +110,13 @@ public struct ModelConfiguration {
 }
 
 extension ModelConfiguration {
+    public static let smolLM_135M_4bit = ModelConfiguration(
+      id: "mlx-community/SmolLM-135M-Instruct-4bit",
```
awni, line 114: Also I would try the 16-bit to start just to make sure the bad output is not related to quantization.

DePasqualeOrg:
Thanks for that tip. Handling `tie_word_embeddings` now makes it produce intelligible output, but it seems that a different chat template from the current Llama template is required.

Libraries/LLM/Models.swift
```diff
@@ -110,6 +110,13 @@ public struct ModelConfiguration {
 }
 
 extension ModelConfiguration {
+    public static let smolLM_135M_4bit = ModelConfiguration(
```
awni, line 113: I noticed the 4-bit model is pretty low quality. It might be better to default this to the 8bit (or 16bit).

awni (approved):
Looks good to me! Thanks for the addition. I'll let @davidkoski merge when ready.

Libraries/LLM/Models.swift
DePasqualeOrg: Even the 1.7B instruction-tuned model at 8-bit doesn't work well for me. Maybe they didn't do enough fine tuning?

Libraries/LLM/Models.swift
```diff
@@ -110,6 +110,13 @@ public struct ModelConfiguration {
...
+        id: "mlx-community/SmolLM-135M-Instruct-4bit",
+        defaultPrompt: "Tell me about the history of Spain."
+    ) {
+        prompt in
+        "<|im_start|>user\n\(prompt)<|im_end|>\n<|im_start|>assistant\n"
```
diegoasua, line 118: Is this correct? I am having issues with the tokenizer for this model Ollama seems to have figured out:

https://github.com/ollama/ollama/issues/5731#issuecomment-2233890666

davidkoski: That looks like the same template they are using.  What problem are you seeing?

PR #98, opened 2024-07-24
Add Llama 3.1

DePasqualeOrg:
Based on https://github.com/ml-explore/mlx-examples/pull/907

I have verified that this works with Llama 3.1 8B, Llama 3 8B, Mistral NeMo, and Mistral 7B.

I don't fully understand some of these changes, so please check them carefully.

You can test Llama 3.1 8B like this:

```swift
let modelConfiguration = ModelConfiguration.llama3_1_8B_4bit
```

### DIFF
--- a/Libraries/LLM/Llama.swift
+++ b/Libraries/LLM/Llama.swift
@@ -9,7 +9,7 @@ import MLXNN
 
 private class DynamicNTKScalingRoPE: Module {
     let dims: Int
-    let maxPositionEmbeddings: Int
+    let maxPositionEmbeddings: Int?
     let traditional: Bool
     let originalBase: Float
     var scale: Float
@@ -17,7 +17,7 @@ private class DynamicNTKScalingRoPE: Module {
     let ropeScaling: [String: StringOrNumber]?
 
     init(
-        dims: Int, maxPositionEmbeddings: Int = 2048, traditional: Bool = false,
+        dims: Int, maxPositionEmbeddings: Int?, traditional: Bool = false,
         base: Float = 10000, scale: Float = 1.0, ropeType: String = "default",
         ropeScaling: [String: StringOrNumber]? = nil
     ) {
@@ -76,7 +76,7 @@ private class DynamicNTKScalingRoPE: Module {
     func callAsFunction(_ x: MLXArray, offset: Int = 0) -> MLXArray {
         let seqLen = x.dim(1) + offset
         var base = computeBaseFrequency()
-        if maxPositionEmbeddings > 0 && seqLen > maxPositionEmbeddings {
+        if let maxPositionEmbeddings, maxPositionEmbeddings > 0, seqLen > maxPositionEmbeddings {
             let factorAdjustment = Float(seqLen) / Float(maxPositionEmbeddings) - 1
             let dimensionRatio = Float(dims) / Float(Float(dims) - 2)
             let adjustedScale = scale * pow(1 + factorAdjustment, dimensionRatio)
@@ -116,7 +116,7 @@ private class Attention: Module {
 
         self.rope = DynamicNTKScalingRoPE(
             dims: headDim,
-            maxPositionEmbeddings: args.maxPositionEmbeddings ?? 2048,
+            maxPositionEmbeddings: args.maxPositionEmbeddings,
             traditional: args.ropeTraditional,
             base: args.ropeTheta,
             scale: 1.0,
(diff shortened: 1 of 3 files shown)


Libraries/LLM/Llama.swift
```diff
@@ -7,6 +7,86 @@ import MLXNN
...
+                0, min(1, (wavelen - highFreqWavelen) / (lowFreqWavelen - highFreqWavelen)))
+            return freq * ((1 - smooth) * factor + smooth)
+        }
+
+        return newBaseFreqs.reduce(0, +) / Float(newBaseFreqs.count)
+    }
+
+    func callAsFunction(_ x: MLXArray, offset: Int = 0) -> MLXArray {
+        let seqLen = x.dim(1) + offset
+        var base = computeBaseFrequency()
```
awni, line 78: This should really be done once on the module init.

Libraries/LLM/Models.swift
```diff
@@ -192,9 +199,17 @@ extension ModelConfiguration {
...
 
+    public static let llama3_1_8B_4bit = ModelConfiguration(
+        id: "mlx-community/Meta-Llama-3.1-8B-Instruct-4bit",
+        defaultPrompt: "What is the difference between a fruit and a vegetable?"
+    ) {
+        prompt in
+        "<|begin_of_text|><|start_header_id|>system<|end_header_id|>\nYou are a helpful assistant<|eot_id|>\n<|start_header_id|>user<|end_header_id|>\n\(prompt)<|eot_id|>\n<|start_header_id|>assistant<|end_header_id|>"
+    }
+
     public static let llama38B4bit = ModelConfiguration(
```
awni, line 210: Unrelated to this PR but could we change this to `llama3_8B_4bit`. It reads like `38B` right now which is confusing

Libraries/LLM/Llama.swift
```diff
@@ -17,43 +97,40 @@ private class Attention: Module {
...
-            dimensions: headDim, traditional: args.ropeTraditional, base: args.ropeTheta,
-            scale: ropeScale)
+        self._wq.wrappedValue = Linear(dim, heads * headDim, bias: args.attentionBias)
+        self._wk.wrappedValue = Linear(dim, kvHeads * headDim, bias: args.attentionBias)
+        self._wv.wrappedValue = Linear(dim, kvHeads * headDim, bias: args.attentionBias)
+        self._wo.wrappedValue = Linear(heads * headDim, dim, bias: args.attentionBias)
+
+        self.rope = DynamicNTKScalingRoPE(
+            dims: headDim,
+            maxPositionEmbeddings: args.maxPositionEmbeddings ?? 2048,
```
awni, line 119: Maybe it's better to have `maxPositionEmbeddings` be an optional value in `DynamicNTKScalingRoPE` instead of arbitrarily setting it to 2048 (which I think is wrong). Also because you test it for `> 0` so we will scale the `base` for models that we don't want to scale for.

Libraries/LLM/Llama.swift
```diff
@@ -7,6 +7,86 @@ import MLXNN
...
+            return freq * ((1 - smooth) * factor + smooth)
+        }
+
+        return newBaseFreqs.reduce(0, +) / Float(newBaseFreqs.count)
+    }
+
+    func callAsFunction(_ x: MLXArray, offset: Int = 0) -> MLXArray {
+        let seqLen = x.dim(1) + offset
+        var base = computeBaseFrequency()
+        if maxPositionEmbeddings > 0 && seqLen > maxPositionEmbeddings {
```
awni, line 79: Probably it should be optional and we should do something like `maxPositionEmbeddings != nil && seqLen > maxPositionEmbeddings!`

Libraries/LLM/Llama.swift
DePasqualeOrg: Are you sure? Because `base` is later mutated in `callAsFunction`.

awni: `computeBaseFrequency` only depends on `originalBase`. So another way to do this is to not even have an `originalBase`. Just set the class variable `base` at initialization and using the `base` provided in the init arguments. That should only be set once. Then modify a local copy in `callAsFunction` (if the condition is true).

DePasqualeOrg: Maybe you can go ahead and make those changes, since I might not do this right if I try it myself.

awni: Sure I'll push an update in a minute.

awni: The change makes a big difference in toks/sec, from 9 to 17 toks/sec on my iphone 15.

awni: Ok I sent the change.

davidkoski:
@DePasqualeOrg Conflicts with previous PR, please resolve. Thanks!

PR #99, opened 2024-07-31
Some fixes for gemma2

awni:
- Add attention and logit soft capping
- Add correct scalar for attention score scale


davidkoski (approved):
Looks good!

PR #100, opened 2024-08-02
Remove AsyncAlgorithms package

DePasqualeOrg:
Correct me if I'm wrong, but it seems that AsyncAlgorithms is not needed. LLMEval builds without it.


mlx-swift-examples.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -83,5 +65,5 @@
       }
     }
   ],
-  "version" : 2
+  "version" : 3
```
davidkoski, line 68: This update has caused problems for people in the past -- can you remove this line?

davidkoski:
Yes, you are correct -- I think this is just leftovers from one of the early versions that had an async sequence as a generator type.

mlx-swift-examples.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
DePasqualeOrg: Done.

davidkoski (approved):
Looks good, thank you!

PR #103, opened 2024-08-10
fix #102 -- extra eval of non-model/mlxarray data

### LINKED ISSUE #102: MLX/Transforms+Eval.swift:166: Fatal error: Unable to extract MLXArray from idle
I can't figure what I'm doing wrong to get this error. It occurs with both downloaded and locally saved models. Tried these 3 `ModelConfiguration` s: 
```
gemma2bQuantized
mistral7B4bit
phi3_4bit
```

davidkoski:


Applications/LoRATrainingExample/ContentView.swift
```diff
@@ -150,7 +150,6 @@ class LoRAEvaluator {
                         limit: 1.0)
                 }
             }
-            eval(model)
```
davidkoski, line 153: This shouldn't have been there -- the load of the model already eval'd it and this model is actually an enum so it hits an assert, see #102.

awni (approved):
Sorry I missed this PR. Thanks for the fix!

PR #108, opened 2024-08-22
refactor llm model load code to build inside the actor

davidkoski:
- this fixes the attempt to send non Sendable data between threads
- builds warning free with swift 5 (strict concurrency) / swift 6

### DIFF
--- a/Libraries/LLM/Load.swift
+++ b/Libraries/LLM/Load.swift
@@ -94,6 +94,8 @@ public func loadModelContainer(
     hub: HubApi = HubApi(), configuration: ModelConfiguration,
     progressHandler: @Sendable @escaping (Progress) -> Void = { _ in }
 ) async throws -> ModelContainer {
-    return try await ModelContainer(
+    let modelDirectory = try await prepareModelDirectory(
         hub: hub, configuration: configuration, progressHandler: progressHandler)
+    return try await ModelContainer(
+        hub: hub, modelDirectory: modelDirectory, configuration: configuration)
 }
--- a/Libraries/LLM/Load.swift
+++ b/Libraries/LLM/Load.swift
@@ -11,82 +11,89 @@ struct LLMError: Error {
     let message: String
 }
 
-/// Load and return the model and tokenizer
-public func load(
-    hub: HubApi = HubApi(), configuration: ModelConfiguration,
-    progressHandler: @Sendable @escaping (Progress) -> Void = { _ in }
-) async throws -> (LLMModel, Tokenizer) {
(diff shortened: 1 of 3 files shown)


Libraries/LLM/Load.swift
```diff
@@ -11,82 +11,91 @@ struct LLMError: Error {
...
 
-/// Load and return the model and tokenizer
-public func load(
-    hub: HubApi = HubApi(), configuration: ModelConfiguration,
-    progressHandler: @Sendable @escaping (Progress) -> Void = { _ in }
-) async throws -> (LLMModel, Tokenizer) {
+func prepareModelDirectory(
+    hub: HubApi, configuration: ModelConfiguration,
+    progressHandler: @Sendable @escaping (Progress) -> Void
+) async throws -> URL {
```
davidkoski, line 17: This is the only part of model loading that is asynchronous and it just produces a URL (with the downloaded weights).

Libraries/LLM/Load.swift
```diff
@@ -11,82 +11,91 @@ struct LLMError: Error {
...
 
-        // create the model (no weights loaded)
-        let configurationURL = modelDirectory.appending(component: "config.json")
-        let baseConfig = try JSONDecoder().decode(
-            BaseConfiguration.self, from: Data(contentsOf: configurationURL))
+/// Load and return the model and tokenizer
+public func load(
+    hub: HubApi = HubApi(), configuration: ModelConfiguration,
+    progressHandler: @Sendable @escaping (Progress) -> Void = { _ in }
+) async throws -> (LLMModel, Tokenizer) {
```
davidkoski, line 41: This is the original public method -- we keep it but the inside is refactored so that we can use the same code in the `actor` case.

Libraries/LLM/Load.swift
```diff
@@ -11,82 +11,91 @@ struct LLMError: Error {
...
-        newConfiguration.id = .directory(configuration.modelDirectory(hub: hub))
-        return try await load(
-            hub: hub, configuration: newConfiguration, progressHandler: progressHandler)
-    }
+    return model
 }
 
+/// Load and return the model and tokenizer wrapped in a ``ModelContainer`` (provides
+/// thread safe access).
 public func loadModelContainer(
```
davidkoski, line 93: Same factory method as before -- this loads a model and produces a `ModelContainer` (actor) that allows thread safe access to the model/tokenizer.

The interior has been redone to avoid passing non-Sendable data between threads/isolation contexts.

awni (approved):
Looks good!

Issue #93, opened 2024-07-10
Swift generates tokens substantially slower than python for Phi-3


neilmehta24:
The python `mlx_lm` implementation generates at ~101 tokens per second for `mlx-community/Phi-3-mini-4k-instruct-4bit`, whereas the swift code here generates at ~60 tokens per second.

Here is my python implementation
```python
from mlx_lm import load, generate

model, tokenizer = load("mlx-community/Phi-3-mini-4k-instruct-4bit")
response = generate(model, tokenizer, max_tokens=100000, prompt="""<|system|>
You are a helpful assistant.<|end|>
<|user|>
How to explain Internet for a medieval knight?<|end|>
<|assistant|>""", verbose=True)
```

Here is my swift command
```zsh
➜  mlx-swift-examples git:(main) ./mlx-run llm-tool eval --model mlx-community/Phi-3-mini-4k-instruct-4bit -m 100000 -p "<|system|>
You are a helpful assistant.<|end|>
<|user|>
How to explain Internet for a medieval knight?<|end|>
<|assistant|>"
```

Any ideas on how I can achieve similar speed in swift?

davidkoski:
I will investigate this

davidkoski:
I can see two things that look like they are contributing here:

- JIT startup costs -- the first time I ran the swift version I saw ~60 tokens per second, but subsequent runs were at 100 tokens per second
- `StreamingDetokenizer` -- if I generate 100 tokens the python and swift version are nearly the same speed (around 100 tokens per second on my laptop)
    - but if I generate 1000 tokens (and modify the various parameters so that it doesn't stop short) I get ~60 tokens per second from swift and still 100 tokens per second from python

In swift we are calling the Tokenizer with the entire list of output tokens and taking any additions to the resulting string to print out.  This is $O(n^2)$ performance as we need to scan $n$ tokens $n$ times.  The python version has `StreamingDetokenizer` that gets `O(n)` performance as it only generates the tail end of the output string as it runs -- we need a version of this in swift.

I suspect you were seeing the latter effect (you probably ran this a few times and the first effect was negligible).

davidkoski:
TASKs:

- port `StreamingDetokenizer`
- use `mx.async_eval(y)` to pipeline the generation
- look at `KVCache` from the python side as well

We can do these in that order as the first one is probably the biggest benefit.

awni:
> port StreamingDetokenizer

[...]

PR #109, opened 2024-08-23
add kvcache, async eval, etc for #93

davidkoski:
- sampling code compiled
- KVCache
- async_eval
- NaiveStreamingDetokenizer

My starting point with mlx v0.16.0:

```
./mlx-run llm-tool --model mlx-community/Phi-3-mini-4k-instruct-4bit -m 100000 --temperature 0 -p @/tmp/p.txt --cache-size 500 --temperature 0
```

Prompt is:

```
<|system|>
You are a helpful assistant.<|end|>
<|user|>
How to explain Internet for a medieval knight?<|end|>
<|assistant|>
```

Here are the initial performance numbers:

```
python Generation: 673 tokens, 114.404 tokens-per-sec, 5.873911874999999s
swift Generation: 673 tokens, 83.136954 tokens/s
```

Looking at the profiler we can see:

| Task | Python | Swift |
| --- | --- | --- |
| Wall Time | 9.6 | 12.1 |
| CPU Time | 10.94 | 9.91 |
| Metal Submit | 1.75 | 3.83 |
| Metal didComplete | 0.48 | 0.83 |
| mlx::core:scheduler | 2.92 | 3.63 |
| blas_thread_server | 2.9 | - |
| main | 1.67 | 0.74 |

I am not sure what the `blas_thread_server` is, but it is soaking up a bunch of CPU time — it looks like it is running async to the main computation, so other than efficiency isn’t affecting performance.  If we subtract the blas_thread_server from the CPU time it is clear the python code is doing less work.

I found the following:

- The sampling code in python is compiled — this made a pretty big difference
- The KVCache also made a big difference
- The async_eval in the python code helped a little
- The NaiveStreamingDetokenizer helped a little

The prompt splitting didn’t help here as the default split size is 512 and that is larger than the prompt.  The output in this case is only 673 tokens, so I don’t think that is particularly long, so some of these may not be getting the full benefit.

With these in place the performance is getting closer:

```
python Generation: 673 tokens, 114.404 tokens-per-sec, 5.873911874999999s
swift 673 tokens, 97.718867 tokens/s, 6.887104s
```

| Task | Python | Swift |
| --- | --- | --- |
| Wall Time | 9.6 | 9.4 |
| CPU Time | 10.94 | 7.6 |
| Metal Submit | 1.75 | 1.21 |
| Metal didComplete | 0.48 | 0.28 |
| mlx::core:scheduler | 2.92 | 1.99 |
| blas_thread_server | 2.9 | - |
| main | 1.67 | 0.69 |

We have a big win in CPU time and even total wall time (startup time I presume) but are still nearly a second slower on the generation, so there is still some more work needed.

# Part 2!

Here is the output from swift -- for comparison I had 114 tokens/s from python (though it was usually lower than that and I found the number to be noisy). 

```
Prompt:     21 tokens, 212.221945 tokens/s
Generation: 673 tokens, 114.662957 tokens/s, 5.869376s
```

Here is output from Instruments showing the synchronous evaluation (via `item()`), the `async_eval()` calls and the GPU activity in the python program:

image

Notice that the GPU stays active most of the time.

Here is a similar view of the swift program (at around 97 tokens/s) from part 1:

image

The pattern is similar, but there are ~1.5ms gaps where the GPU goes idle -- the scheduler didn't have any work to do.

The final version looks like this:

image

The GPU is kept busy and we are hitting our performance targets!

The final changes resolved an issue with the `asyncEval()` and `item()`/`eval()` calls that were causing `item()` to wait for the completion of the `asyncEval()` -- see https://github.com/ml-explore/mlx-swift/pull/127.

### DIFF
--- a/Libraries/LLM/Cohere.swift
+++ b/Libraries/LLM/Cohere.swift
@@ -37,8 +37,8 @@ private class Attention: Module {
     }
 
     public func callAsFunction(
-        _ x: MLXArray, mask: MLXArray? = nil, cache: (MLXArray, MLXArray)? = nil
-    ) -> (MLXArray, (MLXArray, MLXArray)) {
+        _ x: MLXArray, mask: MLXArray? = nil, cache: KVCache?
+    ) -> MLXArray {
         let (B, L) = (x.dim(0), x.dim(1))
 
         var queries = wq(x)
@@ -50,11 +50,10 @@ private class Attention: Module {
         keys = keys.reshaped(B, L, args.kvHeads, -1).transposed(0, 2, 1, 3)
         values = values.reshaped(B, L, args.kvHeads, -1).transposed(0, 2, 1, 3)
 
-        if let (keyCache, valueCache) = cache {
-            queries = rope(queries, offset: keyCache.dim(2))
-            keys = rope(keys, offset: keyCache.dim(2))
-            keys = concatenated([keyCache, keys], axis: 2)
-            values = concatenated([valueCache, values], axis: 2)
+        if let cache {
+            queries = rope(queries, offset: cache.offset)
+            keys = rope(keys, offset: cache.offset)
+            (keys, values) = cache.update(keys: keys, values: values)
         } else {
             queries = rope(queries)
             keys = rope(keys)
@@ -66,7 +65,7 @@ private class Attention: Module {
         .transposed(0, 2, 1, 3)
         .reshaped(B, L, -1)
 
-        return (wo(output), (keys, values))
+        return wo(output)
     }
 }
 
@@ -104,12 +103,12 @@ private class TransformerBlock: Module {
     }
 
     public func callAsFunction(
-        _ x: MLXArray, mask: MLXArray? = nil, cache: (MLXArray, MLXArray)? = nil
-    ) -> (MLXArray, (MLXArray, MLXArray)) {
+        _ x: MLXArray, mask: MLXArray? = nil, cache: KVCache?
+    ) -> MLXArray {
         let h = inputLayerNorm(x)
-        let (attnH, cache) = attention(h, mask: mask, cache: cache)
+        let attnH = attention(h, mask: mask, cache: cache)
         let ffH = mlp(h)
-        return (attnH + ffH + x, cache)
+        return attnH + ffH + x
     }
 }
 
@@ -133,48 +132,41 @@ public class CohereModelInner: Module {
         self.norm = LayerNorm(dimensions: args.hiddenSize, eps: args.layerNormEps)
     }
 
-    public func callAsFunction(_ inputs: MLXArray, cache: [(MLXArray, MLXArray)]? = nil) -> (
-        MLXArray, [(MLXArray, MLXArray)]
-    ) {
+    public func callAsFunction(_ inputs: MLXArray, cache: [KVCache]? = nil) -> MLXArray {
         var h = embedTokens(inputs)
 
-        var mask: MLXArray? = nil
-        if h.dim(1) > 1 {
-            mask = MultiHeadAttention.createAdditiveCausalMask(h.dim(1))
-            mask = mask?.asType(h.dtype)
-        }
-
-        var newCache = [(MLXArray, MLXArray)]()
+        let mask: MLXArray? = createAttentionMask(h: h, cache: cache)
 
         for (i, layer) in layers.enumerated() {
-            var cacheUpdate: (MLXArray, MLXArray)
-            (h, cacheUpdate) = layer(h, mask: mask, cache: cache?[i])
-            newCache.append(cacheUpdate)
+            h = layer(h, mask: mask, cache: cache?[i])
         }
 
-        return (norm(h), newCache)
+        return norm(h)
     }
 }
 
-public class CohereModel: Module, LLMModel {
+public class CohereModel: Module, LLMModel, KVCacheDimensionProvider {
 
     public let vocabularySize: Int
+    public let kvHeads: [Int]
+    public let headDim: IntOrPair
+
     let model: CohereModelInner
     let logitScale: Float
 
     public init(_ args: CohereConfiguration) {
         self.vocabularySize = args.vocabularySize
+        self.kvHeads = (0 ..< args.hiddenLayers).map { _ in args.kvHeads }
+        self.headDim = .init(args.hiddenSize / args.attentionHeads)
         self.model = CohereModelInner(args)
         self.logitScale = args.logitScale
     }
 
-    public func callAsFunction(_ inputs: MLXArray, cache: [(MLXArray, MLXArray)]?) -> (
-        MLXArray, [(MLXArray, MLXArray)]
-    ) {
-        var (out, cache) = model(inputs, cache: cache)
+    public func callAsFunction(_ inputs: MLXArray, cache: [KVCache]?) -> MLXArray {
+        var out = model(inputs, cache: cache)
         out = matmul(out, model.embedTokens.weight.T)
         out = out * self.logitScale
-        return (out, cache)
+        return out
     }
 }
 
--- a/Libraries/LLM/Evaluate.swift
+++ b/Libraries/LLM/Evaluate.swift
@@ -5,56 +5,12 @@ import MLX
 import MLXRandom
 import Tokenizers
 
-private func topPSampling(logits: MLXArray, topP: Float, temp: Float) -> MLXArray {
-    var logits = logits
-    if logits.dtype == .bfloat16 {
-        logits = logits.asType(.float32)
-    }
-
-    let probs = softmax(logits / temp, axis: -1)
-    let sortedIndices = argSort(probs, axis: -1)
-
-    // probs shape is [B,V] and after take it will be [1, B, V], so we squeeze it back to [B, V]
-    let sortedProbs = take(probs, sortedIndices, axis: -1).squeezed(axis: 0)
-
-    let cumulativeProbs = cumsum(sortedProbs, axis: -1)
-
-    let topProbs = MLX.where(cumulativeProbs .> (1 - topP), sortedProbs, zeros(like: sortedProbs))
-
-    let sortedToken = categorical(log(topProbs))
-    return sortedIndices.squeezed(axis: 0)[sortedToken]
-}
-
-private func applyRepetitionPenalty(
-    logits: MLXArray, repetitionContext: MLXArray, penalty: Float
-) -> MLXArray {
-    if repetitionContext.shape[0] > 0 {
-        let indices = repetitionContext
-        var selectedLogits = logits[0..., indices]
-
-        selectedLogits = MLX.where(
-            selectedLogits .< 0, selectedLogits * penalty, selectedLogits / penalty)
-
-        logits[0..., indices] = selectedLogits
-        return logits
-    }
-
-    return logits
-}
-
-private func sample(logits: MLXArray, temp: Float, topP: Float = 1.0) -> MLXArray {
-    if temp == 0 {
-        return argMax(logits, axis: -1)
-    } else {
-        if topP > 0 && topP < 1 {
-            return topPSampling(logits: logits, topP: topP, temp: temp)
-        }
-        return categorical(logits * (1 / temp))
-    }
-}
-
 /// Parameters for text generation, see ``TokenIterator``
 public struct GenerateParameters: Sendable {
+
+    /// Step size for processing the prompt
+    public var prefillStepSize = 512
+
     /// sampling temperature
     public var temperature: Float = 0.6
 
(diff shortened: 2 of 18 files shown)


Libraries/LLM/Cohere.swift
```diff
@@ -50,11 +50,10 @@ private class Attention: Module {
...
 
-        if let (keyCache, valueCache) = cache {
-            queries = rope(queries, offset: keyCache.dim(2))
-            keys = rope(keys, offset: keyCache.dim(2))
-            keys = concatenated([keyCache, keys], axis: 2)
-            values = concatenated([valueCache, values], axis: 2)
+        if let cache {
+            queries = rope(queries, offset: cache.offset)
+            keys = rope(keys, offset: cache.offset)
+            (keys, values) = cache.update(keys: keys, values: values)
```
davidkoski, line 56: There are a lot of very mechanical changes like this to support the new KVCache.  The KVCache is a reference type so we pass it in and mutate it rather than pass in/pass out.

Libraries/LLM/Cohere.swift
```diff
@@ -144,37 +141,36 @@ public class CohereModelInner: Module {
...
+        return norm(h)
     }
 }
 
-public class CohereModel: Module, LLMModel {
+public class CohereModel: Module, LLMModel, KVCacheDimensionProvider {
 
     public let vocabularySize: Int
+    public let kvHeads: [Int]
+    public let headDim: IntOrPair
```
davidkoski, line 156: Support for creating the KVCache

Libraries/LLM/Cohere.swift
```diff
@@ -144,37 +141,36 @@ public class CohereModelInner: Module {
...
-    public func callAsFunction(_ inputs: MLXArray, cache: [(MLXArray, MLXArray)]?) -> (
-        MLXArray, [(MLXArray, MLXArray)]
-    ) {
-        var (out, cache) = model(inputs, cache: cache)
+    public func callAsFunction(_ inputs: MLXArray, cache: [KVCache]?) -> MLXArray {
+        var out = model(inputs, cache: cache)
         out = matmul(out, model.embedTokens.weight.T)
         out = out * self.logitScale
-        return (out, cache)
+        return out
```
davidkoski, line 173: Every model is changed in this way -- they aren't all identical changes, but 70% are.  The ones that are not identical are just different because of the specifics in the attention layers.

Libraries/LLM/Evaluate.swift
```diff
@@ -5,56 +5,12 @@ import MLX
 import MLXRandom
 import Tokenizers
 
-private func topPSampling(logits: MLXArray, topP: Float, temp: Float) -> MLXArray {
```
davidkoski, line 8: These functions move inside the SampleContext for encapsulation.  They also get compiled to match the python side.

Libraries/LLM/Evaluate.swift
```diff
@@ -78,54 +34,186 @@ public struct GenerateParameters: Sendable {
...
+                return topPSampling(logits: logits)
+            } else {
+                return compiledCategorical(logits, temp)
+            }
+        }
+    }
+}
+
+/// Encapsulaton of the repetitionPenalty
+struct RepetitionContext: Sendable {
```
davidkoski, line 99: Refactored out the repetition context code for encapsulation -- I found that this was missing the token append line.

Libraries/LLM/Evaluate.swift
```diff
@@ -78,54 +34,186 @@ public struct GenerateParameters: Sendable {
...
         return y
     }
+
+    mutating public func next() -> MLXArray? {
+        // save current value -- this will be returned
+        let previousY = y
+
+        // compute the next state and async eval the next token
+        y = step(previous: previousY)
+        asyncEval(y)
```
davidkoski, line 213: This restructuring supports async eval now.  We generate the graph for the next token and start the eval while returning the previously generated token.

Libraries/LLM/Evaluate.swift
```diff
@@ -149,13 +237,13 @@ public struct GenerateResult: Sendable {
...
-        Double(tokens.count - 1) / generateTime
+        Double(tokens.count) / generateTime
     }
 
     public func summary() -> String {
         """
-        Prompt Tokens per second:     \(promptTokensPerSecond.formatted())
-        Generation tokens per second: \(tokensPerSecond.formatted())
+        Prompt:     \(promptTokens.count) tokens, \(promptTokensPerSecond.formatted()) tokens/s
+        Generation: \(tokens.count) tokens, \(tokensPerSecond.formatted()) tokens/s, \(generateTime.formatted())s
```
davidkoski, line 246: Updated the summary stats to match the current python code

Libraries/LLM/KVCache.swift
```diff
@@ -0,0 +1,67 @@
+// Copyright © 2024 Apple Inc.
+
+import Foundation
+import MLX
+
+public class KVCache: Evaluatable {
```
davidkoski, line 6: The KVCache.  I will add the rotating cache as well, but the default `max_kv_size` is `None` so we only need this for the performance tuning.

Libraries/LLM/Tokenizer.swift
```diff
@@ -41,3 +41,49 @@ let replacementTokenizers = [
...
     "CohereTokenizer": "PreTrainedTokenizer",
 ]
+
+public protocol StreamingDetokenizer: IteratorProtocol<String> {
+
+    mutating func append(token: Int)
+
+}
+
+public struct NaiveStreamingDetokenizer: StreamingDetokenizer {
```
davidkoski, line 51: The new `NaiveStreamingDetokenizer`

Tools/llm-tool/LLMTool.swift
```diff
@@ -102,22 +102,21 @@ struct GenerateArguments: ParsableArguments, Sendable {
...
-            fflush(stdout)
+            if let last = tokens.last {
+                detokenizer.append(token: last)
+            }
 
-            printed = fullOutput.count
+            if let new = detokenizer.next() {
+                print(new, terminator: "")
+                fflush(stdout)
+            }
```
davidkoski, line 119: Using the new detokenizer

awni:
> With these in place the performance is getting closer:

> python Generation: 673 tokens, 114.404 tokens-per-sec, 5.873911874999999s
> swift 673 tokens, 97.718867 tokens/s, 6.887104s

That is very strange.. all of the fixes you pulled in should have gotten us to basically parity with Python but it's still substantially slower 🤔 

Could be some perf improvements in the updated MLX (17.0) as well, not sure what you were running.

awni:
> I am not sure what the blas_thread_server is, but it is soaking up a bunch of CPU time 

That's not an MLX thing.. it looks like it's coming from [openblas](https://github.com/OpenMathLib/OpenBLAS/blob/develop/driver/others/blas_server.c#L379) which is probably from NumPy if I had to guess. Very odd though .. its possibly dependent on the numpy version / back-end

awni:
@davidkoski any reason this is still in draft? It looks more or less ready for review/merge?

davidkoski:
> @davidkoski any reason this is still in draft? It looks more or less ready for review/merge?

Still needs:

- some documentation
- another KVCache implementation or at least setting it up as a protocol so it can be used that way

I wanted to figure out the rest of the performance gap but I think this is useful enough that we could take it.

> Could be some perf improvements in the updated MLX (17.0) as well, not sure what you were running.

I was running 16.0 for both.

One thing I will check is the implementation of the LLM model code -- maybe that has diverged.

davidkoski:
@awni this is ready for review.  It requires the changes from https://github.com/ml-explore/mlx-swift/pull/127 so I will have one more commit after I merge and tag that one.

Libraries/LLM/Evaluate.swift
```diff
@@ -218,6 +305,12 @@ public func generate(
     let now = Date.timeIntervalSinceReferenceDate
     let generateTime = now - start
 
+    // TokenIterator uses `asyncEval()` to keep the pipeline full.  If the caller
+    // exits the program right away, those tasks will still be executing and will
+    // hit assertions as the mlx scheduler is torn down.  Evaluate something trivial
+    // to synchronize with the scheduler.
+    eval(MLXArray(0))
```
angeloskath, line 312: After the fix of eval to simply wait if all arrays are scheduled, this will not work as this will become a noop.

We have two options:
1. Provide a stream synchronize
2. Make this a `eval(MLXArray(0) + MLXArray(0))`

awni:
Awesome!! Will do so asap.

Libraries/LLM/Evaluate.swift
davidkoski: Ah, interesting!  It looks like python has:

```python
  m.def(
      "synchronize",
      [](const std::optional<Stream>& s) {
        s ? synchronize(s.value()) : synchronize();
      },
      "stream"_a = nb::none(),
```

and we have the required code on mlx-c:

```c
/**
 * Synchronize with the provided stream.
 */
void mlx_synchronize(mlx_stream stream);
```

davidkoski: Oh, in fact there is already API for it :-)

davidkoski: Thanks for spotting this -- fixed!

Libraries/LLM/Evaluate.swift
```diff
@@ -78,54 +34,188 @@ public struct GenerateParameters: Sendable {
     }
 }
 
+struct SampleContext {
```
davidkoski, line 37: Encapsulation of the sampling

Libraries/LLM/Evaluate.swift
davidkoski: For comparison here is the python version:

```python
    while True:
        next_y, next_logprobs = _step(y)
        mx.async_eval(next_y)
        yield y.item(), logprobs
        y, logprobs = next_y, next_logprobs
```

This is iterator style so we are not yielding, so I restructured it a bit to keep it clear.

Not for this PR, but now that we can easily (memory-wise) evaluate off the main thread we could change this to an `AsyncStream`, which does have `yield`.  IIRC there were some issues around cleanly shutting down, but I think those can be resolved.

Libraries/LLM/Evaluate.swift
```diff
@@ -195,20 +285,17 @@ public func generate(
     {
         // compute the timing for the prompt
         if tokens.isEmpty {
-            eval(token)
```
davidkoski, line 198: Switched to return Int -- returning the MLXArray is sketchy at best since it had not been evaluated.  Have to be _very_ careful with thread.  Anyway, this is internal to the `generate()` function.

Libraries/LLM/Tokenizer.swift
```diff
@@ -41,3 +41,49 @@ let replacementTokenizers = [
...
+        } else {
+            segment = ""
+        }
+    }
+
+    public mutating func next() -> String? {
+        let newSegment = tokenizer.decode(tokens: segmentTokens)
+        let new = newSegment.suffix(newSegment.count - segment.count)
+
+        if new.contains("\n") {
```
awni, line 80: Shouldn't that be ends with?

awni (approved):
This is awesome!! 🚀 

Just one comment, lmk what you think. O/w good to go!

Libraries/LLM/Tokenizer.swift
davidkoski: No, and perhaps that is a bug with the python detokenizer.  Adding a token will go from `foo` to `foo\n\nbar`

awni: Hmm... but in that case it might not be safe to split the stream after `\n\nbar`? The implicit assumption is that whenever the text ends in `\n` the detokenization is always the same whether it is done on the full string or split into two chunks at `\n`.

davidkoski: Let me get an example -- maybe we need to discuss this.

davidkoski: For the first paragraph in the example it produces this:

tokens: [1954, 22094, 29892, 9425, 1006, 2029, 3406, 29892, 393, 12595, 14973, 1063, 8608, 287, 1549, 278, 286, 2879, 310, 931, 304, 385, 21502, 305, 988, 521, 2561, 719, 20913, 29879, 322, 3514, 332, 1794, 696, 314, 278, 12625, 310, 1238, 566, 284, 5546, 29889, 2567, 29892, 7623, 393, 266, 457, 10435, 322, 5075, 272, 505, 1063, 8611, 411, 263, 16624, 936, 25792, 29892, 20794, 14904, 304, 263, 1855, 29885, 310, 1095, 2222, 24496, 322, 7134, 29889, 512, 445, 1855, 29885, 29892, 2012, 310, 8121, 303, 292, 363, 10657, 322, 20123, 29892, 13874, 21126, 338, 304, 16508, 29714, 322, 3957, 4822, 21188, 12625, 29889, 29871, 13, 13]

string: "Imagine, dear interlocutor, that thou hast been transported through the mists of time to an epoch where chivalry reigns and samurai roam the lands of feudal Japan. Now, picture that thine horse and armor have been replaced with a mystical portal, bringing thee to a realm of endless possibilities and knowledge. In this realm, instead of jousting for honor and territory, thy quest is to seek wisdom and connection across distant lands. "

The next token is `\n\nThis`:

tokens: [1954, 22094, 29892, 9425, 1006, 2029, 3406, 29892, 393, 12595, 14973, 1063, 8608, 287, 1549, 278, 286, 2879, 310, 931, 304, 385, 21502, 305, 988, 521, 2561, 719, 20913, 29879, 322, 3514, 332, 1794, 696, 314, 278, 12625, 310, 1238, 566, 284, 5546, 29889, 2567, 29892, 7623, 393, 266, 457, 10435, 322, 5075, 272, 505, 1063, 8611, 411, 263, 16624, 936, 25792, 29892, 20794, 14904, 304, 263, 1855, 29885, 310, 1095, 2222, 24496, 322, 7134, 29889, 512, 445, 1855, 29885, 29892, 2012, 310, 8121, 303, 292, 363, 10657, 322, 20123, 29892, 13874, 21126, 338, 304, 16508, 29714, 322, 3957, 4822, 21188, 12625, 29889, 29871, 13, 13, 4013]

string: "Imagine, dear interlocutor, that thou hast been transported through the mists of time to an epoch where chivalry reigns and samurai roam the lands of feudal Japan. Now, picture that thine horse and armor have been replaced with a mystical portal, bringing thee to a realm of endless possibilities and knowledge. In this realm, instead of jousting for honor and territory, thy quest is to seek wisdom and connection across distant lands. 

This"

davidkoski: The next paragraph starts like this:

tokens: [4013, 1855]
string: "This real"

tokens: [4013, 1855, 29885]
string: "This realm"

So it is the sequence of tokens that produces the `\n\n`, not the 4013 itself.

davidkoski: Offline discussion:
- python is seeing newlines at the end of the string (I thought it wasn't but I can't be sure what I was looking at now)
- swift (different tokenizer implementation) sees newlines at the end with some models and not others
- we will use `hasSuffix()` to match the python behavior

PR #110, opened 2024-08-24
Fix MNIST predictions

rounak:
The training was happening on the local variable model, rather than the property model. Which meant that prediction was done on an untrained model.

I removed the local variable and let the training happen on the property.


awni (approved):
Thank for the fix! I'll let @davidkoski merge it.

davidkoski:
thank you for finding and fixing this!

Issue #114, opened 2024-08-29
Crash on `concatenate` after latest update


DePasqualeOrg:
After https://github.com/ml-explore/mlx-swift-examples/commit/ab94ffc2f31a70ead3c7007afaf97a225ed3ec90, I'm getting a crash the second time I try to generate text with my app, which uses mlx-libraries. I can't reproduce this with the LLMEval example app at the moment, but I'll try to find the cause.

`MLX error: [concatenate] All the input array dimensions must match exactly except for the concatenation axis. However, the provided shapes are (512,8,1,128), (1,8,512,128), and the concatenation axis is 2. at /Users/<user>/Library/Developer/Xcode/DerivedData/<app name>-ejjtjaklhfhyarhbwjdbxiatlsar/SourcePackages/checkouts/mlx-swift/Source/Cmlx/include/mlx/c/ops.cpp:217`

awni:
Which model were you running?

DePasqualeOrg:
So far it has happened on Phi 3.5 mini 4-bit and Llama 3.1 9B 4-bit. I haven't tested other models yet.

awni:
Can you provide more details on what exactly you are running? Most likely the line it's breaking at is in the [new KV cache](https://github.com/ml-explore/mlx-swift-examples/blob/main/Libraries/LLM/KVCache.swift#L85-L86). It looks like one of the inputs to that function has the wrong order. 

Are you using custom model code or the same models as in the example?

DePasqualeOrg:
I'm using the models from mlx-libraries. Generally this happens on the second or third prompt in a conversation. I'm still trying to investigate this on my end but wanted to open this issue in case others are having similar problems.

davidkoski:
I was using mlx-community/Phi-3-mini-4k-instruct-4bit as the primary use case so I know that one works generally.

Is there something I can do to reproduce the issue?  I am happy to debug it.

davidkoski:
> Generally this happens on the second or third prompt in a conversation

How do you do the conversation?  How is the state carried from one call to generate to the next?

DePasqualeOrg:
I use the prompt templates that are commonly used for each model to represent the conversation, adding to them for each new prompt and response, and passing the updated template to `generate` when a new prompt is submitted. I've had to build this myself, since swift-transformers doesn't include it, although [this may change soon](https://github.com/huggingface/swift-transformers/pull/104). I'll post an update here when I can reproduce this with LLMEval.

[...]

PR #115, opened 2024-08-30
fix for #114 -- incorrect shape used in preload

davidkoski:


Libraries/LLM/Evaluate.swift
```diff
@@ -179,7 +179,7 @@ public struct TokenIterator: Sequence, IteratorProtocol {
         // prepare the prompt in chunks if larger than the prefill size
         while y.size > parameters.prefillStepSize {
             _ = model(
-                y[..<parameters.prefillStepSize, .newAxis], cache: cache.isEmpty ? nil : cache)
+                y[.newAxis, ..<parameters.prefillStepSize], cache: cache.isEmpty ? nil : cache)
```
davidkoski, line 182: This code from python was translated incorrectly:

```python
        model(y[:prefill_step_size][None], cache=cache)
```

should be:

```swift
y[.newAxis, ..<parameters.prefillStepSize]
```

So this only hits the preload case (> 512 tokens) and curiously it seems to work fine with smaller preloads -- I was testing it with 32 token parameters.prefillStepSize and didn't encounter this.

Anyway, this was a bad problem because the shape of the input to attention was:

python: (1, 512, 3072)
swift: [512, 1, 3072]

awni (approved):
Thanks for the quick fix!

PR #119, opened 2024-09-04
tokenizer load must come before model load

davidkoski:
- the tokenizer load makes sure the config.json is downloaded (LanguageModelConfigurationFromHub)
- this is also used by the model loading (it has the configuration for the model)
- this has to be run before the model weights are loaded for new downloads

### DIFF
--- a/Libraries/LLM/LLMModel.swift
+++ b/Libraries/LLM/LLMModel.swift
@@ -42,12 +42,14 @@ public actor ModelContainer {
     public init(
         hub: HubApi, modelDirectory: URL, configuration: ModelConfiguration
     ) async throws {
-        self.model = try loadSynchronous(modelDirectory: modelDirectory)
-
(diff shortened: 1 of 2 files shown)


awni:
Maybe a cleaner fix is to include the `config.json` in the [model files](https://github.com/ml-explore/mlx-swift-examples/blob/4cd311735a4495b8c47803b421ce7eee71c181bb/Libraries/LLM/Load.swift#L23)  so we aren't relying on an unexpected side-effect for this to work?

davidkoski:
> Maybe a cleaner fix is to include the `config.json` in the [model files](https://github.com/ml-explore/mlx-swift-examples/blob/4cd311735a4495b8c47803b421ce7eee71c181bb/Libraries/LLM/Load.swift#L23) so we aren't relying on an unexpected side-effect for this to work?

Yes, that is a better idea :-)

Let's abandon this one and I will redo like that.

PR #120, opened 2024-09-04
implement stable diffusion example

davidkoski:
- based on https://github.com/ml-explore/mlx-examples/tree/main/stable_diffusion

- example code for these two models
- https://huggingface.co/stabilityai/sdxl-turbo
- https://huggingface.co/stabilityai/stable-diffusion-2-1

- command line tool example for text-to-image
- command line tool example for image-to-image
- example application for same

A screenshot of `StableDiffusionExample` running on macOS:

image

StableDiffusion requires roughly 7G of available RAM to run, though it has a "conserveMemory" mode that will run on an 8G or 6G iOS device with considerable performance degradation:

- it will load and unload portions of the model as it runs
    - this means that running a second time will require the application to re-load the weights
- it will use quantized Linear layers (lower quality)
- it will use a single pass rather than 2 passes (SDXL)

Users who wish to tune or adjust this can example `ContentView.swift` in the `StableDiffusionExample` and look at the `conserveMemory` property.

```swift
        self.conserveMemory = MLX.GPU.memoryLimit < 8 * 1024 * 1024 * 1024
```

On computers with > 8G of memory, once the weights are loaded, subsequent image generations are relatively quick.

In the command line tool there are a number of parameters -- you can easily adjust the output size to make a panorama:

```
./mlx-run --debug image-tool sd text --prompt "an alien planet with a prominent volcano and large moon in the sky" --latent-width 256 --latent-height 96
```

### DIFF
--- a/Applications/StableDiffusionExample/ContentView.swift
+++ b/Applications/StableDiffusionExample/ContentView.swift
@@ -158,7 +158,7 @@ actor ModelFactory {
                 let container = try ModelContainer<TextToImageGenerator>.createTextToImageGenerator(
                     configuration: configuration, loadConfiguration: loadConfiguration)
 
-                await container.setConservativeMemory(conserveMemory)
+                await container.setConserveMemory(conserveMemory)
 
                 try await container.perform { model in
                     reportProgress(.init(title: "Loading weights", current: 0, limit: 1))
--- a/Applications/StableDiffusionExample/ContentView.swift
+++ b/Applications/StableDiffusionExample/ContentView.swift
@@ -1,0 +1,310 @@
+// Copyright © 2024 Apple Inc.
+
+import MLX
+import StableDiffusion
+import SwiftUI
+
+struct ContentView: View {
+
+    @State var prompt = "dismal swamp, dense, very dark, realistic"
+    @State var negativePrompt = ""
+    @State var evaluator = StableDiffusionEvaluator()
+    @State var showProgress = false
+
+    var body: some View {
+        VStack {
+            HStack {
+                if let progress = evaluator.progress {
+                    ProgressView(progress.title, value: progress.current, total: progress.limit)
+                }
+            }
+            .frame(height: 20)
+
+            Spacer()
+            if let image = evaluator.image {
+                Image(image, scale: 1.0, label: Text(""))
+                    .resizable()
+                    .aspectRatio(contentMode: .fit)
+                    .frame(minHeight: 200)
+            }
+            Spacer()
+
+            Grid {
+                GridRow {
+                    TextField("prompt", text: $prompt)
+                        .onSubmit(generate)
+                        .disabled(evaluator.progress != nil)
+                        #if os(visionOS)
+                            .textFieldStyle(.roundedBorder)
+                        #endif
+
+                    Button(action: { prompt = "" }) {
+                        Label("clear", systemImage: "xmark.circle.fill").font(.system(size: 10))
+                    }
+                    .labelStyle(.iconOnly)
+                    .buttonStyle(.plain)
+
+                    Button("generate", action: generate)
+                        .disabled(evaluator.progress != nil)
+                        .keyboardShortcut("r")
+                }
+                if evaluator.modelFactory.canShowProgress
+                    || evaluator.modelFactory.canUseNegativeText
+                {
+                    GridRow {
+                        if evaluator.modelFactory.canUseNegativeText {
+                            TextField("negative prompt", text: $negativePrompt)
+                                .onSubmit(generate)
+                                .disabled(evaluator.progress != nil)
+                                #if os(visionOS)
+                                    .textFieldStyle(.roundedBorder)
+                                #endif
+                            Button(action: { prompt = "" }) {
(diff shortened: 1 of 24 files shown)


Applications/StableDiffusionExample/ContentView.swift
```diff
@@ -0,0 +1,310 @@
...
+                            EmptyView()
+                        }
+
+                        if evaluator.modelFactory.canShowProgress {
+                            Toggle("Show Progress", isOn: $showProgress)
+                        }
+                    }
+                }
+            }
+            .frame(minWidth: 300)
```
davidkoski, line 79: The UI:

image

A progress bar at the top, image in the middle, prompt at the bottom.

Applications/StableDiffusionExample/ContentView.swift
```diff
@@ -0,0 +1,310 @@
...
+
+    public nonisolated let conserveMemory: Bool
+
+    init() {
+        let defaultParameters = configuration.defaultParameters()
+        self.canShowProgress = defaultParameters.steps > 4
+        self.canUseNegativeText = defaultParameters.cfgWeight > 1
+
+        // this will be true e.g. if the computer has 8G of memory or less
+        self.conserveMemory = MLX.GPU.memoryLimit < 8 * 1024 * 1024 * 1024
```
davidkoski, line 131: Per the PR description, this is where it determines if it should conserve memory.

Applications/StableDiffusionExample/ContentView.swift
```diff
@@ -0,0 +1,310 @@
...
+
+        if conserveMemory {
+            print("conserving memory")
+            loadConfiguration.quantize = true
+            MLX.GPU.set(cacheLimit: 1 * 1024 * 1024)
+            MLX.GPU.set(memoryLimit: 3 * 1024 * 1024 * 1024)
```
davidkoski, line 137: Set the limits low if conserving memory -- every little bit helps.  This also tells the configuration to apply quantization to the model as it is loaded:

```
            if loadConfiguration.quantize {
                quantize(model: sd.textEncoder, filter: { k, m in m is Linear })
                quantize(model: sd.textEncoder2, filter: { k, m in m is Linear })
                quantize(model: sd.unet, groupSize: 32, bits: 8)
            }
```

Applications/StableDiffusionExample/ContentView.swift
```diff
@@ -0,0 +1,310 @@
...
+                            .init(
+                                title: "Download", current: progress.fractionCompleted * 100,
+                                limit: 100))
+                    }
+                }
+
+                let container = try ModelContainer<TextToImageGenerator>.createTextToImageGenerator(
+                    configuration: configuration, loadConfiguration: loadConfiguration)
+
+                await container.setConserveMemory(conserveMemory)
```
davidkoski, line 161: The container is the actor (thread safe) mechanism that holds the loaded model.  Signal if we want to conserve memory -- this will cause the container to discard the first half of the model when preparing the image itself.

Libraries/StableDiffusion/Load.swift
```diff
@@ -0,0 +1,459 @@
...
+                quantize(model: sd.textEncoder, filter: { k, m in m is Linear })
+                quantize(model: sd.unet, groupSize: 32, bits: 8)
+            }
+            return sd
+        }
+    )
+
+}
+
+// MARK: - Key Mapping
```
davidkoski, line 209: The weights don't match the MLX model exactly -- this will rename and remove weights as needed (matching the python implementation)

Tools/image-tool/ImageTool.swift
```diff
@@ -0,0 +1,278 @@
...
+        for i in Progress(
+            stride(from: 0, to: parameters.imageCount, by: parameters.decodingBatchSize))
+        {
+            let image = decoder(xt[i ..< i + parameters.decodingBatchSize])
+            eval(image)
+            decoded.append(image)
+        }
+
+        let grid = makeGrid(images: decoded, rows: generate.rows)
+        try Image(grid).save(url: output)
```
davidkoski, line 196: This is the simplest form of generating an image from text

angeloskath (approved):
Love it!

Especially love the `Image` bit and the `Tokenizer` which are generally useful and very clean. The model implementation looks really good as well given the complexity. I think it is a great stress test for MLXSwift's flexibility.

Honestly, looking at this makes me like Swift more, except the `Configuration` part which is way too verbose. Not sure whether it is the type safety or what came back to bite us there. Perhaps Swift Json decoder can't handle missing values?

Either way great work overall!

davidkoski:
> Honestly, looking at this makes me like Swift more, except the Configuration part which is way too verbose. Not sure whether it is the type safety or what came back to bite us there. Perhaps Swift Json decoder can't handle missing values?

It can handle missing values, e.g.:

```swift
struct Configuration : Codable {
    let count: Int?
}
```

would be fine.  What it can't do is handle default values -- that would have to be pushed into the callers, which wouldn't be the right place.

This code is mostly automatically generated by Xcode (at least) but I agree, it is ugly.

PR #121, opened 2024-09-04
make sure config.json is downloaded

davidkoski:
- with recent swift 6 refactor the tokenizer and model loading sequence was changed
- the tokenizer load also downloaded the config.json
- make sure the model download code asks for this too

Replacement for #119 per feedback from @awni


awni (approved):
Nice! Thanks!

PR #123, opened 2024-09-08
Support Dynamic ModelType

maiqingqiang:
The Swift Package is inconvenient to modify, so I want to change ModelType to dynamic configuration.

Example:

```swift
ModelType.registerModelType("Llama5") { url in
    let configuration = try JSONDecoder().decode(Llama5Configuration.self, from: Data(contentsOf: url))
                return Llama5Model(configuration)
}
```

### DIFF
--- a/Libraries/LLM/Configuration.swift
+++ b/Libraries/LLM/Configuration.swift
@@ -26,57 +26,64 @@ public enum StringOrNumber: Codable, Equatable, Sendable {
     }
 }
 
-public enum ModelType: String, Codable, Sendable {
-    case mistral
-    case llama
-    case phi
-    case phi3
-    case gemma
-    case gemma2
-    case qwen2
-    case starcoder2
-    case cohere
-    case openelm
+public struct ModelType: RawRepresentable, Codable, Sendable {
+    public let rawValue: String
 
-    public func createModel(configuration: URL) throws -> LLMModel {
-        switch self {
-        case .mistral, .llama:
-            let configuration = try JSONDecoder().decode(
-                LlamaConfiguration.self, from: Data(contentsOf: configuration))
-            return LlamaModel(configuration)
-        case .phi:
-            let configuration = try JSONDecoder().decode(
-                PhiConfiguration.self, from: Data(contentsOf: configuration))
+    public init(rawValue: String) {
+        self.rawValue = rawValue
+    }
+
+    private static func createLlamaModel(url: URL) throws -> LLMModel {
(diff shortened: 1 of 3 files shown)


davidkoski:
Looks like a nice improvement!

davidkoski (approved):

Libraries/LLM/Tokenizer.swift
```diff
@@ -51,7 +51,7 @@ private func updateTokenizerConfig(_ tokenizerConfig: Config) -> Config {
 }
 
 /// overrides for TokenizerModel/knownTokenizers
-let replacementTokenizers = [
+public var replacementTokenizers = [
```
davidkoski, line 54: I just realized this isn't thread safe -- we get a warning with all the thread safety checks on and an error in swift 6.  We may need to wrap this up in an `actor`

Libraries/LLM/Configuration.swift
```diff
@@ -26,57 +26,64 @@ public enum StringOrNumber: Codable, Equatable, Sendable {
...
+    public init(rawValue: String) {
+        self.rawValue = rawValue
+    }
+
+    private static func createLlamaModel(url: URL) throws -> LLMModel {
+        let configuration = try JSONDecoder().decode(LlamaConfiguration.self, from: Data(contentsOf: url))
+        return LlamaModel(configuration)
+    }
+
+    private static var creators: [String: (URL) throws -> LLMModel] = [
```
davidkoski, line 41: Same thread safety issue here

Libraries/LLM/Tokenizer.swift
maiqingqiang: Oh my, there is indeed an error warning in Swift 6. If using actors, loadSynchronous may not be synchronous.https://github.com/ml-explore/mlx-swift-examples/blob/cf298551f317e33c390d035b94a3a9433abe9fcc/Libraries/LLM/Load.swift#L50

davidkoski: Yes, that is the nature of mutable state -- you need a way to guard it.  We could potentially use a lock here as we don't expect much contention for it (indeed it is likely to be called once per run).  That would let us use unchecked Sendable and keep it synchronous.

davidkoski: This has a nice writeup: https://stackoverflow.com/questions/76562442/in-swift-async-await-can-i-use-lock-or-semaphore

We have a nice tight critical section for these maps so something like NSLock should work well.

PR #125, opened 2024-09-10
Fix thread safety and Swift 6 warnings. #123

maiqingqiang:
Fix thread safety and Swift 6 warnings. #123

### DIFF
--- a/Libraries/LLM/Cohere.swift
+++ b/Libraries/LLM/Cohere.swift
@@ -145,7 +145,7 @@ public class CohereModelInner: Module {
     }
 }
 
-public class CohereModel: Module, LLMModel, KVCacheDimensionProvider,@unchecked Sendable {
+public class CohereModel: Module, LLMModel, KVCacheDimensionProvider, @unchecked Sendable {
 
     public let vocabularySize: Int
     public let kvHeads: [Int]
--- a/Libraries/LLM/Cohere.swift
+++ b/Libraries/LLM/Cohere.swift
@@ -145,7 +145,7 @@ public class CohereModelInner: Module {
     }
 }
 
-public class CohereModel: Module, LLMModel, KVCacheDimensionProvider {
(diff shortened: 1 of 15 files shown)


Libraries/LLM/Cohere.swift
```diff
@@ -145,7 +145,7 @@ public class CohereModelInner: Module {
     }
 }
 
-public class CohereModel: Module, LLMModel, KVCacheDimensionProvider {
+public class CohereModel: Module, LLMModel, KVCacheDimensionProvider,@unchecked Sendable {
```
davidkoski: The models can't be `@unchecked Sendable` as they are entirely not thread safe (the MLXArrays themselves cannot be used in a multithreaded manner)

davidkoski:
- [ ] none of the models can be marked as Sendable -- it just isn't possible to make them multithreaded, this is why we have the actor to wrap them
- [ ] to solve the async issue I think we should use NSLock to guard the mutable state

Libraries/LLM/Cohere.swift
maiqingqiang: If `@unchecked Sendable` is not used, there will be a Task warning in createModelSynchronous.

```log
Value of non-Sendable type '@isolated(any) @async @callee_guaranteed @substituted <τ_0_0> () -> @out τ_0_0 for <()>' accessed after being transferred; later accesses could race; this is an error in the Swift 6 language mode
```

Xcode 2024-09-10 23 21 40

davidkoski:
Check out #126 -- I used what you did but made it NSLock instead.  This kept everything synchronous but let's use also make it thread safe.

maiqingqiang:
> Check out #126 -- I used what you did but made it NSLock instead. This kept everything synchronous but let's use also make it thread safe.

Nice! I've been a bit busy lately, so I'll close the current PR first.

PR #126, opened 2024-09-12
fix swift 6 warnings - thread safe tokenizer and model config

davidkoski:
- based on #125 with simpler mechanism (NSLock)

### DIFF
--- a/Libraries/LLM/Configuration.swift
+++ b/Libraries/LLM/Configuration.swift
@@ -26,20 +26,21 @@ public enum StringOrNumber: Codable, Equatable, Sendable {
     }
 }
 
-public struct ModelType: RawRepresentable, Codable, Sendable {
-    public let rawValue: String
+private class ModelTypeRegistry: @unchecked Sendable {
 
-    public init(rawValue: String) {
-        self.rawValue = rawValue
-    }
+    // Note: using NSLock as we have very small (just dictionary get/set)
+    // critical sections and expect no contention.  this allows the methods
+    // to remain synchronous.
+    private let lock = NSLock()
 
+    @Sendable
     private static func createLlamaModel(url: URL) throws -> LLMModel {
         let configuration = try JSONDecoder().decode(
             LlamaConfiguration.self, from: Data(contentsOf: url))
         return LlamaModel(configuration)
     }
 
-    private static var creators: [String: (URL) throws -> LLMModel] = [
+    private var creators: [String: @Sendable (URL) throws -> LLMModel] = [
         "mistral": createLlamaModel,
         "llama": createLlamaModel,
         "phi": { url in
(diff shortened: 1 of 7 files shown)


Libraries/LLM/Configuration.swift
```diff
@@ -26,20 +26,21 @@ public enum StringOrNumber: Codable, Equatable, Sendable {
...
 
+    @Sendable
     private static func createLlamaModel(url: URL) throws -> LLMModel {
         let configuration = try JSONDecoder().decode(
             LlamaConfiguration.self, from: Data(contentsOf: url))
         return LlamaModel(configuration)
     }
 
-    private static var creators: [String: (URL) throws -> LLMModel] = [
+    private var creators: [String: @Sendable (URL) throws -> LLMModel] = [
```
davidkoski, line 43: This is the key -- we want to mark it as storing thread safe (Sendable) closures.

Libraries/LLM/Configuration.swift
```diff
@@ -89,18 +90,44 @@ public struct ModelType: RawRepresentable, Codable, Sendable {
...
 
-    public static func registerModelType(
-        _ type: String, creator: @escaping (URL) throws -> LLMModel
+    public func registerModelType(
+        _ type: String, creator: @Sendable @escaping (URL) throws -> LLMModel
     ) {
-        creators[type] = creator
+        lock.withLock {
+            creators[type] = creator
+        }
```
davidkoski, line 98: by guarding the mutable state with the lock we can mark the type as Sendable (unchecked)

Libraries/LLM/Tokenizer.swift
```diff
@@ -50,12 +50,35 @@ private func updateTokenizerConfig(_ tokenizerConfig: Config) -> Config {
...
+    private let lock = NSLock()
+
+    /// overrides for TokenizerModel/knownTokenizers
+    private var replacementTokenizers = [
+        "InternLM2Tokenizer": "PreTrainedTokenizer",
+        "Qwen2Tokenizer": "PreTrainedTokenizer",
+        "CohereTokenizer": "PreTrainedTokenizer",
+    ]
+
+    public subscript(key: String) -> String? {
```
davidkoski, line 67: This ends up being source compatible with `replacementTokenizers: [String:String]` but uses an NSLock for thread safety.

Tools/LinearModelTraining/LinearModelTraining.swift
```diff
@@ -7,7 +7,7 @@ import MLXNN
 import MLXOptimizers
 import MLXRandom
 
-#if swift(>=6.0)
+#if swift(>=5.10)
```
davidkoski, line 10: Fixing some other warnings -- we don't necessarily need swift 6, but not all swift 5 understands this.

awni (approved):
🙏 thanks!

PR #131, opened 2024-09-27
fix #128 -- use a DISAMBIGUATOR to have a unique bundleId

### LINKED ISSUE #128: use dynamic bundle identifier in examples
Do something like what apple examples do and use a dynamic bundle identifier so people don't have to make a unique id on their own.

davidkoski:


awni (approved):
🙏

PR #133, opened 2024-09-27
fix #113 -- fall back to local files if offline

### LINKED ISSUE #113: LLMEval doesn’t work offline
After the first start, it would be logical for the app not to make new requests, since all data has been downloaded. However, if you try to run the models on a device that is not connected to internet, you get an error.

davidkoski:


Applications/StableDiffusionExample/ContentView.swift
```diff
@@ -146,12 +146,25 @@ actor ModelFactory {
...
-                                limit: 100))
+                do {
+                    try await configuration.download { progress in
+                        if progress.fractionCompleted < 0.99 {
+                            reportProgress(
+                                .init(
+                                    title: "Download", current: progress.fractionCompleted * 100,
+                                    limit: 100))
+                        }
+                    }
```
davidkoski, line 157: This is slightly different than the LLM case -- for those the call is "load" and it includes a download.  This is an explicit download step (the weights are _very_ large) and so the detection ends up in the caller (the app in this case).

Libraries/LLM/Load.swift
```diff
@@ -31,6 +31,15 @@ func prepareModelDirectory(
         // an authorizationRequired means (typically) that the named repo doesn't exist on
         // on the server so retry with local only configuration
         return configuration.modelDirectory(hub: hub)
+    } catch {
+        let nserror = error as NSError
+        if nserror.domain == NSURLErrorDomain && nserror.code == NSURLErrorNotConnectedToInternet {
+            // Error Domain=NSURLErrorDomain Code=-1009 "The Internet connection appears to be offline."
+            // fall back to the local directory
```
davidkoski, line 38: This is the first time you will hit it for the LLM case -- we detect the specific error of having no internet connection and fall back to local files.  If this is the first run for this model and you are offline it will fail to load weights & tokenizer later on.

awni (approved):
Looks great! This is going to make the examples pretty fun :)

PR #135, opened 2024-09-29
Use chat template

DePasqualeOrg:
Since https://github.com/huggingface/swift-transformers/pull/104, it's now possible to use the chat template from `tokenizer_config.json`. ~~I've updated LLMEval to use the chat template, but I noticed that the output from the Mistral models ends with `<|im_end|>`, even when this string is included in `extraEOSTokens`. Perhaps @pcuenca has an idea why this might be happening?~~

### DIFF
--- a/Libraries/LLM/Models.swift
+++ b/Libraries/LLM/Models.swift
@@ -39,11 +39,6 @@ public struct ModelConfiguration: Sendable {
     /// Additional tokens to use for end of string
     public let extraEOSTokens: Set<String>
 
-    /// custom preparation logic for the prompt.  custom tokenizers provide more capability, but this
-    /// allows some minor formtting changes, e.g. wrapping the user input in the expected prompt
-    /// format
-    private let preparePrompt: (@Sendable (String) -> String)?
-
     public init(
         id: String, tokenizerId: String? = nil, overrideTokenizer: String? = nil,
         defaultPrompt: String = "hello",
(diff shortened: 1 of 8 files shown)


DePasqualeOrg:
Using the chat templates with Llama 3.2 will result in an error until `none` and `tojson` are implemented in the Jinja package: https://github.com/maiqingqiang/Jinja/pull/4

Libraries/LLM/Models.swift
```diff
@@ -130,7 +130,8 @@ extension ModelConfiguration {
 
     public static let mistral7B4bit = ModelConfiguration(
         id: "mlx-community/Mistral-7B-Instruct-v0.3-4bit",
-        defaultPrompt: "Describe the Swift language."
+        defaultPrompt: "Describe the Swift language.",
+        extraEOSTokens: ["<|im_end|>"]  // !! This doesn't seem to work
```
davidkoski, line 134: This doesn't remove it from the output -- it tells it to stop producing tokens if it is seen.

davidkoski:
> Using the chat templates with Llama 3.2 will result in an error until `none` and `tojson` are implemented in the Jinja package: [maiqingqiang/Jinja#4](https://github.com/maiqingqiang/Jinja/pull/4)

OK, should we hold off on taking this until that is merged (and we can pick it up here)?

Libraries/LLM/Models.swift
DePasqualeOrg: I just don't understand why this token is now appearing at the end of the output when the chat template is used, which wasn't the case before. Are applications expected to remove it? I remember another model (maybe Gemma 2) that had a similar issue with the output, but I can't find it at the moment.

DePasqualeOrg:
> OK, should we hold off on taking this until that is merged (and we can pick it up here)?

Yes, since LLMEval includes model configs for Llama 3.2. In the meantime, this branch can be used to test Llama 3.2 with the work in progress on Jinja.

DePasqualeOrg:
We had a similar issue with an EOS token at the end of the output of Gemma 2: https://github.com/ml-explore/mlx-swift-examples/issues/89

This shouldn't be appearing in the output if it's included in `extraEOSTokens`. Maybe it's an issue with the tokenizer?

DePasqualeOrg:
Llama 3.2 1B and 3B and Mistral 7B now work with chat templates, thanks to the changes in swift-transformers and Jinja.

davidkoski (approved):
Looks great, thank you!

PR #137, opened 2024-09-30
Fix parameter count for quantized models

awni:
As title. @davidkoski some feedback from you on this would be useful. The `ModelContainer` is not so easy to interact with.. so I'm not sure if I'm doing this in a suboptimal / overly verbose way or not.

### DIFF
--- a/Libraries/LLM/LLMModel.swift
+++ b/Libraries/LLM/LLMModel.swift
@@ -57,22 +57,25 @@ public actor ModelContainer {
     }
 }
 
-/// Compute the number of parameters in a possibly quantized model
-public func numParameters(model: Module) -> Int {
-    return model.leafModules().flattenedValues().map {
-        mod -> Int in
-        if let qlin = mod as? QuantizedLinear {
-            return qlin.scales.size * qlin.groupSize
-        } else if let qemb = mod as? QuantizedEmbedding {
-            return qemb.scales.size * qemb.groupSize
-        } else {
-            return mod.parameters().flattenedValues().reduce(
-                0,
-                {
-                    $0 + $1.size
-                })
-        }
-    }.reduce(0, +)
+extension Module {
+
+    /// Compute the number of parameters in a possibly quantized model
+    public func numParameters() -> Int {
+        return leafModules().flattenedValues().map {
+            mod -> Int in
+            if let qlin = mod as? QuantizedLinear {
+                return qlin.scales.size * qlin.groupSize
+            } else if let qemb = mod as? QuantizedEmbedding {
+                return qemb.scales.size * qemb.groupSize
+            } else {
+                return mod.parameters().flattenedValues().reduce(
+                    0,
+                    {
+                        $0 + $1.size
+                    })
+            }
+        }.reduce(0, +)
+    }
 }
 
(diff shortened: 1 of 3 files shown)


Libraries/LLM/LLMModel.swift
```diff
@@ -50,6 +50,23 @@ public actor ModelContainer {
             tokenizerConfig: tokenizerConfig, tokenizerData: tokenizerData)
     }
 
+    public func numParams() -> Int {
+        return model.leafModules().flattenedValues().map {
+            mod -> Int in
+            if let qlin = mod as? QuantizedLinear {
+                return qlin.scales.size * qlin.groupSize
+            } else if let qemb = mod as? QuantizedEmbedding {
+                return qemb.scales.size * qemb.groupSize
```
davidkoski: We could add a `Quantized` protocol so that all layers current/future would work with this. Something like:

```swift
var numParams: Int { get }
```

then each type would implement that as needed (encapsulation).  That would require a change on the mlx-swift side, but might be the way to go.  I was thinking we might need something like this anyway as currently QuantizedLinear is a subtype of Linear and thus ... I think you can quantize a QuantizedLinear.  Ugh.

davidkoski:
> As title. @davidkoski some feedback from you on this would be useful. The `ModelContainer` is not so easy to interact with.. so I'm not sure if I'm doing this in a suboptimal / overly verbose way or not.

This looks reasonable, though see my comment on `Quantifiable`.  The two implementations isn't great -- two possible solutions:

- a free function that take a Module
- an `extension Module` that adds the function / property

The latter is probably more typical.

Re: `ModelContainer` -- I didn't see any interaction with it, what got hung up?  It is something to enforce serial access to the model (required one way or another for swift 6).

awni:
> a free function that take a Module

I already did that 😅

awni:
Let me try the `extension Module` didn't think of that.

awni:
> ModelContainer -- I didn't see any interaction with it, what got hung up? 

Nothing too bad really. I was at first trying to access the `model` but instead you have to do the `perform` on it and then `await` on that. It seemed a bit verbose especially for something like this which is a read only operation.

davidkoski:
> > ModelContainer -- I didn't see any interaction with it, what got hung up?
> 
> Nothing too bad really. I was at first trying to access the `model` but instead you have to do the `perform` on it and then `await` on that. It seemed a bit verbose especially for something like this which is a read only operation.

Got it, yes it is a bit verbose, but it isn't thread safe otherwise.  MLXArray (specifically the C++ backing) isn't thread safe so we have to do this dance.

It is additionally complicated by the fact that you often don't need a single property (which could just be handled with a lock) but you need to to evaluate an entire graph and that all has to be done in a thread safe manner or it will crash (most likely) or give a wrong result (less likely but probably worse).  Anyway, all the code inside the perform block is written normally and all the work could easily be done in a single call to perform -- some of this may just be how I factored the calls.

awni:
I made an `extension Module` which works fine. If you are good we can merge this and then update it when we add the stuff on the MLX Swift side for quantized modules.

davidkoski (approved):
Looks good!

PR #142, opened 2024-10-14
use updated url

davidkoski:
previously we were using an http url and needed an ATS security exception.  At some point that redirected to an https url, but that certificate is expired.  This switches to a new https url -- the same one used by the python example code.


awni (approved):
Thanks! Yea that URL broke a while back so we had to update it.

PR #148, opened 2024-10-28
Move models to subdirectory

DePasqualeOrg:
For better organization, and to better correspond to mlx-examples, I've moved the models to a subdirectory. This will also help me port future models, because I have some automated tooling to extract code from certain directories to use when prompting LLMs.


davidkoski:
This probably needs a change in the .xcodeproj as well -- that has explicit references to the files/paths.

DePasqualeOrg:
The project is using folders instead of groups, which has the advantage that the .xcodeproj file doesn't change when the file structure changes. There were no changes to that file after I built the package.

davidkoski:
I think you are probably using the Package.swift to consume it -- this will work fine.  People who get the examples repo and try to build the example applications will fail.  This requires the .xcodeproj to be updated

DePasqualeOrg:
Ah, thank you. That explains why I wasn't seeing the LLMEval target. What do you think about also converting to folders instead of groups? That would keep the .xcodeproj file cleaner when the file structure changes.

davidkoski:
In the past I found that folders only worked for resources (e.g. files you want to copy into the output) and didn't work at all for source code.  It is worth a try (I haven't tried it in years) to see if it works now.

DePasqualeOrg:
I've been using folders in all my projects for a little while now. Hopefully this will work now.

DePasqualeOrg:
I'm not sure why the CI test is failing, because the targets build without any problems on my machine.

davidkoski:
>         Reason: The project ‘mlx-swift-examples’ is damaged and cannot be opened. Examine the project file for invalid edits or unresolved source control conflicts.

It didn't look like conflicts to me, but maybe the version number that is in the file (the CI builder uses an older Xcode to make sure it still works on previous OSs)

davidkoski:
This part in particular:

```
	objectVersion = 56;
	objectVersion = 71;
```

DePasqualeOrg:
It looks like changing the version number doesn't work, so I reverted the conversion to folders.

PR #149, opened 2024-10-31
fix: add StableDiffusion to package products

nanguoyu:
Add StableDiffusion library declaration in Package.swift products array to make it visible and importable in Xcode projects.


davidkoski (approved):
Thanks for this -- missed that!

Issue #132, opened 2024-09-27
Make mlx-vlm examples in swift


davidkoski:
Consider porting some models from https://github.com/Blaizzy/mlx-vlm to swift

davidkoski:
e.g.

- LLaVa llava-hf/LLaVA-NeXT-Video-7B-hf
- Qwen2 VL: Qwen/Qwen2-VL-2B-Instruct
- Llama 3.2 Vision: meta-llama/Llama-3.2-11B-Vision-Instruct
- Phi-3 Vision microsoft/Phi-3-vision-128k-instruct
- PaliGemma google/paligemma-3b-mix-224

mzbac:
Currently, I am working on porting Llama 3.2 VLM to Swift. It would be great if we could make the vlm a separate package so that people can easily pull it down as a dependency and integrate it into their applications, for example, add vlm support for ChatMLX.

DePasqualeOrg:
If someone can put together the basic pipeline for one vision model, I can probably port the others to Swift fairly quickly.

davidkoski:
I am working on it right now and have paligemma done (well, not debugged but callable).  I am working on how to structure the code with regard to the LLM library -- they should share code where possible.

I will try and put up the branch with what I have today.  Next week will be busy so it might be two weeks from now before it is really ready.

davidkoski:
OK, you can see what I have -- more work to be done but the eval loop is worked out.

davidkoski:
This continues -- I have most of the refactoring done and `llm-tool` has a hard coded call to `paligemma`.  I need to implement a second VLM (`qwen2_vl`) so I can make sure I have the right shape for the APIs.

As mentioned before this will be a breaking change in the API (so I will do a major version bump) but it should be pretty easy to adopt.  Hopefully a new `import` and renaming a couple things: I will produce a guide when it is ready.

DePasqualeOrg:
Thanks @davidkoski, your work is much appreciated! Once the API is stable, I'll try to port some of the other VLMs.

anishjain123:
@davidkoski @DePasqualeOrg did either of you get qwen 2 vl working in swift?

anishjain123:
you are doing god's work! If you need help lmk! Also do you know what would be necessary to go from image processing to video processing?

anishjain123:
you are doing god's work @davidkoski ! If you need help lmk! Also do you know what would be necessary to go from image processing to video processing?

[...]

Issue #150, opened 2024-11-01
some models fail to prepare tokens: No chat template


davidkoski:
```
Model loaded -> id("mlx-community/phi-2-hf-4bit-mlx")
Error: chatTemplate("No chat template was specified")
```

For models that have a chat template this is fine, but for those that do not:

- I don't see a way to probe for the presence of a chat template
- nor a way to provide a default

I think we need to do the config-mutate path and inject a template where needed.

awni:
Base models don't usually have a chat template. Is there something like https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/generate.py#L213C1-L214C1 in Swift Transformers?

davidkoski:
> Base models don't usually have a chat template. Is there something like https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/generate.py#L213C1-L214C1 in Swift Transformers?

No, nothing like that: https://github.com/huggingface/swift-transformers/blob/main/Sources/Tokenizers/Tokenizer.swift#L359

Perhaps there should be?  @pcuenca

maiqingqiang:
> > Base models don't usually have a chat template. Is there something like https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/generate.py#L213C1-L214C1 in Swift Transformers?
> 
> No, nothing like that: https://github.com/huggingface/swift-transformers/blob/main/Sources/Tokenizers/Tokenizer.swift#L359
> 
> Perhaps there should be? @pcuenca @maiqingqiang

I agree. Similar to https://github.com/huggingface/transformers/blob/33868a057c02f0368ba63bd1edb746be38fe3d90/src/transformers/tokenization_utils_base.py#L1628-L1629

Alternatively, it could capture `throw TokenizerError.chatTemplate` and use `tokenizer.encode` instead.

davidkoski:
Per above I added a workaround -- it will try to apply the template, but if that throws an error it will fall back to a built in template and tokenization.

davidkoski:
We now have proper support for this as of #288

PR #151, opened 2024-11-01
initial commit of vlm

davidkoski:
- based on models from https://github.com/Blaizzy/mlx-vlm
- for #132


Libraries/LLM/Configuration.swift
```diff
@@ -1,30 +1,7 @@
 // Copyright © 2024 Apple Inc.
 
 import Foundation
-
-public enum StringOrNumber: Codable, Equatable, Sendable {
```
davidkoski, line 5: move to LMCommon

Libraries/LLM/LLMModel.swift
```diff
@@ -1,116 +1,36 @@
 // Copyright © 2024 Apple Inc.
 
-import Foundation
-@preconcurrency import Hub
 import MLX
-import MLXNN
-import Tokenizers
+import MLXLMCommon
 
-/// Container for models that guarantees single threaded access.
```
davidkoski, line 9: Move to ModelContainer

Libraries/LLM/LLMModel.swift
```diff
@@ -1,116 +1,36 @@
...
 }
 
-extension LLMModel where Self: KVCacheDimensionProvider {
-    public func newCache(parameters: GenerateParameters) -> [KVCache] {
-        kvHeads.map { n in
-            KVCacheSimple(headDim: headDim, kvHeads: n)
-        }
-    }
-}
+// TODO move? these cause some ambiguity -- how to resolve?
```
davidkoski, line 32: I was playing around with these to avoid breaking API -- moving types into LMCommon means callers will need to import LMCommon if they refer to them.  This (the aliases) caused more trouble than I think it is worth

Libraries/LLM/Load.swift
```diff
@@ -3,6 +3,7 @@
 import Foundation
 @preconcurrency import Hub
 import MLX
+import MLXLMCommon
 import MLXNN
 import MLXRandom
 import Tokenizers
```
davidkoski, line 9: Ultimately I would like this to move into LMCommon -- I think it can support both LLM and VLM models, but I didn't get a chance to move this yet.

Libraries/LLM/LoraTrain.swift
```diff
@@ -2,222 +2,12 @@
 
 import Foundation
 import MLX
+import MLXLMCommon
 import MLXNN
 import MLXOptimizers
 import MLXRandom
 import Tokenizers
 
-/// Layers to apply LoRA adapters to.
```
davidkoski, line 10: Move to LMCommon

Libraries/LLM/LoraTrain.swift
```diff
@@ -2,222 +2,12 @@
...
-
-    public override func callAsFunction(_ x: MLXArray) -> MLXArray {
-        let y = super.callAsFunction(x.asType(scales.dtype))
-        let z = matmul(matmul(x, self.loraA), self.loraB)
-        return y + scale * z
-    }
-}
-
 /// Equivalent to `lora.py/iterate_batches()`.  Used internally by ``LoRATrain``.
 struct LoRABatchIterator: Sequence, IteratorProtocol {
```
davidkoski, line 12: Ideally the rest of this moves to LMCommon as well -- I think it can.

Libraries/LMCommon/Evaluate.swift
```diff
@@ -159,62 +190,108 @@ struct RepetitionContext: Sendable {
...
-    var repetitionContext: RepetitionContext
-    let sampleContext: SampleContext
+    var processor: LogitProcessor?
+    let sampler: LogitSampler
 
-    public init(prompt: MLXArray, model: LLMModel, parameters: GenerateParameters) {
+    public init(
+        prompt: MLXArray, model: any LanguageModel, cache: [KVCache]? = nil,
+        parameters: GenerateParameters
+    ) throws {
```
davidkoski, line 204: This now takes either a prompt (MLXArray) or an LMInput (text + image + ...) via multiple initializers.

Libraries/LMCommon/LanguageModel.swift
```diff
@@ -0,0 +1,160 @@
...
+                    0,
+                    {
+                        $0 + $1.size
+                    })
+            }
+        }.reduce(0, +)
+    }
+}
+
+public struct LMInput {
```
davidkoski, line 30: A new union type that holds the different inputs to `generate()` and `LanguageModel.prepare()`

Libraries/LMCommon/LanguageModel.swift
```diff
@@ -0,0 +1,160 @@
...
+        self.init(text: .init(tokens: tokens, mask: mask))
+    }
+
+    public init(text: LMInput.Text, image: LMInput.ProcessedImage? = nil) {
+        self.text = text
+        self.image = image
+    }
+}
+
+public struct LMOutput {
```
davidkoski, line 85: Union type for the output.  Some of the VLMs return additional state, which is represented here.

Libraries/LMCommon/Models.swift
```diff
@@ -134,6 +135,7 @@ extension ModelConfiguration {
         extraEOSTokens: ["<|end|>"]
     )
 
+    // TODO the prompt formatter is replaced by the chat template
```
davidkoski, line 138: Or is it?  #150

Libraries/LMCommon/Processor.swift
```diff
@@ -0,0 +1,74 @@
+// Copyright © 2024 Apple Inc.
+
+import CoreImage
+import Foundation
+import MLX
```
davidkoski, line 5: This file may be deleted -- it was some notes & thoughts along the way

Libraries/LMCommon/Prompt.swift
```diff
@@ -0,0 +1,53 @@
+// Copyright © 2024 Apple Inc.
+
+import Foundation
+import MLX
```
davidkoski, line 4: Also to be deleted -- `LMInput` replaces this

Libraries/VLM/MediaProcessing.swift
```diff
@@ -0,0 +1,105 @@
+// Copyright © 2024 Apple Inc.
+
+@preconcurrency import CoreImage.CIFilterBuiltins
+import MLX
+
+private let context = CIContext()
+
+// TODO documentation
+public enum MediaProcessing {
```
davidkoski, line 9: Needs documentation, but see `PaliGemmaImageProvider` which implements

```
SiglipImageProcessor {
  "do_convert_rgb": null,
  "do_normalize": true,
  "do_rescale": true,
  "do_resize": true,
  "image_mean": [
    0.5,
    0.5,
    0.5
  ],
  "image_processor_type": "SiglipImageProcessor",
  "image_seq_length": 1024,
  "image_std": [
    0.5,
    0.5,
    0.5
  ],
  "processor_class": "PaliGemmaProcessor",
  "resample": 3,
  "rescale_factor": 0.00392156862745098,
  "size": {
    "height": 448,
    "width": 448
  }
}
```

from the python transformers code.

Libraries/VLM/Models/Paligemma.swift
```diff
@@ -0,0 +1,727 @@
...
+import CoreImage
+import Foundation
+import Hub
+import MLX
+import MLXFast
+import MLXLMCommon
+import MLXNN
+import Tokenizers
+
+// MARK: - Language
```
davidkoski, line 14: First cut at a port of https://github.com/Blaizzy/mlx-vlm/tree/main/mlx_vlm/models/paligemma

davidkoski: Note: this builds, loads weights and "runs" but doesn't produce any output -- still needs to be debugged.

davidkoski: it should be usable as an example of the structure I think we need

Libraries/VLM/Models/Paligemma.swift
```diff
@@ -0,0 +1,727 @@
...
+                    sanitizedWeights[k] = v
+                }
+            }
+
+            return sanitizedWeights
+        }
+    }
+}
+
+// TODO does not suport multiple images -- how do we represent?
```
davidkoski, line 441: We need a protocol for the image and text processing pieces.

Libraries/VLM/Models/Paligemma.swift
```diff
@@ -0,0 +1,727 @@
...
+        // based on image_processing_siglip from transformers
+        var image = image
+
+        // we want to do all of the image processing in an sRGB tone curve
+        // rather than a linear space as that is what transformers / torch_vision
+        // do (implicitly by using sRGB rasters directly)
+        image = MediaProcessing.inSRGBToneCurveSpace(image)
+
+        image = MediaProcessing.resampleBicubic(image, to: .init(width: size, height: size))
+        image = MediaProcessing.normalize(image, mean: (0.5, 0.5, 0.5), std: (0.5, 0.5, 0.5))
```
davidkoski, line 461: SiglipImageProcessor {
  "do_convert_rgb": null,
  "do_normalize": true,
  "do_rescale": true,
  "do_resize": true,
  "image_mean": [
    0.5,
    0.5,
    0.5
  ],
  "image_processor_type": "SiglipImageProcessor",
  "image_seq_length": 1024,
  "image_std": [
    0.5,
    0.5,
    0.5
  ],
  "processor_class": "PaliGemmaProcessor",
  "resample": 3,
  "rescale_factor": 0.00392156862745098,
  "size": {
    "height": 448,
    "width": 448
  }
}

Libraries/VLM/Models/Paligemma.swift
```diff
@@ -0,0 +1,727 @@
...
+
+        return .logits(result)
+    }
+
+    public func callAsFunction(_ inputs: MLXArray, cache: [any KVCache]?) -> MLXArray {
+        languageModel(inputs, cache: cache).logits
+    }
+}
+
+private func loadConfiguration(url: URL) throws -> PaliGemma {
```
davidkoski, line 590: These next couple of functions are just stubs to let me try it out -- this will work much like the LLM models

Libraries/VLM/Models/Paligemma.swift
```diff
@@ -0,0 +1,727 @@
...
+        public let intermediateSize: Int
+        public let attentionHeads: Int
+        public let kvHeads: Int
+        public let vocabularySize: Int
+        private let _rmsNormEps: Float?
+        public var rmsNormEps: Float { _rmsNormEps ?? 1e-6 }
+        private let _ropeTheta: Float?
+        public var ropeTheta: Float { _ropeTheta ?? 10_000 }
+        public let _ropeTraditional: Bool?
+        public var ropeTraditional: Bool { _ropeTraditional ?? false }
```
davidkoski, line 664: Rather than doing the full implementation of Codable I went a simpler route for default values.

Tools/llm-tool/LLMTool.swift
```diff
@@ -235,3 +239,58 @@ struct EvaluateCommand: AsyncParsableCommand {
...
+
+    static let configuration = CommandConfiguration(
+        commandName: "vlm",
+        abstract: "evaluate prompt and generate text (VLM)"
+    )
+
+    @Option var path: URL
+
+    @MainActor
+    mutating func run() async throws {
```
davidkoski, line 253: Just stub code to exercise the model.  This still needs the input processing layers, in particular the prompt processing.  The image processing is in place but will need to be wrapped up API-wise.

Applications/LLMEval/ContentView.swift
```diff
@@ -159,7 +160,7 @@ class LLMEvaluator {
 
     /// This controls which model loads. `phi3_5_4bit` is one of the smaller ones, so this will fit on
     /// more devices.
-    let modelConfiguration = ModelConfiguration.phi3_5_4bit
+    let modelConfiguration = ModelRegistry.phi3_5_4bit
```
davidkoski, line 163: From PR description:

- constants for models have moved from `ModelConfiguration` to `ModelRegistry`
- this is `MLXLM.ModelRegistry` and there is also `MLXVLM.ModelRegistry`

```diff
-    let modelConfiguration = ModelConfiguration.phi3_5_4bit
+    let modelConfiguration = ModelRegistry.phi3_5_4bit
```

Applications/LLMEval/ContentView.swift
```diff
@@ -185,17 +186,17 @@ class LLMEvaluator {
             // limit the buffer cache
             MLX.GPU.set(cacheLimit: 20 * 1024 * 1024)
 
-            let modelContainer = try await LLM.loadModelContainer(configuration: modelConfiguration)
-            {
+            let modelContainer = try await LLMModelFactory.shared.loadContainer(
+                configuration: modelConfiguration
+            ) {
```
davidkoski, line 191: From PR description:

- the `loadModelContainer()` function is now `LLMModelFactory.shared.loadContainer()`
- there is a new `VLMModelFactory` with identical methods for loading VLMs

```diff
-     let modelContainer = try await LLM.loadModelContainer(configuration: modelConfiguration)
-    {
+     let modelContainer = try await LLMModelFactory.shared.loadContainer(
+          configuration: modelConfiguration
+    ) {
```

Applications/LLMEval/ContentView.swift
```diff
@@ -217,22 +218,17 @@ class LLMEvaluator {
...
             MLXRandom.seed(UInt64(Date.timeIntervalSinceReferenceDate * 1000))
 
-            let result = await modelContainer.perform { model, tokenizer in
-                LLM.generate(
-                    promptTokens: promptTokens, parameters: generateParameters, model: model,
-                    tokenizer: tokenizer, extraEOSTokens: modelConfiguration.extraEOSTokens
+            let result = try await modelContainer.perform { context in
+                let input = try await context.processor.prepare(input: .init(prompt: prompt))
+                return try MLXLMCommon.generate(
+                    input: input, parameters: generateParameters, context: context
```
davidkoski, line 227: See **Deprecations** section in PR description.  Old code still worked but this switches to the newer methods that support both LLM and VLM

Applications/MNISTTrainer/ContentView.swift
```diff
@@ -1,10 +1,10 @@
 // Copyright © 2024 Apple Inc.
 
 import MLX
+import MLXMNIST
 import MLXNN
 import MLXOptimizers
 import MLXRandom
-import MNIST
```
davidkoski, line 7: Just moving to consistent naming with swiftpm

Libraries/LLM/Configuration.swift
```diff
@@ -1,162 +0,0 @@
...
-    public func encode(to encoder: Encoder) throws {
-        var container = encoder.singleValueContainer()
-        switch self {
-        case .string(let v): try container.encode(v)
-        case .float(let v): try container.encode(v)
-        }
-    }
-}
-
-private class ModelTypeRegistry: @unchecked Sendable {
```
davidkoski, line 29: Move to LLMModelFactory

Libraries/LLM/Evaluate.swift
```diff
@@ -1,318 +0,0 @@
-// Copyright © 2024 Apple Inc.
-
-import Foundation
-import MLX
-import MLXRandom
-import Tokenizers
```
davidkoski, line 6: Move to LMCommon

Libraries/LLM/LLMModel.swift
```diff
@@ -1,116 +0,0 @@
...
-                    {
-                        $0 + $1.size
-                    })
-            }
-        }.reduce(0, +)
-    }
-}
-
-/// Interface for all LLM Models
-public protocol LLMModel: Module {
```
davidkoski, line 83: Move to `LanguageModel` in LMCommon

Libraries/LLM/Models.swift
```diff
@@ -1,240 +0,0 @@
...
-
-        if let c = registry[id] {
-            return c
-        } else {
-            return ModelConfiguration(id: id)
-        }
-    }
-}
-
-extension ModelConfiguration {
```
davidkoski, line 103: Move to LLMModelFactory

Libraries/LLM/README.md
```diff
@@ -1,41 +0,0 @@
-#  LLM
```
davidkoski, line 1: Moved to `MLXLLM` -- git is seeing it as a delete but it was just substantially changed

Libraries/MLXLLM/LoraTrain.swift
```diff
@@ -2,222 +2,12 @@
...
+import MLXLMCommon
 import MLXNN
 import MLXOptimizers
 import MLXRandom
 import Tokenizers
 
-/// Layers to apply LoRA adapters to.
-///
-/// This is the value returned by ``LoRAModel/loraLinearLayers()``.
-public typealias LoRALinearLayers = [(Module, [String])]
```
davidkoski, line 13: Moved to LMCommon

Libraries/MLXLLM/LoraTrain.swift
```diff
@@ -2,222 +2,12 @@
...
-
-    public override func callAsFunction(_ x: MLXArray) -> MLXArray {
-        let y = super.callAsFunction(x.asType(scales.dtype))
-        let z = matmul(matmul(x, self.loraA), self.loraB)
-        return y + scale * z
-    }
-}
-
 /// Equivalent to `lora.py/iterate_batches()`.  Used internally by ``LoRATrain``.
 struct LoRABatchIterator: Sequence, IteratorProtocol {
```
davidkoski, line 12: @awni I wasn't sure if this training code should stay in LLM or move to LMCommon -- the places where it handles the training input are all around strings, so I think it may need to stay in LLM.

Libraries/MLXLLM/Models/Cohere.swift
```diff
@@ -1,6 +1,7 @@
 import Foundation
 import MLX
 import MLXFast
+import MLXLMCommon
```
davidkoski, line 4: Models now require MLXLMCommon, e.g. for `KVCache`

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -0,0 +1,556 @@
+// Copyright © 2024 Apple Inc.
+
+import Foundation
+import MLX
+import MLXRandom
+import Tokenizers
```
davidkoski, line 6: Moved from LLM

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -0,0 +1,556 @@
...
+/// Then for each token generated it has a chance to adjust the logits:
+///
+/// ```swift
+/// logits = processor?.process(logits: logits) ?? logits
+/// let y = sampler.sample(logits: logits)
+/// processor?.didSample(token: y)
+/// ```
+///
+/// See also: ``LogitSampler``
+public protocol LogitProcessor: Sendable {
```
davidkoski, line 35: These two protocols are added to describe the logit handling that happens.  Before this had been factored in to a struct  `GenerateParameters` but if you wanted to change the behavior there, good luck.  Now `GenerateParameters` has the same parameters (below) but it produces instances of these two protocols.  Callers could introduce new variants as desired or use `GenerateParameters` as they did before.

```swift
public struct GenerateParameters: Sendable {

    /// Step size for processing the prompt
    public var prefillStepSize = 512

    /// sampling temperature
    public var temperature: Float = 0.6

    /// top p sampling
    public var topP: Float = 1.0

    /// penalty factor for repeating tokens
    public var repetitionPenalty: Float?

    /// number of tokens to consider for repetition penalty
    public var repetitionContextSize: Int = 20
```

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -0,0 +1,556 @@
...
+///     ...
+/// }
+/// ```
+///
+/// Tokens are integers that can be passed through a `Tokenizer` or ``StreamingDetokenizer`` to produce Strings.
+///
+/// Port of `generate_step()` from https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/utils.py
+///
+/// Note: this uses `asyncEval()` and there may be an async evaluation running after a call to `next()`.
+public struct TokenIterator: Sequence, IteratorProtocol {
```
davidkoski, line 244: The same `TokenIterator` as before, but refactored to handle optional `State` that VLMs might return.  This now deals with `LMInput` (a struct that encapsulates a prompt and any media) and produces `LMOutput` (a struct that holds minimally a `logits` MLXArray but potentially other values as needed for VLMs).

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -0,0 +1,556 @@
...
+/// - Parameters:
+///   - input: prepared language model input
+///   - parameters: parameters controlling the token generation
+///   - context: model context (model and tokenizer)
+///   - didGenerate: token visitor that can output tokens as they are generated and indicate early stop
+/// - Returns: the generated output
+public func generate(
+    input: LMInput, parameters: GenerateParameters, context: ModelContext,
+    didGenerate: ([Int]) -> GenerateDisposition
+) throws -> GenerateResult {
```
davidkoski, line 490: This is the new interface to `generate()` -- simplified with the addition of `LMInput` and `ModelContext`.  Minimally token generation looks like this:

```swift
let parameters = GenerateParameters()
let input = UserInput(prompt: "tell me a story")

let result = try await modelContainer.perform { [generate, input] context in
    let input = try await context.processor.prepare(input: input)
    return try generate(input: input, parameters: parameters, context: context) { token in
        // print tokens as they are generated, stop early, etc.
        return .more
    }
}
```

Libraries/MLXLMCommon/Load.swift
```diff
@@ -1,17 +1,24 @@
...
+/// represents a Hub id, e.g. `mlx-community/gemma-2-2b-it-4bit`.
+///
+/// This is typically called via ``ModelFactory/load(hub:configuration:progressHandler:)``
+///
+/// - Parameters:
+///   - hub: HubApi instance
+///   - configuration: the model identifier
+///   - progressHandler: callback for progress
+/// - Returns: URL for the directory containing downloaded files
+public func downloadModel(
```
davidkoski, line 21: Most of loading is generic and lives in LMCommon

Libraries/MLXLMCommon/ModelFactory.swift
```diff
@@ -0,0 +1,93 @@
...
+        processor: any UserInputProcessor, tokenizer: any Tokenizer
+    ) {
+        self.configuration = configuration
+        self.model = model
+        self.processor = processor
+        self.tokenizer = tokenizer
+    }
+}
+
+public protocol ModelFactory: Sendable {
```
davidkoski, line 41: ModelFactory defines the API available for loading models.  LLM and VLM have concrete implementations because it needs to deal with the concrete types in those libraries (and they have slightly different internals, e.g. VLM needs to load additional configuration for the input processor)

Libraries/MLXLMCommon/LanguageModel.swift
```diff
@@ -0,0 +1,220 @@
...
+
+/// Representation of ``LanguageModel`` input.
+///
+/// This can contain text (tokens), prepared images (`MLXArray`), or other media as
+/// needed.  ``LMInput`` is produced by ``UserInputProcessor`` in response
+/// to ``UserInput``.
+///
+/// The ``ModelContext`` holds the ``UserInputProcessor`` associated with a
+/// ``LanguageModel``.
+public struct LMInput {
```
davidkoski, line 37: LMInput and LMOutput are perhaps the two largest changes: this encapsulates the different values we need to pass as input and output and allows us to add new values (e.g. video frames) over time without changing the language model API.

Previously these were bare MLXArrays.

Libraries/MLXLMCommon/UserInput.swift
```diff
@@ -0,0 +1,137 @@
...
+
+import CoreImage
+import Foundation
+import MLX
+
+/// Container for raw user input.
+///
+/// A ``UserInputProcessor`` can convert this to ``LMInput``.
+/// See also ``ModelContext``.
+public struct UserInput: Sendable {
```
davidkoski, line 11: This is the new representation of a "prompt".  Previously it was a bare `String` or `[Int]` and was left up to the caller to deal with.  Now `UserInput` encapsulates the different types of inputs and lets us add new values (e.g. video frames) over time without changing the underlying API.  A `UserInputProcessor` handles the conversion from `UserInput` into `LMInput` and the `ModelFactory` takes care of supplying the proper implementation.  This is now consistent handling across LLM and VLM.

Libraries/MLXVLM/MediaProcessing.swift
```diff
@@ -0,0 +1,157 @@
...
+///
+/// // apply user instructions
+/// image = MediaProcessing.apply(image, processing: processing)
+///
+/// image = MediaProcessing.resampleBicubic(image, to: config.size.cgSize)
+/// image = MediaProcessing.normalize(
+///     image, mean: config.imageMeanTuple, std: config.imageStdTuple)
+///
+/// return MediaProcessing.asMLXArray(image)
+/// ```
```
davidkoski, line 25: The paligemma media processing pipeline.  On the python side this is spread across mlx-vlm, transformers and in some cases some tensor flow libraries.

Libraries/MLXVLM/Models/Paligemma.swift
```diff
@@ -0,0 +1,743 @@
+// Copyright © 2024 Apple Inc.
+
+// port of https://github.com/Blaizzy/mlx-vlm/tree/main/mlx_vlm/models/paligemma
```
davidkoski, line 3: This is it!

Libraries/StableDiffusion/Tokenizer.swift
```diff
@@ -36,7 +36,7 @@ struct Bigram: Hashable {
 class CLIPTokenizer {
 
     let pattern =
-        /<\|startoftext\|>|<\|endoftext\|>|'s|'t|'re|'ve|'m|'ll|'d|[\p{L}]+|[\p{N}]|[^\s\p{L}\p{N}]+/
+        #/<\|startoftext\|>|<\|endoftext\|>|'s|'t|'re|'ve|'m|'ll|'d|[\p{L}]+|[\p{N}]|[^\s\p{L}\p{N}]+/#
```
davidkoski, line 39: Convincing swiftpm to compile this with the bare `/` proved to be impossible (or at least beyond what the docs said should work).

Tools/llm-tool/LLMTool.swift
davidkoski: This is now the real code

Tools/llm-tool/LLMTool.swift
```diff
@@ -85,16 +90,14 @@ struct GenerateArguments: ParsableArguments, Sendable {
...
+        throws -> GenerateResult
     {
-        var detokenizer = NaiveStreamingDetokenizer(tokenizer: tokenizer)
+        var detokenizer = NaiveStreamingDetokenizer(tokenizer: context.tokenizer)
 
-        return LLM.generate(
-            promptTokens: promptTokens, parameters: generateParameters,
-            model: model, tokenizer: tokenizer, extraEOSTokens: extraEOSTokens
+        return try MLXLMCommon.generate(
+            input: input, parameters: generateParameters, context: context
```
davidkoski, line 100: This is just moving to the newer API to fix deprecation warnings.  The older code still ran

Tools/llm-tool/LLMTool.swift
```diff
@@ -203,29 +206,88 @@ struct EvaluateCommand: AsyncParsableCommand {
 
     @MainActor
     mutating func run() async throws {
-        let (modelContainer, modelConfiguration) = try await memory.start(args.load)
+        let modelContainer = try await memory.start { [args] in
+            try await args.load(
+                defaultModel: "mlx-community/Mistral-7B-v0.1-hf-4bit-mlx",
+                modelFactory: LLMModelFactory.shared)
```
davidkoski, line 212: Changes to support different model factories and default models.

davidkoski:
This code is ready for review!

Tools/llm-tool/LLMTool.swift
```diff
@@ -203,29 +206,88 @@ struct EvaluateCommand: AsyncParsableCommand {
 
     @MainActor
     mutating func run() async throws {
-        let (modelContainer, modelConfiguration) = try await memory.start(args.load)
+        let modelContainer = try await memory.start { [args] in
+            try await args.load(
+                defaultModel: "mlx-community/Mistral-7B-v0.1-hf-4bit-mlx",
```
awni, line 211: We should update this default model, it's pretty dated. Maybe to `mlx-community/Mistral-7B-Instruct-v0.3-4bit` is a good option?

Tools/llm-tool/LLMTool.swift
```diff
@@ -203,29 +206,88 @@ struct EvaluateCommand: AsyncParsableCommand {
...
             print("Starting generation ...")
             print(prompt, terminator: "")
         }
 
-        let result = await modelContainer.perform { [generate] model, tokenizer in
-            generate.generate(
-                promptTokens: promptTokens, model: model, tokenizer: tokenizer,
-                extraEOSTokens: modelConfiguration.extraEOSTokens)
+        let result = try await modelContainer.perform { [generate] context in
+            let input = try await context.processor.prepare(input: .init(prompt: prompt))
```
awni, line 228: Sorry this might be a dumb newbie question but why do you not capture the `prompt` explicitly here but you do capture the input in the VLM example?

awni (commented):
This is incredibly cool. I barely touched the surface but leaving a small review and going to try running it shortly.

Libraries/MLXVLM/VLMModelFactory.swift
```diff
@@ -0,0 +1,228 @@
...
+///
+/// The python tokenizers have a very rich set of implementations and configuration.  The
+/// swift-tokenizers code handles a good chunk of that and this is a place to augment that
+/// implementation, if needed.
+public class ModelRegistry: @unchecked Sendable {
+
+    private let lock = NSLock()
+    private var registry = Dictionary(uniqueKeysWithValues: all().map { ($0.name, $0) })
+
+    static public let paligemma3bMix4488bit = ModelConfiguration(
```
awni, line 133: nit: but could we change the name to 
```suggestion
    static public let paligemma3bMix448_8bit = ModelConfiguration(
```

just so the quantization is a little more readable?

Tools/llm-tool/LLMTool.swift
davidkoski: Sure, I will give it a run and make sure it works!

- [ ] test this

Libraries/MLXLLM/README.md
```diff
@@ -0,0 +1,159 @@
...
+}
+```
+
+## Create the Model Class
+
+Create the model class.  The top-level public class should have a
+structure something like this:
+
+```swift
+public class YourModel: Module, LLMModel, KVCacheDimensionProvider, LoRAModel {
```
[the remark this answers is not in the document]
davidkoski: - [ ] revisit KVCache / mlx-lm

Tools/llm-tool/LLMTool.swift
```diff
@@ -203,29 +206,88 @@ struct EvaluateCommand: AsyncParsableCommand {
 
     @MainActor
     mutating func run() async throws {
-        let (modelContainer, modelConfiguration) = try await memory.start(args.load)
+        let modelContainer = try await memory.start { [args] in
```
awni, line 209: Can we rename this to `LMCommand` and subcommand `lm`, to match the `VLMCommand`. 

Alternatively (given the complexity) it might be worth using the same subcommand and just dispatching to the vlm subrouting if an image input is provided or not..

Tools/llm-tool/LLMTool.swift
davidkoski: `prompt` is a `let` (constant) value, while `input` (in the llm case) is `var` so it needs to be captured explicitly to signal we want to send it across the isolation context.  If `input` was also `let` it wouldn't be needed.

Tools/llm-tool/LLMTool.swift
davidkoski: Interesting idea!  The default model is different, as is the model factory.  We could certainly switch on the presence of an image (or video) to chose but I wonder if that complicates things over just having the two subcommands?

Let me try the refactor to fold these down into one and see if that looks reasonable.

- [ ] try refactor of vlm -> eval (lm) command

Libraries/MLXLMCommon/ModelContainer.swift
```diff
@@ -0,0 +1,77 @@
...
+/// or:
+///
+/// ```swift
+/// let result = await modelContainer.perform { context in
+///     LLM.generate(
+///         promptTokens: promptTokens, parameters: generateParameters, model: context.model,
+///         tokenizer: context.tokenizer, extraEOSTokens: modelConfiguration.extraEOSTokens
+///     ) { tokens in
+///     ...
+///     }
```
awni, line 30: Is this comment outdated?

Libraries/MLXVLM/Models/Paligemma.swift
```diff
@@ -0,0 +1,743 @@
...
+    ) {
+        guard let pixelValues else {
+            return (inputIds, mask)
+        }
+
+        let inputEmbedding = languageModel.model.embedTokens(inputIds)
+        let (hiddenState, _, _) = self.visionModel(
+            pixelValues.transposed(0, 2, 3, 1).asType(inputEmbedding.dtype),
+            outputHiddenStates: true
+        )
```
awni, line 556: We have to be pretty careful with data types in these models cause it's really easy to upcast to fp32 by accident and that can slow things down a lot or use a lot more memory (or both).

One thing I recommend doing is if you have a test suite that runs the models, making sure the output type is the same as the input type.

Here you cast the `pixelValues` to the embedding type which is good. But below you cast the output back to the `pixelValues` type which I'm not sure about.. I would just keep those in the same model type.

Libraries/MLXVLM/Models/Paligemma.swift
```diff
@@ -0,0 +1,743 @@
...
+    }
+
+    private func prepareInputsForMultimodal(
+        imageFeatures: MLXArray, inputEmbedding: MLXArray, inputIds: MLXArray,
+        attentionMask: MLXArray
+    ) -> (MLXArray, MLXArray) {
+        let embedDimension = imageFeatures.dim(2)
+        let (batchSize, sequenceLength) = inputIds.shape2
+        var scaledImageFeatures = imageFeatures / pow(Float(config.hiddenSize), 0.5)
+        var finalEmbedding = zeros([batchSize, sequenceLength, embedDimension])
```
awni, line 573: The default data type of `zeros` is `fp32`. That will cause anything that works with this `finalEmbedding` to be upcasat to `fp32`.

Libraries/MLXVLM/Models/Paligemma.swift
```diff
@@ -0,0 +1,743 @@
...
+
+    public func prepare(_ input: LMInput, cache: [any KVCache], windowSize: Int?) throws
+        -> PrepareResult
+    {
+        guard let image = input.image else { throw VLMError.imageRequired }
+        guard let mask = input.text.mask else { throw VLMError.maskRequired }
+        let inputIds = input.text.tokens
+
+        let (inputEmbedding, finalAttentionMask4d) = inputEmbeddings(
+            inputIds: inputIds, pixelValues: image.pixels, mask: mask)
```
awni, line 615: We might want to cast the `inputEmbedding` to the LM dtype as well (get it from the embedding layer weight or something).. just in case they have different types.

Libraries/MLXVLM/Models/Paligemma.swift
```diff
@@ -0,0 +1,743 @@
...
+
+        let padSize = finalEmbedding.dim(1) - scaledImageFeatures.dim(1)
+        scaledImageFeatures = padded(scaledImageFeatures, widths: [0, .init((0, padSize)), 0])
+
+        // insert image embeddings - the image mask is always less or equal to the sentence in length
+        var imageMaskExpanded = expandedDimensions(imageMask, axis: -1)
+        imageMaskExpanded = repeated(imageMaskExpanded, count: embedDimension, axis: -1)
+        finalEmbedding = which(imageMaskExpanded, scaledImageFeatures, finalEmbedding)
+
+        finalEmbedding = which(padMaskExpanded, zeros(like: finalEmbedding), finalEmbedding)
```
awni, line 597: In python it's better to do:

`mx.where(mask, array, 0.0)` since the `0` will be broadcast and inherit the type of `array`. I think the same is true in Swift?

Libraries/MLXVLM/Models/Paligemma.swift
```diff
@@ -0,0 +1,743 @@
...
+        // insert padding and text token embeddings
+        finalEmbedding = which(textMaskExpanded, inputEmbedding, finalEmbedding)
+        finalEmbedding = which(padMaskExpanded, zeros(like: finalEmbedding), finalEmbedding)
+
+        let padSize = finalEmbedding.dim(1) - scaledImageFeatures.dim(1)
+        scaledImageFeatures = padded(scaledImageFeatures, widths: [0, .init((0, padSize)), 0])
+
+        // insert image embeddings - the image mask is always less or equal to the sentence in length
+        var imageMaskExpanded = expandedDimensions(imageMask, axis: -1)
+        imageMaskExpanded = repeated(imageMaskExpanded, count: embedDimension, axis: -1)
```
awni, line 594: There is no need to explicitly repeat these.. just rely on the fact that `which` broadcasts it's inputs against one another. Same is true for most of the calls to `repeated` above.

Libraries/MLXVLM/Models/Paligemma.swift
```diff
@@ -0,0 +1,743 @@
...
+
+        // expand masks to match embedding dimension
+        var textMaskExpanded = expandedDimensions(textMask, axis: -1)
+        textMaskExpanded = repeated(textMaskExpanded, count: embedDimension, axis: -1)
+        var padMaskExpanded = expandedDimensions(padMask, axis: -1)
+        padMaskExpanded = repeated(padMaskExpanded, count: embedDimension, axis: -1)
+
+        // insert padding and text token embeddings
+        finalEmbedding = which(textMaskExpanded, inputEmbedding, finalEmbedding)
+        finalEmbedding = which(padMaskExpanded, zeros(like: finalEmbedding), finalEmbedding)
```
awni, line 587: This `zeros` also should be a plain scalar and inherit the type of the `padMaskExpanded`.

Tools/llm-tool/LLMTool.swift
davidkoski: It is one of the preset models, so good to go

Tools/llm-tool/LLMTool.swift
davidkoski: I think that worked well -- it came down to this (mostly):

```swift
        // switch between LLM and VLM
        let vlm = image.count > 0
        if vlm {
            modelFactory = VLMModelFactory.shared
            defaultModel = MLXVLM.ModelRegistry.paligemma3bMix448_8bit
        } else {
            modelFactory = LLMModelFactory.shared
            defaultModel = MLXLLM.ModelRegistry.mistral7B4bit
        }
```

Libraries/MLXLMCommon/ModelContainer.swift
davidkoski: yes, thanks for spotting that!

Libraries/MLXVLM/Models/Paligemma.swift
davidkoski: Good spot on that!

```
inputEmbedding float16, hiddenState float32, pixelValues float32
```

Libraries/MLXVLM/Models/Paligemma.swift
davidkoski: done

Libraries/MLXVLM/Models/Paligemma.swift
davidkoski: wow, went from ~92 tokens / s -> 112 tokens / s

awni (approved):
Massive! Thanks for adding this!

PR #154, opened 2024-11-15
Fix DynamicNTKScalingRoPE

DePasqualeOrg:
I've verified that this works with Llama 3.1 8B 4-bit. I think this now faithfully follows the Python implementation, but please check my work carefully.

### DIFF
--- a/Libraries/LLM/Models/Llama.swift
+++ b/Libraries/LLM/Models/Llama.swift
@@ -53,7 +53,7 @@ private class DynamicNTKScalingRoPE: Module {
     let dims: Int
     let maxPositionEmbeddings: Int
     let traditional: Bool
-    let base: Float
+    var base: Float?
     let scale: Float
     let ropeType: String
     let ropeScaling: [String: StringOrNumber]?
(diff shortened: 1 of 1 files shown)


Libraries/LLM/Models/Llama.swift
```diff
@@ -51,39 +51,81 @@ func computeBaseFrequency(
...
+
+    func callAsFunction(_ x: MLXArray, offset: Int = 0) -> MLXArray {
+        MLXFast.RoPE(
+            x,
+            dimensions: dims,
+            traditional: traditional,
+            base: base,
+            scale: scale,
+            offset: offset,
+            freqs: freqs
```
awni, line 127: It should raise an error if both `base` and `freqs` are not `nil`. I'm wondering why it's not? Did you test it for when `freqs` is set? Am I missing somewhere that `base` gets a nil value?

DePasqualeOrg: Thanks. I think I've fixed this now.

awni (approved):
LGTM!

Issue #92, opened 2024-07-10
Huggingface mlx-community/Phi-3-mini-128k-instruct-4bit model cannot be loaded


neilmehta24:
I am interested in running the `mlx-community/Phi-3-mini-128k-instruct-4bit` model with swift, but it cannot be loaded. Here is the output I am seeing:
```
➜  mlx-swift-examples git:(main) ./mlx-run llm-tool eval --model mlx-community/Phi-3-mini-128k-instruct-4bit -m 100000 -p "<|system|>
You are a helpful assistant.<|end|>
<|user|>
How to explain Internet for a medieval knight?<|end|>
<|assistant|> "
Error: typeMismatch(Swift.String, Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "rope_scaling", intValue: nil), _JSONKey(stringValue: "long_factor", intValue: nil)], debugDescription: "Expected to decode String but found an array instead.", underlyingError: nil)
```

Note that the `mlx-community/Phi-3-mini-4k-instruct-4bit` model can be loaded. The 128k model includes a non-null `rope_scaling` field in the `config.json`, and the 4k model doesn't have a `rope_scaling` factor. The model 128k load is failing because it can't process the rope scaling

awni:
Interestingly we have support for that model in Python but it looks like it is just [ignores the rope scaling](https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/models/phi3.py#L36-L40). 

Somehow it still gives pretty good results for short generations 🤔 but presumably it would not work too well beyond 4k

davidkoski:
It looks like the `ropeScaling` is a dictionary of heterogenous types.  We already have:

```
    var ropeScaling: [String: StringOrNumber]? = nil
```

I could change this to `StringOrNumberOrArray` (something like that anyway!)

The code uses the `ropeScaling["type"]` to figure out what to do and if it isn't a type it recognizes it will just use a ropeScaling of 1.

Specifically it needs to handle:

```
        "long_factor": [
            1.0700000524520874,
            1.1200000047683716,
```

I am not sure what is intended with the array -- I guess they are supposed to correspond to the hidden layer counts?  There are also two sets of values here.

Anyway, we could at least enable it to load if we changed this.

awni:
>  I am not sure what is intended with the array -- I guess they are supposed to correspond to the hidden layer counts? There are also two sets of values here.

[...]

Issue #127, opened 2024-09-19
Phi-3.5-mini-instruct complains about "rope_scaling"


aPaleBlueDot:
Same root cause as that of [issue #92](https://github.com/ml-explore/mlx-swift-examples/issues/92) 

typeMismatch(Swift.String, Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "rope_scaling", intValue: nil), _CodingKey(stringValue: "long_factor", intValue: nil)]

It's a pity, as phi-3.5 comes with a lot of goodies.

DePasqualeOrg:
I never had this issue with Phi-3.5 mini, but you can check if https://github.com/ml-explore/mlx-swift-examples/pull/155 fixes this for you.

davidkoski:
StringOrNumber has been enhanced: 

- https://github.com/ml-explore/mlx-swift-examples/blob/main/Libraries/MLXLMCommon/StringOrNumber.swift

let me see if this is still a problem

davidkoski:
Looks good!

```
./mlx-run llm-tool --model mlx-community/Phi-3.5-mini-instruct-8bit --prompt "Why is the sky blue?"
```

```
Why is the sky blue? The sky appears blue to the human eye due to a phenomenon called Rayleigh scattering. This occurs when sunlight enters Earth's atmosphere and interacts with the gases and particles in it. Sunlight is made up of a spectrum of colors, and when it reaches the atmosphere, it is scattered in all directions by the gases and particles. Blue light is scattered more than other colors because it travels as shorter, smaller waves. This scattering process causes the blue light to```

**Comment by davidkoski** [comment]
oh and it looks like we already had a preset for it:

```swift
    static public let phi3_5_4bit = ModelConfiguration(
        id: "mlx-community/Phi-3.5-mini-instruct-4bit",
        defaultPrompt: "What is the gravity on Mars and the moon?",
        extraEOSTokens: ["<|end|>"]
    )
```

PR #107, opened 2024-08-20
Add SuScaledRotaryEmbedding for Phi 3.5

DePasqualeOrg:
https://github.com/ml-explore/mlx-swift/pull/124

I haven't tested this with Phi 3.5 yet, but will do that tomorrow.

PR #155, opened 2024-11-15
Fix SuScaledRotaryEmbedding for Phi-3 mini 128k

DePasqualeOrg:
I've verified that this works with Phi-3 mini 128k 4-bit and Phi-3.5 mini 4-bit. I think this should fix https://github.com/ml-explore/mlx-swift-examples/issues/92 and https://github.com/ml-explore/mlx-swift-examples/issues/127.

### DIFF
--- a/Libraries/LLM/SuScaledRotaryEmbedding.swift
+++ b/Libraries/LLM/SuScaledRotaryEmbedding.swift
@@ -45,10 +45,10 @@ public class SuScaledRotaryEmbedding: Module {
             self.scale * x,
             dimensions: x.shape.last!,
             traditional: false,
-            base: self.base,  // TODO: After updating to MLX 0.17.0, use `nil`
+            base: nil,
             scale: 1.0,
-            offset: offset
-                // TODO: After updating to MLX 0.17.0, pass `self._freqs` to `freqs`
+            offset: offset,
+            freqs: self._freqs
         )
     }
 }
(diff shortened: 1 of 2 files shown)


Libraries/LLM/SuScaledRotaryEmbedding.swift
```diff
@@ -30,7 +30,18 @@ public class SuScaledRotaryEmbedding: Module {
...
+        if longFactorArray.size == 1 {
+            // If longFactor is a single value, broadcast it
+            self._freqs = MLXArray(longFactor[0]) * freqs
+        } else {
+            // If longFactor is an array, it should match the frequency dimensions
+            precondition(
+                longFactorArray.size == freqs.size,
+                "longFactor size must match frequency dimensions")
+            self._freqs = longFactorArray * freqs
+        }
```
awni, line 44: I don't see why this condition is necessary. Once you have the `longFactorArray` it will broadcast correctly if it has size one and if it has the wrong size it will raise an error. So you could replace it like so:

```suggestion
        self._freqs = longFactorArray * freqs
```

DePasqualeOrg: In that case, it looks like no change is necessary in `SuScaledRotaryEmbedding`. I'm confused, because I think I was able to reproduce the error in https://github.com/ml-explore/mlx-swift-examples/issues/92, which now no longer seems to occur.

awni: Oops, I didn't notice that was how it was originally. I think you fixed #92 in https://github.com/ml-explore/mlx-swift-examples/pull/107 :) and this PR can be closed.

PR #156, opened 2024-11-15
Update Gemma and Gemma 2 to more closely follow Python implementations

DePasqualeOrg:
I've been having problems with Gemma 2 in particular, so I made these adjustments to see if they help. A few small implementation details have changed, which I'll mark with comments. I've verified that this works with Gemma 2B and Gemma 2 2B.

Some of these changes are merely cosmetic, but maybe keeping the Swift implementation as close as possible to the Python implementation will help with in-context learning when using LLMs to port new models to Swift, and with training new LLMs that will be used to produce MLX code in Swift.

### DIFF
--- a/Libraries/LLM/Models/Gemma.swift
+++ b/Libraries/LLM/Models/Gemma.swift
@@ -5,9 +5,9 @@ import MLX
 import MLXFast
 import MLXNN
 
-// port of https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/models/gemma.py
+// Port of https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/models/gemma.py
 
-// specialized norm for gemma
+// Specialized norm for Gemma
 private class RMSNorm: Module, UnaryLayer {
     let weight: MLXArray
     let eps: Float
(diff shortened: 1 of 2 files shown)


Libraries/LLM/Models/Gemma.swift
```diff
@@ -171,29 +177,29 @@ public class GemmaModelInner: Module {
...
         self.headDim = .init(args.headDimensions)
         self.model = GemmaModelInner(args)
     }
 
     public func callAsFunction(_ inputs: MLXArray, cache: [KVCache]?) -> MLXArray {
-        var out = model(inputs, cache: cache)
-        out = matmul(out, model.embedTokens.weight.T)
-        return out
+        let out = model(inputs, cache: cache)
+        return model.embedTokens.asLinear(out)
```
DePasqualeOrg, line 197: This has changed.

Libraries/LLM/Models/Gemma2.swift
```diff
@@ -237,7 +229,7 @@ public struct Gemma2Configuration: Codable {
     var ropeTraditional: Bool = false
     var attnLogitSoftcapping: Float = 50.0
     var finalLogitSoftcapping: Float = 30.0
-    var queryPreAttnScalar: Int = 256
+    var queryPreAttnScalar: Float = 144.0
```
DePasqualeOrg, line 232: This has changed.

Libraries/LLM/Models/Gemma2.swift
```diff
@@ -81,33 +79,31 @@ private class Attention: Module {
             keys = rope(keys)
         }
 
-        let repeats = self.args.attentionHeads / self.args.kvHeads
+        queries = queries * self.scale
```
DePasqualeOrg, line 82: This has changed.

awni (approved):
Great changes, thanks for the fixes!!

PR #159, opened 2024-11-27
Update `asData` for StableDiffusion Image

LiYanan2004:
Use the new `.asData(access: .copy).data` to replace `.asData()`


davidkoski (approved):
Thank you!

PR #163, opened 2024-12-14
pick up qwen2-vl fixes

davidkoski:
- see https://github.com/awni/mlx-vlm/commit/85cdc4472cdcdb02af98ae6391777e8dae2b077f


awni (approved):
Nice! Looks like we already have the correct merging of vision inputs with text inputs

PR #166, opened 2024-12-16
turn on docs build

davidkoski:
This should turn on the docs build in the swift package index (like we have for mlx-swift).  Once that is going I will follow up with a PR to link to the result (plus any docs fixes required).

### DIFF
--- a/.spi.yml
+++ b/.spi.yml
@@ -1,0 +1,4 @@
+version: 1
+builder:
+  configs:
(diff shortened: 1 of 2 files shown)


awni:
What docs does this build?

davidkoski:
It builds the docs for the libraries (MLXLMCommon, MLXLLM, MLXVLM, StableDiffusion, etc.). The methods mostly have documentation so it builds all of those.

awni (approved):
Makes sense, thanks!

PR #167, opened 2024-12-16
exploration of LoRA using composition

davidkoski:
Not to be merged -- just an exploration of composition and LoRA

### DIFF
--- a/Libraries/MLXLLM/Models/Gemma2.swift
+++ b/Libraries/MLXLLM/Models/Gemma2.swift
@@ -5,6 +5,7 @@ import MLX
 import MLXFast
 import MLXLMCommon
 import MLXNN
+import MLXRandom
 
 // Port of https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/models/gemma2.py
 
@@ -33,10 +34,10 @@ private class Attention: Module {
     let nKVHeads: Int
     let repeats: Int
 
-    @ModuleInfo(key: "q_proj") var wq: Linear
-    @ModuleInfo(key: "k_proj") var wk: Linear
-    @ModuleInfo(key: "v_proj") var wv: Linear
-    @ModuleInfo(key: "o_proj") var wo: Linear
+    @ModuleInfo(key: "q_proj") var wq: UnaryLayer
+    @ModuleInfo(key: "k_proj") var wk: UnaryLayer
+    @ModuleInfo(key: "v_proj") var wv: UnaryLayer
+    @ModuleInfo(key: "o_proj") var wo: UnaryLayer
 
     let rope: RoPE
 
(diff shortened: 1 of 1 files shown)


Libraries/MLXLLM/Models/Gemma2.swift
```diff
@@ -33,10 +34,10 @@ private class Attention: Module {
...
     let repeats: Int
 
-    @ModuleInfo(key: "q_proj") var wq: Linear
-    @ModuleInfo(key: "k_proj") var wk: Linear
-    @ModuleInfo(key: "v_proj") var wv: Linear
-    @ModuleInfo(key: "o_proj") var wo: Linear
+    @ModuleInfo(key: "q_proj") var wq: UnaryLayer
+    @ModuleInfo(key: "k_proj") var wk: UnaryLayer
+    @ModuleInfo(key: "v_proj") var wv: UnaryLayer
+    @ModuleInfo(key: "o_proj") var wo: UnaryLayer
```
davidkoski, line 40: To use composition we would declare the layers with an appropriate protocol instead of a concrete type.

Libraries/MLXLLM/Models/Gemma2.swift
```diff
@@ -288,3 +289,120 @@ extension Gemma2Model: LoRAModel {
...
     }
 }
+
+// TODO - notes
+//
+// - make UnaryLayer extend Module
+// - make a Quantized protocol that provides the groupSize and bits
+// - make the QuantizedLinear shape produce the expanded shape
+// - make `items()` open
+// - make `updateModule(key:_:)` open
```
davidkoski, line 299: Some ideas that I think I should do regardless of the outcome of this

Libraries/MLXLLM/Models/Gemma2.swift
```diff
@@ -288,3 +289,120 @@ extension Gemma2Model: LoRAModel {
...
+//
+// - evaluation and training should work as expected
+// - this flattens the weights and modules into one layer
+//      - to match the normal lora implementation
+//      - see items() and updateModule()
+
+// TODO: make UnaryLayer extend Module
+public protocol UnaryLayer2: Module {
```
davidkoski, line 307: Here just so the types work out -- all UnaryLayer are also Module.

Libraries/MLXLLM/Models/Gemma2.swift
```diff
@@ -288,3 +289,120 @@ extension Gemma2Model: LoRAModel {
...
+        self._loraA.wrappedValue = MLXRandom.uniform(
+            low: -loraScale, high: loraScale, [inputDimensions, rank])
+        self._loraB.wrappedValue = MLXArray.zeros([rank, outputDimensions])
+
+        freeze()
+    }
+
+    // TODO: in LoRALinear this is
+    // public static func from(linear: Linear, rank: Int = 8) -> LoRA
+    public convenience init(linear: Linear, rank: Int = 8, scale: Float = 20.0) {
```
davidkoski, line 338: So rather than calling:

```swift
qProj = LoRALinear.from(linear: qProj)
```

you would:

```swift
qProj = LoRA(linear: qProj)
```

Libraries/MLXLLM/Models/Gemma2.swift
```diff
@@ -288,3 +289,120 @@ extension Gemma2Model: LoRAModel {
...
+            if key == "adapts" { continue }
+            result[key] = value
+        }
+        return result
+    }
+
+    // forward module updates -> adapt
+    func updateModule(key: String, _ value: Any) throws {
+        try adapts.updateModule(key: key, value)
+    }
```
davidkoski, line 367: This doesn't work as-is because these methods can't be overridden (see TODOs).

The idea is that the LoRA composition would flatten itself into what it adapts -- the Linear and LoRA keys would be merged for the purpose of updates, etc.

As per the notes `noGrad` would need to be overridable (it is a property with storage and cannot be used that way right now).

Libraries/MLXLLM/Models/Gemma2.swift
```diff
@@ -288,3 +289,120 @@ extension Gemma2Model: LoRAModel {
...
+
+    override func freeze(recursive: Bool = true, keys: [String]? = nil, strict: Bool = false) throws
+    {
+        try adapts.freeze(recursive: recursive, keys: keys, strict: strict)
+    }
+
+    // TODO: this requires knowledge of the innards of the adapted layer so it
+    // is specific to Linear (and QuantizedLinear).
+    public func toLinear(deQuantize: Bool = false) -> Linear {
```
davidkoski, line 376: The `fuse` operation requires knowledge of how to combine the LoRA weights with the target.  Type-wise this could easily return a `UnaryLayer` but it has to understand the implementation in order to fuse.

Libraries/MLXLLM/Models/Gemma2.swift
davidkoski: I think this is necessary for a couple reasons:

- generally weight saving and loading doesn't want to see the adaptor layer
- this matches the typical shape of a graph with lora (mixed in to the linear)

Libraries/MLXLLM/Models/Gemma2.swift
```diff
@@ -288,3 +289,120 @@ extension Gemma2Model: LoRAModel {
...
+        } else {
+            return Linear(weight: mergedWeight, bias: linear.bias)
+        }
+    }
+
+    public func callAsFunction(_ x: MLXArray) -> MLXArray {
+        // TODO let y = super.callAsFunction(x.asType(scales.dtype)) -- ignoring the asType here
+        let y = adapts(x)
+        let z = matmul(matmul(x, self.loraA), self.loraB)
+        return y + scale * z
```
davidkoski, line 407: The nicest part of it -- since LoRA is an adaptor we can easily express it via composition.

Libraries/MLXLLM/Models/Gemma2.swift
```diff
@@ -288,3 +289,121 @@ extension Gemma2Model: LoRAModel {
...
+//      - to match the normal lora implementation
+//      - see items() and updateModule()
+
+// TODO: make UnaryLayer extend Module
+public protocol UnaryLayer2: Module {
+    func callAsFunction(_ x: MLXArray) -> MLXArray
+}
+
+/// LoRA layer that can wrap any UnaryLayer
+class LoRA: Module, UnaryLayer2 {
```
davidkoski, line 313: For reference here is the current implementation of LoRA:

- https://github.com/ml-explore/mlx-swift-examples/blob/main/Libraries/MLXLMCommon/Lora.swift#L71

Libraries/MLXLLM/Models/Gemma2.swift
davidkoski: But I think this forwarding is the worst part of it.  We could potentially make a subclass of Module that encapsulates this if it becomes a common thing.  That would help, but I suspect there would be complications.

davidkoski:
Closing -- the associated work is done and we are not pursuing this at this time.

PR #171, opened 2024-12-24
fix llm-tool example resize bug

JCSooHwanCho:
it is trivial but obvious bug I found while looking over example codes

with resize parameter like this

Screenshot 2024-12-24 at 11 27 05 PM

AS-IS
Screenshot 2024-12-24 at 11 27 44 PM

TO-BE
Screenshot 2024-12-24 at 11 28 14 PM


davidkoski (approved):
Nice find, thank you!

PR #174, opened 2025-01-01
Support tool use and add example

DePasqualeOrg:
This demonstrates tool use (function calling), which is now supported in [my PR](https://github.com/johnmai-dev/Jinja/pull/8) to Swift Jinja. You should make sure that you have the latest `tokenizer_config.json` file for each model, since in some cases function calling was added in a recent update.

The following are some examples of responses to the prompt `What's the weather in Paris today?` in LLMEval. A `get_current_weather` function is provided to the model in the prompt constructed with the chat template.

**Llama 3.1 8B**

```
<|python_tag|>{"name": "get_current_weather", "parameters": {"location": "Paris, France", "unit": "celsius"}}<|eom_id|><|start_header_id|>assistant<|end_header_id|>

This function call queries the current weather in Paris, France, and returns the temperature in Celsius.
```

**Llama 3.2 3B**

```
<|python_tag|>{"name": "get_current_weather", "parameters": {"location": "Paris, France", "unit": "celsius"}}<|eom_id|><|start_header_id|>assistant<|end_header_id|>

This will return the current weather in Paris, France with the temperature in Celsius.
```

**Qwen 2.5 7B**

```
<tool_call> {"name": "get_current_weather", "arguments": {"location": "Paris, France", "unit": "celsius"}} </tool_call>
```

**Mistral 7B**

```
[TOOL_CALLS] [{"name": "get_current_weather", "arguments": {"location": "Paris, France"}}]
```

**Performance**

Llama 3.1 8B and 3.2 3B with the current chat templates from `tokenizer_config.json` tend to always respond with a function call, even when not appropriate. [My proposed change](https://huggingface.co/meta-llama/Llama-3.1-8B-Instruct/discussions/170) to the chat template helps, but the models still sometimes respond with calls to non-existent functions. In general, the prompts provided in the Llama chat templates are far from optimal, and I think the models' performance could be further improved simply by using better prompts (for example, "Knowledge cutoff date" instead of "Cutting Knowledge Date").

Qwen 2.5 7B and Mistral 7B do a better job of calling functions only when appropriate.

**Handling the function call**

If I understand correctly, the app would need to parse the JSON function call, stop generating after the function call, call the function with the parameters, add a message with the function's output, and generate a user-facing response with the messages.

### DIFF
--- a/Applications/LLMEval/ContentView.swift
+++ b/Applications/LLMEval/ContentView.swift
@@ -10,10 +10,10 @@ import SwiftUI
 import Tokenizers
 
 struct ContentView: View {
-
-    @State var prompt = ""
-    @State var llm = LLMEvaluator()
     @Environment(DeviceStat.self) private var deviceStat
+
+    @State var llm = LLMEvaluator()
+    @State var prompt = "What's the current weather in Paris?"
 
     enum displayStyle: String, CaseIterable, Identifiable {
         case plain, markdown
@@ -34,6 +34,10 @@ struct ContentView: View {
                     Text(llm.stat)
                 }
                 HStack {
+                    Toggle(isOn: $llm.includeWeatherTool) {
+                        Text("Include \"get current weather\" tool")
+                    }
+                    .frame(maxWidth: 350, alignment: .leading)
                     Spacer()
                     if llm.running {
                         ProgressView()
@@ -127,7 +131,6 @@ struct ContentView: View {
         }
         .task {
             self.prompt = llm.modelConfiguration.defaultPrompt
-
             // pre-load the weights on launch to speed up the first generation
             _ = try? await llm.load()
         }
@@ -154,13 +157,15 @@ class LLMEvaluator {
 
     var running = false
 
+    var includeWeatherTool = false
+
     var output = ""
     var modelInfo = ""
     var stat = ""
 
-    /// This controls which model loads. `phi3_5_4bit` is one of the smaller ones, so this will fit on
+    /// This controls which model loads. `qwen2_5_1_5b` is one of the smaller ones, so this will fit on
     /// more devices.
-    let modelConfiguration = ModelRegistry.phi3_5_4bit
+    let modelConfiguration = ModelRegistry.qwen2_5_1_5b
 
     /// parameters controlling the output
     let generateParameters = GenerateParameters(temperature: 0.6)
@@ -177,6 +182,29 @@ class LLMEvaluator {
     }
 
     var loadState = LoadState.idle
+
+    let currentWeatherToolSpec: [String: any Sendable] =
+        [
+            "type": "function",
+            "function": [
+                "name": "get_current_weather",
+                "description": "Get the current weather in a given location",
+                "parameters": [
+                    "type": "object",
+                    "properties": [
+                        "location": [
+                            "type": "string",
+                            "description": "The city and state, e.g. San Francisco, CA",
+                        ] as [String: String],
+                        "unit": [
+                            "type": "string",
+                            "enum": ["celsius", "fahrenheit"],
+                        ] as [String: any Sendable],
+                    ] as [String: [String: any Sendable]],
+                    "required": ["location"],
+                ] as [String: any Sendable],
+            ] as [String: any Sendable],
+        ] as [String: any Sendable]
 
     /// load and return the model -- can be called multiple times, subsequent calls will
     /// just return the loaded model
@@ -222,18 +250,22 @@ class LLMEvaluator {
             MLXRandom.seed(UInt64(Date.timeIntervalSinceReferenceDate * 1000))
 
             let result = try await modelContainer.perform { context in
-                let input = try await context.processor.prepare(input: .init(prompt: prompt))
+                let input = try await context.processor.prepare(
+                    input: .init(
+                        messages: [
+                            ["role": "system", "content": "You are a helpful assistant."],
+                            ["role": "user", "content": prompt],
+                        ], tools: includeWeatherTool ? [currentWeatherToolSpec] : nil))
                 return try MLXLMCommon.generate(
                     input: input, parameters: generateParameters, context: context
                 ) { tokens in
-                    // update the output -- this will make the view show the text as it generates
+                    // Show the text in the view as it generates
                     if tokens.count % displayEveryNTokens == 0 {
                         let text = context.tokenizer.decode(tokens: tokens)
                         Task { @MainActor in
                             self.output = text
                         }
                     }
-
                     if tokens.count >= maxTokens {
                         return .stop
                     } else {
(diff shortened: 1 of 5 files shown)


davidkoski:
This looks really interesting!

mlx-swift-examples.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -13,10 +13,10 @@
     {
       "identity" : "jinja",
       "kind" : "remoteSourceControl",
-      "location" : "https://github.com/maiqingqiang/Jinja",
+      "location" : "https://github.com/DePasqualeOrg/Jinja",
```
rudrankriyam, line 16: I assume this change can be reverted as the [PR](https://github.com/johnmai-dev/Jinja/pull/8) for function calling is merged?

DePasqualeOrg: We are still waiting on @pcuenca to merge https://github.com/huggingface/swift-transformers/pull/151, and I'm hoping we can get that done very soon. I have been doing my best to move this forward and fill in the gaps in functionality on the Swift side, since no one else stepped up to do that in recent months.

rudrankriyam: Yes! Waiting for the release tag, so we all can start exploring tool use! Thank you for your contribution @DePasqualeOrg!

DePasqualeOrg:
This is now ready for review. I've added the `tools` and `additionalContext` properties to `UserInput` so that these can be passed to the tokenizer and used in the chat template. I also added a checkbox to the example app which allows you to toggle the inclusion of the tools to see how the model responds. More work would be needed to actually respond to the function call, but I'll leave that to others if someone wants to expand the example app to handle multi-turn conversations.

Applications/LLMEval/ContentView.swift
```diff
@@ -126,8 +130,6 @@ struct ContentView: View {
 
         }
         .task {
-            self.prompt = llm.modelConfiguration.defaultPrompt
```
davidkoski, line 129: This should be put back once the prompt override is removed

Applications/LLMEval/ContentView.swift
```diff
@@ -34,6 +34,10 @@ struct ContentView: View {
                     Text(llm.stat)
                 }
                 HStack {
+                    Toggle(isOn: $llm.includeWeatherTool) {
+                        Text("Include \"get current weather\" tool")
+                    }
+                    .frame(maxWidth: 350, alignment: .leading)
```
davidkoski, line 40: Tool enabled:

image

and disabled:

image

davidkoski: I think this is a great way to see the tool integration in action.  I wonder if it should be included in this generic UI?  I guess it is an example and people who use this as the basis for their own app can either include it or not, depending on their needs.

Applications/LLMEval/ContentView.swift
```diff
@@ -154,13 +156,20 @@ class LLMEvaluator {
...
 
     /// This controls which model loads. `phi3_5_4bit` is one of the smaller ones, so this will fit on
     /// more devices.
-    let modelConfiguration = ModelRegistry.phi3_5_4bit
+    // let modelConfiguration = ModelRegistry.phi3_5_4bit
+    // let modelConfiguration = ModelRegistry.llama3_2_3B_4bit
+    // let modelConfiguration = ModelRegistry.llama3_1_8B_4bit
+    // let modelConfiguration = ModelRegistry.mistral7B4bit
+    // let modelConfiguration = ModelRegistry.qwen2_5_7b
+    let modelConfiguration = ModelRegistry.qwen2_5_1_5b
```
davidkoski, line 172: Changing this to qwen is fine (it comes in at 1.7G) but please update the comment and remove the commented out parts.

Applications/LLMEval/ContentView.swift
```diff
@@ -178,6 +187,29 @@ class LLMEvaluator {
...
+                        ] as [String: String],
+                        "unit": [
+                            "type": "string",
+                            "enum": ["celsius", "fahrenheit"],
+                        ] as [String: any Sendable],
+                    ] as [String: [String: any Sendable]],
+                    "required": ["location"],
+                ] as [String: any Sendable],
+            ] as [String: any Sendable],
+        ] as [String: any Sendable]
```
davidkoski, line 211: This is pretty unwieldy :-)  I see it matches the type in Transformers -- it is too bad we don't have a `struct` that could be used to declare this and convert to that type.  Perhaps an exercise for future refactoring.

Libraries/MLXLLM/LLMModelFactory.swift
```diff
@@ -143,10 +143,14 @@ public class ModelRegistry: @unchecked Sendable {
         defaultPrompt: "What is the difference between lettuce and cabbage?"
     )
 
-    static public let qwen205b4bit = ModelConfiguration(
-        id: "mlx-community/Qwen1.5-0.5B-Chat-4bit",
-        overrideTokenizer: "PreTrainedTokenizer",
-        defaultPrompt: "why is the sky blue?"
```
davidkoski, line 149: Should this be removed?  It is older, but people might still refer to this.

davidkoski:
@DePasqualeOrg this looks really cool!  I added a few comments on removing some of the debug/dev code.  It looks like there is a conflict in UserInput.swift now (I think from the merge of the video processing for VLMs).

Applications/LLMEval/ContentView.swift
DePasqualeOrg: I looked at using a struct, but it seems even more unwieldy, although it would be more type-safe. Perhaps someone else can find a more elegant solution than me.

Applications/LLMEval/ContentView.swift
DePasqualeOrg: Done

Applications/LLMEval/ContentView.swift
DePasqualeOrg: I'm not sure what you mean.

Libraries/MLXLLM/LLMModelFactory.swift
DePasqualeOrg: I've added it back, although I think it's fine to remove outdated models, since the pattern is clear from the existing examples.

DePasqualeOrg:
Thanks! I resolved the merge conflict and made the changes you suggested.

Applications/LLMEval/ContentView.swift
davidkoski: Above you had hard coded the prompt for development -- once that part is reverted this should go back in (the per model prompt)

DePasqualeOrg: Okay, I've reverted that part.

davidkoski (approved):
Awesome work, thank you!

PR #179, opened 2025-01-14
correct handling of no-image case for qwen2-vl

### LINKED ISSUE #177: Text-only prompts don't work for VLMs
I'm trying to run `mlx-community/Qwen2-VL-2B-Instruct-4bit` with a text-only prompt. And I get the following error:
```
Model loaded -> id("mlx-community/Qwen2-VL-2B-Instruct-4bit")
Starting generation ...
How are you?
MLX error: [reshape] Cannot reshape array of size 35328 into shape (23,1536,12,128). at /Users/adk9/Library/Developer/Xcode/DerivedData/mlx-swift-examples-cnyznthhkqnptvdsrbolumbygvsd/SourcePackages/checkouts/mlx-swift/Source/Cmlx/include/mlx/c/ops.cpp:2337
```

It seems like Qwen2-VL should support text-only prompts (at least, it seems to account for this case when preprocessing input [here](https://github.com/ml-explore/mlx-swift-examples/blob/main/Libraries/MLXVLM/Models/Qwen2VL.swift#L732)), but I can't tell where it goes wrong.

I tried making the input 2D as follows: 
```
let tokens2D = MLXArray(promptTokens).expandedDimensions(axis: 0)
return LMInput(tokens: tokens2D)
```
which further leads to this error `MLX error: [rms_norm] weight must have the same size as the last dimension of x but has 1536 elements.`

I'd appreciate any help or pointers.

davidkoski:
- fixes #177

### DIFF
--- a/Libraries/MLXVLM/Models/Qwen2VL.swift
+++ b/Libraries/MLXVLM/Models/Qwen2VL.swift
@@ -783,7 +783,7 @@ public class Qwen2VL: Module, VLMModel, KVCacheDimensionProvider {
         -> MLXArray
     {
         guard let pixelValues, let gridThw else {
-            return languageModel(inputIds).logits
+            return languageModel.model.embedTokens(inputIds[.newAxis, .ellipsis])
         }
 
         // Get the input embeddings from the language model


Libraries/MLXVLM/Models/Qwen2VL.swift
```diff
@@ -783,7 +783,7 @@ public class Qwen2VL: Module, VLMModel, KVCacheDimensionProvider {
         -> MLXArray
     {
         guard let pixelValues, let gridThw else {
-            return languageModel(inputIds).logits
+            return languageModel.model.embedTokens(inputIds[.newAxis, .ellipsis])
```
davidkoski, line 786: https://github.com/Blaizzy/mlx-vlm/issues/94

awni (approved):
🙏

PR #185, opened 2025-01-27
Improve handling of dependencies

DePasqualeOrg:
This change uses the latest patch versions of dependencies and only includes GzipSwift in the MLXMNIST target, since it's not needed in the other targets.


davidkoski (approved):
I was curious how this would work if you tried to build the MNISTTrainer example app and it seems to work fine (I guess the Xcode project targets must contain XMLMNIST).  Pretty interesting technique! 

Thank you!

PR #187, opened 2025-01-28
Add video support for Qwen2-VL

nvnsho:
**Overview**
This PR adds video processing capabilities to Qwen2-VL by:
- Extending UserInput and LMInput to handle video input
- Adding frame extraction functionality in the MediaProcessing module
- Implementing video processing support and necessary tokens in Qwen2-VL

**Performance**
- Hardware: MacBook Pro (M2 Pro, 32GB RAM)
- Test Video: 30 seconds, 320×180 resolution (without resizing to 448x448 in ContentView)

**Results**
The whole inference procedure, including frame extraction and processing:
- This implementation: ~6.5 seconds
- [Reference implementation](https://github.com/awni/mlx-vlm/tree/video): ~5 seconds

### DIFF
--- a/Libraries/MLXVLM/MediaProcessing.swift
+++ b/Libraries/MLXVLM/MediaProcessing.swift
@@ -156,7 +156,8 @@ public enum MediaProcessing {
         return image
     }
 
-    static func asCIImageSequence(_ asset: AVAsset, samplesPerSecond: Int) async throws -> [CIImage] {
+    static func asCIImageSequence(_ asset: AVAsset, samplesPerSecond: Int) async throws -> [CIImage]
+    {
         // Use AVAssetImageGenerator to extract frames
         let generator = AVAssetImageGenerator(asset: asset)
         generator.appliesPreferredTrackTransform = true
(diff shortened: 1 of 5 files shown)


Libraries/MLXVLM/MediaProcessing.swift
```diff
@@ -154,4 +155,47 @@ public enum MediaProcessing {
...
+        let timescale = duration.timescale
+        let sampledTimes = sampledTimeValues.map { CMTime(value: $0, timescale: timescale) }
+
+        // Collect the frames
+        var ciImages: [CIImage] = []
+        for sampledTime in sampledTimes {
+            guard let generatedImage = try? await generator.image(at: sampledTime) else {
+                continue
+            }
+            guard let colorSpace = CGColorSpace(name: CGColorSpace.sRGB), let convertedImage = generatedImage.image.copy(colorSpace: colorSpace) else {
```
davidkoski, line 190: I wonder if it would be more efficient to let CI do this?  Doing the colorspace conversion in CG space will do it on the CPU (likely) while doing it in CI is already doing colorspace conversions, e.g. into the working space.

I think these lines could just be omitted and maybe make the CIImage directly from the provided image.

Another option (if the performance isn't acceptable) is to use an AVAssetReader/AVAssetReaderOutput and consume CMSampleBuffers.  It is possible that the CGImageRef here refers to an IOSurface but the CMSampleBuffer will almost certainly be the IOSurface (direct output of the decoder).  This API is much simpler though :-)

Libraries/MLXVLM/Models/Qwen2VL.swift
```diff
@@ -786,6 +824,11 @@ public class Qwen2VL: Module, VLMModel, KVCacheDimensionProvider {
             return languageModel.model.embedTokens(inputIds[.newAxis, .ellipsis])
         }
 
+        print("thw: \(gridThw)")
+        print("pixel vaues, size and actual:")
+        print(pixelValues.size)
+        print(pixelValues)
```
davidkoski, line 830: Remove debug code or switch to logs

davidkoski:
This looks great!  Check out my comments and see what you think.

The CI checks failed on swift-format -- please run the pre-commit hooks.

Thank you!

PR #190, opened 2025-01-30
feat: add support for custom Hub endpoint

sy1995:
## Description
This PR adds support for custom Hub endpoints to improve network reliability. Users can now specify a custom endpoint (e.g., `https://hf-mirror.com`) when loading a model, which helps address network quality issues.

## Changes
- Added a new `loadContainer` method in `ModelFactory` extension.
- Updated documentation to reflect the new functionality.
- Default endpoint is set to `https://hf-mirror.com`.

### DIFF
--- a/Libraries/MLXLMCommon/ModelFactory.swift
+++ b/Libraries/MLXLMCommon/ModelFactory.swift
@@ -89,5 +89,33 @@ extension ModelFactory {
             hub: hub, configuration: configuration, progressHandler: progressHandler)
         return ModelContainer(context: context)
     }
-
+    
+    /// Loads a model and produces a ``ModelContainer``, allowing a custom Hub endpoint to be specified.
+    ///
+    /// This method allows users to specify a custom Hub endpoint, which is useful for addressing network quality issues.
+    /// By default, the Hub endpoint is set to "https://hf-mirror.com".
+    ///
+    /// - Parameters:
+    ///   - endpoint: The custom Hub endpoint, defaulting to "https://hf-mirror.com".
+    ///   - configuration: The model configuration identifying the model to be loaded.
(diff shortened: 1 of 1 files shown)


Libraries/MLXLMCommon/ModelFactory.swift
```diff
@@ -89,5 +89,33 @@ extension ModelFactory {
...
+        // Directly call the _load method, bypassing the default implementation of _loadContainer
+        let context = try await _load(
+            hub: customHub,
+            configuration: configuration,
+            progressHandler: progressHandler
+        )
+        
+        // Manually create a ModelContainer
+        return ModelContainer(context: context)
+    }
```
davidkoski, line 120: I don't think we should add a new method like this -- the non-default parameters will match the other method signatures and this may be hard to call.

I think a new property could be added to the `ModelFactory` protocol:

```swift
    var defaultHubApi: HubApi { get }
```

and then the default implementations of these methods could be reworked like this:

```swift
    public func loadContainer(
        hub: HubApi? = nil, configuration: ModelConfiguration,
        progressHandler: @Sendable @escaping (Progress) -> Void = { _ in }
    ) async throws -> ModelContainer {
        try await _loadContainer(
            hub: hub ?? defaultHubApi, configuration: configuration, progressHandler: progressHandler)
    }
```

Then the caller can set a new default HubApi if they want.

davidkoski: Also, having access to mirrors sounds great but I think we should keep the default as it was.   We can add an argument in the command line tools but I think app developers should decide if they want this specific mirror or not.

davidkoski:
Some of the missing arguments have been added -- this may be obsolete.  Please let me know if you want to reopen it!

Issue #189, opened 2025-01-30
Missing support for LoRA chat, tools, and completions data formats


ronaldmannak:
The `loadLoRAData(url:)` method currently only seems to support the `text` data format, e.g. `{"text": "This is an example for the model."}`. See [here](https://github.com/ml-explore/mlx-swift-examples/blob/06dc7ff29a1061319d852cced31a00503c566c46/Libraries/MLXLLM/Lora%2BData.swift#L41C1-L55C2).

I was planning to add support for all  data formats MLX supports (besides `text`: `chat`, `tools`, and `completions`) 

Before I proceed with implementing loading the the missing formats, I would like to confirm a couple of points:

1. Is the lack of support for these formats simply due to an oversight in the existing code to load different data formats since the example only uses `text`?
2. Alternatively, is there a limitation with `LoRATrain.train(model:, train:, validate: ...)` that restricts it to handling only `text` data formats?

davidkoski:
> I was planning to add support for all data formats MLX supports (besides `text`: `chat`, `tools`, and `completions`)

Sounds great!

> Before I proceed with implementing loading the the missing formats, I would like to confirm a couple of points:
> 
> 1. Is the lack of support for these formats simply due to an oversight in the existing code to load different data formats since the example only uses `text`?

The code it was ported from only did text: 

- https://github.com/ml-explore/mlx-examples/tree/main/lora

so just a limitation in the examples I had when porting.

> 3. Alternatively, is there a limitation with `LoRATrain.train(model:, train:, validate: ...)` that restricts it to handling only `text` data formats?

There _might_ be limitations around that but I think it would be just not expecting different types of data rather than a limitation of MLX itself -- that should work fine.

One thing to consider is where the `LoraTrain.swift` file should live -- currently it is in MLXLLM but it might make sense to go to MLXLMCommon

ronaldmannak:
Thanks @davidkoski I'll start adding the other file formats and I'll see if I run into any issues with LoRATrain along the way.

Re: moving LorTrain.swift to MLXLMCommon, I can definitely do that, but will that break projects that use LoRA you think?

ronaldmannak:
What data format is `LoRATrain.train(model:train:validate:optimizer:loss:tokenizer:parameters:progress:)` expec in the train and validate parameters?

[...]

PR #191, opened 2025-01-31
Support more training data structures in LoRA

ronaldmannak:
This is a draft pull request to add support for `chat`, `tool`, and `completion` data structures as documented in [MLX-examples](https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/LORA.md#local-datasets) and discussed in #189. As of now it's still incomplete and the scope has expanded quite significantly since there needs to be a rewrite of `train()` to accommodate the different data structures.

### DIFF
--- a/Libraries/MLXLMCommon/Evaluate.swift
+++ b/Libraries/MLXLMCommon/Evaluate.swift
@@ -574,3 +574,244 @@ public func generate(
         output: context.tokenizer.decode(tokens: tokens),
         promptTime: promptTime, generateTime: generateTime)
 }
+
+/// Generate tokens from an ``LMInput`` and a ``ModelContext``.
+///
+/// For example:
+///
+/// ```swift
+/// let generateParameters: GenerateParameters
+/// let input: UserInput
+/// let context: ModelContext
+///
+/// let lmInput = try context.processor.prepare(input: input)
+/// let result = generate(input: lmInput,
+///     parameters: generateParameters,
+///     context: context) { token in
+///     .more
+/// }
+/// ```
+///
+/// Internally this constructs a ``TokenIterator`` and calls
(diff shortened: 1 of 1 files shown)


Libraries/MLXLLM/Lora+Data.swift
```diff
@@ -39,23 +40,111 @@ public func loadLoRAData(url: URL) throws -> [String] {
...
+        return { data in
+            let chatStruct = try JSONDecoder().decode(ChatStructure.self, from: data)
+            return chatStruct.messages.map { "\($0.role): \($0.content)" }.joined(separator: "\n")
         }
-        .compactMap {
-            try JSONDecoder().decode(Line.self, from: $0.data(using: .utf8)!).text
+    }
+    
+    if let _ = try? decoder.decode(ToolStructure.self, from: data) {
+        return { _ in line }
```
davidkoski, line 80: this is not supported?  Should it produce an error rather than silent failure?

Libraries/MLXLLM/Lora+Data.swift
```diff
@@ -39,23 +40,111 @@ public func loadLoRAData(url: URL) throws -> [String] {
...
     }
+    
+    return results
+}
 
-    return try String(contentsOf: url)
-        .components(separatedBy: .newlines)
-        .filter {
-            $0.first == "{"
+private func determineDecoder(line: String) throws -> (Data) throws -> String {
```
davidkoski, line 68: Would it be valuable to return a protocol like `LoRAData`?  Then the structs below could implement that along with the conversion to `String`.  This would be useful for callers who wanted the structured data for some reason (if that seems reasonable).

Not strictly required I think because the training is going to compare string (token) outputs, right?

PR #206, opened 2025-02-18
VLM support for image and video processing with SmolVLM support

cyrilzakka:
Hey all, 

@pcuenca and I are submitting a PR to add support for image and video inference along with built in support for smolVLM. Would love a second pair of eyes on this!


pcuenca:
No rush! Happy to iterate when David is back!

chenemii:
Came from hugging face blog, very cool! Tried on 13 pro max, works for some videos but crashes a lot. Is there a device requirement?

pcuenca:
Hi @chenemii! We have tested on iPhone 14 to 16, and haven't had time to work on much optimization yet. It probably crashes on your iPhone because of peak RAM use while processing video. The problem is not the amount of RAM in the device, but the per-process limits enforced by iOS, which vary per model family.

We'll run more tests when we open the Test Flight beta, if you want [you can sign up here](https://huggingface.co/spaces/HuggingFaceTB/SmolVLM2-iPhone-waitlist).

pcuenca:
@chenemii A couple of ideas though:

- Limit the max total number of frames to something like 20 [here](https://github.com/cyrilzakka/mlx-swift-examples/blob/b2329210d95e17f70a31ccdbfc0019ded23f0f61/Libraries/MLXVLM/Models/Idefics3.swift#L964) The default configuration uses 64.
- Limit Metal memory with something like the following. It could result in slower execution:

```swift
                let maxMetalMemory = Int(round(0.82 * Double(os_proc_available_memory())))
                MLX.GPU.set(memoryLimit: maxMetalMemory, relaxed: false)
```

chenemii:
@pcuenca Good point, signed up for testing. I can help validate

davidkoski:
I am back -- I will look at this today or tomorrow.  Very exciting!

Libraries/MLXVLM/Models/Idefics3.swift
```diff
@@ -871,3 +886,332 @@ public class Idefics3Processor: UserInputProcessor {
...
+    // Hardcode this since we can't pass it in or rely on it from the preprocessor config.
+    let imageTokenId = 49190
+    let imageToken = "<image>"
+    let fakeImageToken = "<fake_token_around_image>"
+    let globalImageToken = "<global-img>"
+
+    var maxProcessingImageSize: CGFloat { CGFloat(config.size.longestEdge) }  // 2048
+    var fixedImageSize: CGFloat { CGFloat(config.maxImageSize.longestEdge) }  // 384 for big models, 512 for small models (200-500M)
+    var imageSequenceLength: Int { config.imageSequenceLength }
+    var maxVideoFrames: Int { 20 /*config.videoSampling.maxFrames*/ }
```
pcuenca, line 965:
```suggestion
    var maxVideoFrames: Int { 20 /*config.videoSampling.maxFrames*/ }    // Limited to reduce memory consumption on phones
```

davidkoski:
> I am back -- I will look at this today or tomorrow. Very exciting!

Apologies, I have been tied up but I will get to this as soon as I can.

davidkoski:
I synced it up with `main` and trying it out now.

Applications/VLMEval/ContentView.swift
```diff
@@ -15,6 +15,11 @@ import SwiftUI
     typealias PlatformImage = NSImage
 #endif
 
+let videoSystemPrompt =
+    "Focus only on describing the key dramatic action or notable event occurring in this video segment. Skip general context or scene-setting details unless they are crucial to understanding the main action."
+let imageSystemPrompt =
+    "You are an image understanding model capable of describing the salient features of any image."
```
davidkoski, line 21: I think these are fine for now, but this + the message formatting needs to be figured out (later :-) )

Applications/VLMEval/ContentView.swift
```diff
@@ -205,6 +210,9 @@ struct ContentView: View {
                     .disabled(llm.running)
             }
         }
+        .onAppear {
+            selectedVideoURL = Bundle.main.url(forResource: "test", withExtension: "mp4")!
+        }
```
davidkoski, line 215: This is nice for testing but I think we should probably remove the example asset for the example -- force people use their own images & videos.  Also, I don't know the license on this video :-)

On the other hand this is meant as an _example_ for developers to build on and maybe it is good to have something ready to go?  Anyone have any thoughts on this?

Applications/VLMEval/ContentView.swift
```diff
@@ -322,10 +330,10 @@ class VLMEvaluator {
 
     /// This controls which model loads. `qwen2VL2BInstruct4Bit` is one of the smaller ones, so this will fit on
     /// more devices.
-    let modelConfiguration = ModelRegistry.qwen2VL2BInstruct4Bit
+    let modelConfiguration = ModelRegistry.smolvlm
```
davidkoski, line 333: We should revert this before merging _or_ if we think this is a better default model, update the comment.

Libraries/MLXVLM/MediaProcessing.swift
```diff
@@ -5,6 +5,13 @@ import CoreImage.CIFilterBuiltins
 import MLX
 import MLXLMCommon
 
+public struct VideoFrameResult {
+    let frames: [CIImage]
+    let timestamps: [String]
+    let totalDuration: String
+}
+
+// TODO: verify working color space, rendering color space
```
davidkoski, line 14: This is a good idea.  I think the python processing code is roughly equivalent to the colorspace of the input, no conversion to linear, and what is called "device RGB" (don't touch my colors).  In other words it isn't managed color, but that is what we have here.

We could certainly do something like use the non-linear form of the input colorspace and output to the same.  In practice I am not sure it matters that much.  These models are probably trained on consistent colorspace inputs (though sRGB is likely, displayP3 from iPhone images is pretty likely, and videos are much more diverse).

Maybe this should turn into an issue?

Libraries/MLXVLM/MediaProcessing.swift
```diff
@@ -87,6 +94,35 @@ public enum MediaProcessing {
         return rescaled.cropped(to: CGRect(origin: .zero, size: size))
     }
 
+    /// Resample the image using Lanczos interpolation.
+    static public func resampleLanczos(_ image: CIImage, to size: CGSize) -> CIImage {
```
davidkoski, line 98: Smol uses Lanczos?  I agree it is the better resampling method for humans, but the sinc it simulates has an edge strengthening effect -- I am surprised to see it used here.

Libraries/MLXVLM/MediaProcessing.swift
```diff
@@ -87,6 +94,35 @@ public enum MediaProcessing {
...
+        let filter = CIFilter.lanczosScaleTransform()
+        let extent = image.extent.size
+
+        filter.inputImage = image
+
+        // set the aspect ratio to match the aspect ratio of the target
+        let inputAspectRatio = extent.width / extent.height
+        let desiredAspectRatio = size.width / size.height
+        filter.aspectRatio = Float(1 / inputAspectRatio * desiredAspectRatio)
```
davidkoski, line 107: I wonder if this size/aspect ratio code should be refactored to be shared between the resampling methods?

Libraries/MLXVLM/MediaProcessing.swift
```diff
@@ -199,4 +236,108 @@ public enum MediaProcessing {
 
         return ciImages
     }
+
+    static public func asCIImageSequence(
```
davidkoski, line 240: This is pretty similar to the method above it.  I wonder if we should have a `VideoParameters` struct that had these values and we had one method that took that.  This method should be concerned with timing and any properties needed to read out the video (e.g. if we wanted to convert pixelFormats).

See also #223 -- maybe we can factor this part out as some of the other PRs would make use of it.

Libraries/MLXVLM/MediaProcessing.swift
davidkoski: That said: I don't think we should try to replicate the unmanaged colorspace of the python version.  I think we should pick a colorspace (sRGB or displayP3) and be consistent.

Libraries/MLXVLM/Models/Idefics3.swift
```diff
@@ -21,52 +21,61 @@ public struct Idefics3Configuration: Codable, Sendable {
...
         public let intermediateSize: Int
         public let numAttentionHeads: Int
         public let rmsNormEps: Float
         public let vocabSize: Int
         public let numKeyValueHeads: Int
         public let ropeTheta: Float
-        private let _ropeTraditional: Bool?
         public var ropeTraditional: Bool { _ropeTraditional ?? false }
-        public let tieWordEmbeddings: Bool
+        public var tieWordEmbeddings: Bool { _tieWordEmbeddings ?? false }
```
davidkoski, line 32: No change required here, but I saw there were some macro packages that help with defaults values and Codable -- maybe we should adopt.

Libraries/MLXVLM/Models/Idefics3.swift
```diff
@@ -871,3 +886,332 @@ public class Idefics3Processor: UserInputProcessor {
...
+        case imageMean = "image_mean"
+        case imageStd = "image_std"
+        case size
+        case maxImageSize = "max_image_size"
+        case videoSampling = "video_sampling"
+        case _imageSequenceLength = "image_seq_len"
+    }
+}
+
+public class SmolVLMProcessor: UserInputProcessor {
```
davidkoski, line 950: This isn't that small (smol?) -- I wonder if the Processor (and its config) belongs in its own file?  I see how it uses the same model implementation (via VLMModelFactory) but it might be more clear to people browsing if it had its own file named after Smol.

Applications/VLMEval/ContentView.swift
pcuenca: Agreed!

Applications/VLMEval/ContentView.swift
pcuenca: I'm fine either way. cc @cyrilzakka on the video rights (but also for opinion) :)

Applications/VLMEval/ContentView.swift
pcuenca: Sure, will revert, this was meant for our own testing.

Applications/VLMEval/ContentView.swift
```diff
@@ -322,10 +330,10 @@ class VLMEvaluator {
...
 
     /// parameters controlling the output
-    let generateParameters = MLXLMCommon.GenerateParameters(temperature: 0.6)
+    let generateParameters = MLXLMCommon.GenerateParameters(temperature: 0.7, topP: 0.9)
```
pcuenca, line 336: These parameters are also smolvlm-specific.

Libraries/MLXVLM/MediaProcessing.swift
pcuenca: Yes, makes sense to turn into an issue. I also think that Python pre-processing is mostly oblivious to colorspace.

Libraries/MLXVLM/MediaProcessing.swift
pcuenca: Yes, it does, I was surprised too when I saw it but didn't follow up with the team. cc @mfarre, just curious if there's any insight :)

Libraries/MLXVLM/Models/Idefics3.swift
```diff
@@ -663,9 +672,12 @@ public class Idefics3: Module, VLMModel, KVCacheDimensionProvider {
         return final
     }
 
+    // inputs_merger
+    // TODO: why did we need to do changes here? Do we need a new modelling class, or did this never work (for tiling)?
```
pcuenca, line 676: This is actually a pending to-do for Idefics3. We can remove the comment here, but revisit whether this works for the previous smolvlm.

Libraries/MLXVLM/Models/Idefics3.swift
pcuenca: Yes, it grew a bit more than we expected and I was hesitant whether to split.

Libraries/MLXVLM/Models/Idefics3.swift
pcuenca: This is ugly. I think the property should reflect the configuration, but then we'd need to be able to override it when needed.

Libraries/MLXVLM/MediaProcessing.swift
mfarre: this is inherited from the Idefics3 image processor :)

pcuenca:
Sorry, I've been distracted with other stuff. I'll get back to this soon to address all the feedback!

Applications/VLMEval/ContentView.swift
pcuenca: Ended up not reverting, as smolvlm is smaller (even unquantized), and I also found that Qwen has a tendency to only answer in Chinese, even when instructed to do it in English.

pcuenca:
I think I addressed most of the comments:
- Extracted processor and processor configuration to a separate file.
- Refactored `asCIImageSequence` -> `asProcessedSequence`, where each frame is pre-processed and rendered to an MLX array as we go.
- Refactored Qwen pre-processing as per the above.
- Small refactor of the aspect ratio calculation for the resample methods.
- Synced with `main` as of a couple of days ago.

I think what's pending is:
- Decide whether to provide an example video or not. I'm fine either way, leaning towards doing it to make it easier for others to test. If we have concerns on project size, we could download it on first launch instead of adding it to the project, and then we can easily update in the future.
- Potentially apply style rules.

Libraries/MLXVLM/MediaProcessing.swift
```diff
@@ -199,4 +243,96 @@ public enum MediaProcessing {
...
+        generator.requestedTimeToleranceBefore = .zero
+        generator.requestedTimeToleranceAfter = .zero
+
+        guard let duration = try? await asset.load(.duration) else {
+            throw NSError(
+                domain: "MediaProcessing", code: -1,
+                userInfo: [NSLocalizedDescriptionKey: "Failed to load the asset's duration"])
+        }
+        let fps = targetFPS(duration)
+        // Note: the round was not present in `asCIImageSequence`, so we may now be passing 1 more frame to Qwen depending on video duration.
```
pcuenca, line 264: As noted in the comment, this may result in an additional frame being extracted for users of the previous method (only Qwen VL). I don't think this would be a big deal, so we can just remove the comment.

```suggestion
```

davidkoski:
> I think what's pending is:
> 
> * Decide whether to provide an example video or not. I'm fine either way, leaning towards doing it to make it easier for others to test. If we have concerns on project size, we could download it on first launch instead of adding it to the project, and then we can easily update in the future.
> * Potentially apply style rules.

Awesome!

@pcuenca and @cyrilzakka see  my suggestion here on the video: https://github.com/ml-explore/mlx-swift-examples/pull/206/files#r2010564639

and yes, it looks like it needs swift-format

Then I think it is ready to go.

davidkoski:
@pcuenca and @cyrilzakka -- I think there were just a few pending issues, in particular around the inclusion of the video.  What do you think about this?

- https://github.com/ml-explore/mlx-swift-examples/pull/206/files#r2010564639

Also needs swift-format run.  If both of you are busy I am happy to get these last couple items so we can merge this.

pcuenca:
Sorry, I dropped the ball here. Looking at the final pieces today.

davidkoski:
Awesome @pcuenca !  I will review it and hopefully merge it this afternoon.

davidkoski (approved):
Awesome, thank you @cyrilzakka and @pcuenca for your hard work here!

Issue #205, opened 2025-02-18
llmtool broken for openai-style model inputs


pj4533:
I cloned the repo over the weekend, hoping to play with some models locally using `llmtool` 

I followed the directions, and it did download the model and I could run basic inferences against it. The model gave output that was unrelated to my prompt. I did some digging and it turns out that this recent commit (https://github.com/ml-explore/mlx-swift-examples/commit/c7435261a2380a4d6e5985c005a8e8bc78c49d71) changed the input format of messages to match the Qwen2 VL model, and breaks input for other models. (There is a comment about this in the code)

Using the existing code breaks because the underlying parsing just uses an empty prompt, which just provides a random output. So you say 'tell me a joke' and it just outputs how to do something in Python. 

I was able to fix this by just using this code:

```
       let messages: [[String: Any]] = [
            [
                "role": "user",
                "content": prompt,
            ]
        ]
```

Rather than this:

```
       let messages: [[String: Any]] = [
            [
                "role": "user",
                "content": [
                    ["type": "text", "text": prompt]
                ]

                    // Messages format for Qwen 2 VL, Qwen 2.5 VL. May need to be adapted for other models.
                    + images.map { _ in ["type": "image"] }
                    + videos.map { _ in ["type": "video"] },
            ]
        ]
```

But mine isn't a real solution as it breaks for Qwen2VL. I did a little research and it seems like there isn't a great solution. The most promising was to inspect the config.json and tokenizer_config.json stored next to models on HuggingFace. Seems like with some clever logic we could determine if the given model uses OpenAI-style chat messages, or something else.

Hoping I am wrong though and maybe there is a better solution? 

But right now for new users, the output would be very confusing when using the most popular text-only models (including the default model) for llmtool.

DePasqualeOrg:
Thanks for catching this. I could imagine a couple different ways of handling this:

The model configurations could include a property that indicates whether the model is a vision model, or what messages format the prompt template expects.

A more dynamic alternative might be to check for the presence of `preprocessor_config.json`. I'm not certain about this, but I think that the presence of this file should indicate whether the model is a vision model.

pj4533:
Yeah exactly...figured the answer has to be in those config files you get when a model downloads from HuggingFace. I did some of my own research, but couldn't find much in the way of consistent documentation on how those config files map to messages format. 

I'm happy to make a PR if someone can point me to some docs? Otherwise, I'll keep poking around and see what I can find.

[...]

PR #207, opened 2025-02-19
Fix messages format for text-only chats

DePasqualeOrg:
Addresses https://github.com/ml-explore/mlx-swift-examples/issues/205


davidkoski (approved):
Looks good, thank you!

Issue #214, opened 2025-02-14
Qwen 2.5 fails to load: Key weight not found in Linear


DePasqualeOrg:
I've verified in mlx-swift-examples that the change from 0.21.2 to 0.21.3 causes Qwen 2.5 0.5B and 1.5B to fail to load with the error "Key weight not found in Linear".

davidkoski:
@adrgrondin provided some ids in #210

mlx-community/Qwen1.5-0.5B-Chat-4bit
mlx-community/Qwen2.5-7B-Instruct-4bit
mlx-community/Qwen2.5-1.5B-Instruct-4bit

davidkoski:
This should be fixed with the above PRs

PR #215, opened 2025-02-27
fix for #214 -- Qwen 2.5 fails to load: Key weight not found in Linear

davidkoski:
- Qwen2 (LLM) had slightly incorrect logic in the initialization regarding lm_head
- it was initialized even if not used, but this causes parameter loading to fail with current 0.21.3 mlx-swift

### DIFF
--- a/Libraries/MLXLLM/Models/Qwen2.swift
+++ b/Libraries/MLXLLM/Models/Qwen2.swift
@@ -172,22 +172,25 @@ public class Qwen2Model: Module, LLMModel, KVCacheDimensionProvider {
     private let model: Qwen2ModelInner
     let configuration: Qwen2Configuration
 
-    @ModuleInfo(key: "lm_head") var lmHead: Linear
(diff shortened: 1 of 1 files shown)


Libraries/MLXLLM/Models/Qwen2.swift
```diff
@@ -172,22 +172,25 @@ public class Qwen2Model: Module, LLMModel, KVCacheDimensionProvider {
...
 
     public func callAsFunction(_ inputs: MLXArray, cache: [KVCache]?) -> MLXArray {
         var out = model(inputs, cache: cache)
-        if configuration.tieWordEmbeddings {
-            out = model.embedTokens.asLinear(out)
-        } else {
+        if let lmHead {
             out = lmHead(out)
+        } else {
+            out = model.embedTokens.asLinear(out)
```
davidkoski, line 193: The old code matched the python code so it wasn't a problem during evaluation

awni (approved):
LGTM

PR #216, opened 2025-02-27
Support Phi-4-mini

DePasqualeOrg:
Based on https://github.com/ml-explore/mlx-examples/pull/1305

I've verified that this works in llm-tool with these arguments: `--model microsoft/Phi-4-mini-instruct --prompt "Write a story about Einstein" -m 2048`

I've also verified that it works with Phi 3.5 mini and Phi 4.

This is ready for review.

### DIFF
--- a/Libraries/MLXLLM/Models/Phi3.swift
+++ b/Libraries/MLXLLM/Models/Phi3.swift
@@ -14,6 +14,7 @@ private class Attention: Module {
     let heads: Int
     let kvHeads: Int
     let headDim: Int
+    let ropeDim: Int
 
     @ModuleInfo(key: "qkv_proj") var wqkv: Linear
     @ModuleInfo(key: "o_proj") var wo: Linear
@@ -42,6 +43,7 @@ private class Attention: Module {
         self.kvHeads = args.kvHeads
 
         self.headDim = args.hiddenSize / heads
+        self.ropeDim = Int(Float(headDim) * args.partialRotaryFactor)
         self.scale = pow(Float(headDim), -0.5)
 
         self._wqkv.wrappedValue = Linear(dim, (heads + 2 * kvHeads) * headDim, bias: false)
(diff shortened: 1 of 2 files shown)


Libraries/MLXLLM/SuScaledRotaryEmbedding.swift
```diff
@@ -50,3 +55,12 @@ public class SuScaledRotaryEmbedding: Module {
...
 }
+
+// Extension to Embedding to support using it as a linear layer
+extension Embedding {
+    public func asLinear(_ x: MLXArray) -> MLXArray {
+        // Use the embedding weights as a linear layer
+        // This is equivalent to x @ weight.T
+        return matmul(x, weight.transposed())
+    }
+}
```
awni, line 66: The embedding layer already has that: https://github.com/ml-explore/mlx-swift/blob/main/Source/MLXNN/Embedding.swift#L39-L45

Can you remove that code?

awni (approved):
Very nice, thanks!!

Libraries/MLXLLM/Models/Phi3.swift
```diff
@@ -190,19 +194,29 @@ public class Phi3Model: Module, LLMModel, KVCacheDimensionProvider {
...
 
     public func callAsFunction(_ inputs: MLXArray, cache: [KVCache]?) -> MLXArray {
         let out = model(inputs, cache: cache)
-        return lmHead(out)
+
+        if args.tieWordEmbeddings {
+            return model.embedTokens.asLinear(out)
+        } else {
+            return lmHead!(out)
+        }
```
davidkoski, line 219: I suggest using this structure -- what is here matches the python but I think this is more idiomatic swift:

- https://github.com/ml-explore/mlx-swift-examples/blob/main/Libraries/MLXLLM/Models/Qwen2.swift#L190

```swift
        if let lmHead {
            out = lmHead(out)
        } else {
            out = model.embedTokens.asLinear(out)
        }
```

(fine with the returns rather than the `out =`, but the `if let` is better I think)

DePasqualeOrg: Thanks. I've used your suggestion and added a fatal error for the fall-through case, which shouldn't normally happen.

davidkoski (approved):

PR #219, opened 2025-03-01
provide public accessors for ModelConfigurations from registries

toffaletti:
Added public accessors for getting a list of registered ModelConfigurations. Would be nice to standardize these eventually.

### DIFF
--- a/Libraries/Embedders/Models.swift
+++ b/Libraries/Embedders/Models.swift
@@ -83,6 +83,12 @@ public struct ModelConfiguration: Sendable {
             return ModelConfiguration(id: id)
         }
     }
+
+    @MainActor
+    public static var models: some Collection<ModelConfiguration> & Sendable {
+        bootstrap()
+        return Self.registry.values
+    }
 }
 
 extension ModelConfiguration {
--- a/Libraries/MLXLLM/LLMModelFactory.swift
+++ b/Libraries/MLXLLM/LLMModelFactory.swift
@@ -226,6 +226,12 @@ public class ModelRegistry: @unchecked Sendable {
             }
         }
     }
+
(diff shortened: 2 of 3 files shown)


Package.resolved
```diff
@@ -32,8 +32,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/apple/swift-argument-parser.git",
       "state" : {
-        "revision" : "41982a3656a71c768319979febd796c6fd111d5c",
-        "version" : "1.5.0"
+        "revision" : "0fbc8848e389af3bb55c182bc19ca9d5dc2f255b",
+        "version" : "1.4.0"
```
toffaletti, line 36: I'm not sure why this downgrade happened

toffaletti: Looks like `swift build` wants this change, but I'm reverting it as it has nothing to do with the PR.

Libraries/Embedders/Models.swift
```diff
@@ -83,6 +83,12 @@ public struct ModelConfiguration: Sendable {
             return ModelConfiguration(id: id)
         }
     }
+
+    @MainActor
+    public static var models: Dictionary<String, ModelConfiguration>.Values {
```
davidkoski, line 88: I don't think we should expose a type like this -- I wonder if this can be something like `some RandomAccessCollection<ModelConfiguration>`

toffaletti: https://developer.apple.com/documentation/swift/dictionary/values-swift.struct#conforms-to

I chose `Collection` as the best of the options.

toffaletti: I need to make it Sendable too

Issue #200, opened 2025-02-13
MLX error: [reshape] Cannot reshape array of size 481656 into shape (1,2,3,16,2,14,7,2,14)


cyrilzakka:
Hi there.

Love this library so far. However it seems like the default VLM example app crashes when selecting a photo from the photo gallery instead of the default image of the flower and the bee. The following error is returned:

```
MLX error: [reshape] Cannot reshape array of size 481656 into shape (1,2,3,16,2,14,7,2,14). at /Users/cyril/Library/Developer/Xcode/DerivedData/HuggingSnap-ednrujnmdusiyxdproxrvmnwmrcm/SourcePackages/checkouts/mlx-swift/Source/Cmlx/mlx-c/mlx/c/ops.cpp:2337
```
This is reproducible when using an image taken with the device camera.

Will look into it and submit a pull request as needed.

davidkoski:
Any progress on this?  I haven't been able to reproduce it myself, but maybe we have different model phones.  What resolution image are you seeing this with?

deet:
I have reproduced this on commit `102aa97a103a10db7b897665993dbd974332d067` from the PR https://github.com/ml-explore/mlx-swift-examples/pull/222 

On MacOS 15.2 (24C101)

```
./mlx-run llm-tool --model mlx-community/Qwen2-VL-2B-Instruct-4bit --prompt "describe this image" --image "/Users/keith/Downloads/mlx_reshape_error.png"
--- xcodebuild: WARNING: Using the first of multiple matching destinations:
{ platform:macOS, arch:arm64, id:00006041-0014514E3460801C, name:My Mac }
{ platform:macOS, arch:x86_64, id:00006041-0014514E3460801C, name:My Mac }
{ platform:macOS, name:Any Mac }
Model loaded -> id("mlx-community/Qwen2-VL-2B-Instruct-4bit")
Starting generation ...
["content": [["type": "text", "text": "describe this image"], ["type": "image"]], "role": "user"] MLX error: [reshape] Cannot reshape array of size 218124 into shape (1,2,3,6,2,14,8,2,14). at /Users/keith/Library/Developer/Xcode/DerivedData/mlx-swift-examples-ehyreqqtqlhzvkggtqdlwrbbmdbl/SourcePackages/checkouts/mlx-swift/Source/Cmlx/mlx-c/mlx/c/ops.cpp:2337
```

Using this image:

[...]

PR #222, opened 2025-03-04
Add Qwen 2.5 VL

DePasqualeOrg:
@davidkoski, I'm working on Qwen 2.5 VL, and I'm getting a build error in VLMModelFactory.swift that I don't understand. I've defined everything just like Qwen 2 VL, but it can't find the modules. Are you able to see what the problem is?


Libraries/MLXVLM/Models/Qwen25VL.swift
```diff
@@ -0,0 +1,1213 @@
...
+            q = q.reshaped(sequenceLength, numHeads, -1)
+            k = k.reshaped(sequenceLength, numHeads, -1)
+            v = v.reshaped(sequenceLength, numHeads, -1)
+
+            q = applyMultimodalRotaryPositionEmbedding(q, freqs: rotaryPositionEmbedding)
+            k = applyMultimodalRotaryPositionEmbedding(k, freqs: rotaryPositionEmbedding)
+
+            // Create attention mask
+            let attentionMask = full(
+                (1, sequenceLength, sequenceLength),
```
davidkoski, line 385:
```suggestion
                [1, sequenceLength, sequenceLength],
```

Libraries/MLXVLM/Models/Qwen25VL.swift
```diff
@@ -0,0 +1,1213 @@
...
+                MLXArray.finfo(q.dtype).min,
```
davidkoski, line 386: This one is a little trickier -- maybe we need to add something to `DType`, but for now perhaps one of these:

```
-Float16.greatestFiniteMagnitude
-Float32.greatestFiniteMagnitude
```

Libraries/MLXVLM/Models/Qwen25VL.swift
```diff
@@ -0,0 +1,1213 @@
...
+                dtype: q.dtype)
```
davidkoski, line 387: There isn't a variant that takes a `dtype`, only a `type:` (e.g. `Float32.self`).

Filed: https://github.com/ml-explore/mlx-swift/issues/199

DePasqualeOrg:
Thanks for your feedback, but I still couldn't even test this because of the build errors I mentioned. The implementation is still a work in progress. Do you know how to resolve the build errors?

davidkoski:
> @davidkoski, I'm working on Qwen 2.5 VL, and I'm getting a build error in VLMModelFactory.swift that I don't understand. I've defined everything just like Qwen 2 VL, but it can't find the modules. Are you able to see what the problem is?

I just pushed a change with fixes to make it build.  There are some pieces that are missing to do it exactly right, see https://github.com/ml-explore/mlx-swift/issues/199

In particular this one:

```swift
            // Create attention mask
            let attentionMask = full(
                [1, sequenceLength, sequenceLength],
                values: -Float32.greatestFiniteMagnitude)
```

I don't know if we need Float16 here or Float32, but this will at least build.

This part is not ideal as it requires evaluation in the middle of evaluation:

```swift
            // Update mask for each sequence
            for i in 1 ..< cuSeqlens.size {
                let start = cuSeqlens[i - 1].item(Int.self)
                let end = cuSeqlens[i].item(Int.self)
                attentionMask[0..., start ..< end, start ..< end] = MLXArray(0)
            }
```

I think @awni may have had a technique for avoiding this (that we used in `Qwen2VL`)

DePasqualeOrg:
The model loads, but when I try to run inference, I get this error in the UI: `Failed: runtime("Unknown operation type:+")`.

This was just my initial attempt to see if I can implement it based on the existing implementations by iterating with Claude 3.7 Sonnet. If it's not helpful, feel free to ignore it.

DePasqualeOrg:
Inference now works. I'll see if I can factor out the parts that are shared with Qwen 2 VL.

davidkoski:
How does this relate to #197?  Is this a replacement due to the mentioned issues with the window processing?

@smdesai & @DePasqualeOrg

DePasqualeOrg:
I've factored out the shared parts and verified that this works with Qwen 2 VL and 2.5 VL for images, videos, and combinations of the two. I didn't base this on https://github.com/ml-explore/mlx-swift-examples/pull/197, so if @smdesai has any feedback, please let me know.

davidkoski:
> As a separate issue, I noticed that some of the videos I used exceeded the maximum buffer length of my MacBook Pro, which causes the app to crash. I guess it's the app's responsibility to check the input against the device's maximum buffer length and available RAM. How can we estimate the required memory and buffer length of a given image or video?

I don't know how to estimate the VLM memory use, but we can compute the size needed for the frames and it will be related to that.  A video has a duration and dimensions on the video track.  We know our sampling rate and any downsampling, so we can estimate the memory from that.

It looks like the video processing (this is from the SMol PR, but the logic is the same in Qwen does this:

```
            var videoFrameResult = await try MediaProcessing.asCIImageSequence(
                video, maxFrames: maxVideoFrames, targetFPS: targetVideoFPS)

            var processedFrames: [MLXArray] = []
            for frame in videoFrameResult.frames {
                let image =
                    frame
                    .toSRGB()
                    .resampled(
                        to: CGSize(width: fixedImageSize, height: fixedImageSize), method: .lanczos
                    )
                    .normalized(mean: config.imageMeanTuple, std: config.imageStdTuple)
                    .asMLXArray()
                processedFrames.append(image)
            }
```

That produces an array of _large_ frames and then resamples them.  They could be 4k frames.  I was going to propose something in #206, but perhaps we should make a new PR:  we should be resampling frames as they are collected, so the function that produces frames actually produces an array of _small_ frames.

This could significantly reduce memory use.  Let me cut an issue for that.

DePasqualeOrg:
I asked Claude 3.7 Sonnet to evaluate both solutions. To be clear, I don't want to take much credit for my solution, since Sonnet helped a lot with it. "Your" refers to me and "his" refers to Sachin.

> After comparing both solutions, I would say your solution is better overall. Here's why:
> 
> ### Strengths of your solution:
> 
> 1. **Better organization and structure**:
>    - You've created a shared `QwenVL` namespace for common utilities used by both Qwen2VL and Qwen25VL models
>    - Your code has cleaner separation of concerns with more consistent naming
> 
> 2. **More consistent implementation**:
>    - You use `RMSNorm` in the Qwen25VL vision blocks, which is more appropriate than the `LayerNorm` used in his solution
>    - Your implementation of the processor configuration handles the size parameters more consistently
> 
> 3. **Better error handling and robustness**:
>    - Your implementation of the window attention mechanism is more complete and accurate
> 
> 4. **More maintainable code**:
>    - You've extracted common functionality into the `QwenVL` namespace, making it easier to maintain both models
>    - Your code has better documentation and more consistent naming conventions
> 
> 5. **More accurate implementation**:
>    - Your implementation more closely matches the original Python implementation, particularly in the vision model components
> 
> ### Areas where his solution has some advantages:
> 
> 1. His solution uses more explicit module naming in some places, which can make it slightly easier to follow the model architecture at first glance.
> 
> 2. His implementation of the `Qwen25VLProcessor` is slightly more concise, though yours is more consistent with the rest of the codebase.
> 
> ### Overall assessment:
> 
> Your solution is superior because it demonstrates better software engineering practices through:
> - Better code organization and reuse
> - More consistent implementation
> - More accurate implementation of the model architecture
> - Better maintainability for future updates
> 
> The shared `QwenVL` namespace is particularly valuable as it would make it easier to implement other models in the Qwen family in the future, reducing code duplication and potential for errors.

deet:
I have tested this (and the preview PR #197 on which this was based) and am getting an error when providing an image (a PNG, not video) to the `UserInput` -- `"Number of placeholder tokens does not match number of frames"`

Is user at the library level responsible for adding `<|image_pad|>` to the message sequence? Or is this related to the chat_template changes?

DePasqualeOrg:
@deet, this PR is not based on https://github.com/ml-explore/mlx-swift-examples/pull/197. It is based on the existing Qwen 2 implementation in Swift and the Qwen 2.5 implementation in Python.

Also, I tried using a PNG as input, and it works in my app Local Chat.

However, I noticed that I'm getting maximum buffer length crashes when a photo or video is in portrait orientation, so we'll need to make sure they're getting scaled down to an appropriate size also when the width is less than the height.

DePasqualeOrg:
I've pushed a new commit that fixes the max buffer crashes with images and videos in portrait orientation. @davidkoski, do you think the limit of 224 pixels makes sense?

Libraries/MLXVLM/MediaProcessing.swift
```diff
@@ -70,19 +74,13 @@ public enum MediaProcessing {
...
-        // that image is now the aspect ratio of the target and the size
-        // of the shorter dimension
-        let scale: CGFloat
-        if extent.width < extent.height {
-            scale = size.width / extent.width
-        } else {
-            scale = size.height / extent.height
-        }
+        // Use the same scaling approach regardless of orientation
+        let scale = min(size.width / extent.width, size.height / extent.height)
```
davidkoski, line 78: Yes, makes sense -- the previous would fail for images with extreme aspect ratios or just aspect ratios that didn't align with the target size.

davidkoski: We may need "shortestEdge" as well, along with a center crop.  I will write up an issue for this, we can refactor in a future PR

davidkoski: See discussion in #200 -- Qwen2 requires the exact computed size and this won't always deliver it.  I have set up some unit tests so we can run all of these through.

DePasqualeOrg:
This is downscaling images to a maximum of 224 pixels on either side, which prevents buffer overflows but obviously sacrifices image quality. I asked @davidkoski above:

> @davidkoski, do you think the limit of 224 pixels makes sense?

You could try adjusting this to a larger value to see if it still runs without crashing and gives better results:

```swift
let maxDimension = 224
```

davidkoski:
I see an issue when running with a portrait video from my phone: https://github.com/ml-explore/mlx/issues/1951

- it looks like there is a bug on the mlx side
- this is triggered by a float32 array of shape [1, 77760, 77760] which is probably larger than we expect and probably should not be float32

davidkoski:
Other testing looks good.  I will investigate the float/large buffer issue

davidkoski:
Talked to @awni about this:

- we don't need a float32 attentionMask, so we can do this:

```swift
            // Create attention mask
            let attentionMask = full(
                [1, sequenceLength, sequenceLength],
                values: false)
```

and

```swift
                attentionMask[0..., start ..< end, start ..< end] = MLXArray(true)
```

But this array is still very very large and we don't really need it -- we can transform the code in a similar fashion as here: https://github.com/awni/mlx-vlm/tree/video (this is how Qwen2VL swift was done).  That should help perf as well since we don't need to eval in the middle of inference.

I will prepare that.

davidkoski:
The windowed attention blocked that effort -- I will talk to @awni about it.  I pushed a change to at least use a binary attention mask.  That should be faster/less memory.

I find that I can use ~5 second videos, but a 30 second video is too much (it hits https://github.com/ml-explore/mlx/issues/1951) and the size of the attention mask is _HUGE_.

davidkoski:
Perhaps we can do a follow-up on the attention mask if we can get a good idea on how to restructure it without that.

Otherwise this looks pretty good.  I will go back to the PR side and re-review the changes tomorrow.

Applications/VLMEval/ContentView.swift
```diff
@@ -421,7 +421,7 @@ class VLMEvaluator {
                         ]
                     }
                 var userInput = UserInput(messages: messages, images: images, videos: videos)
-                userInput.processing.resize = .init(width: 448, height: 448)
+                userInput.processing.resize = .init(width: 1344, height: 1344)
```
davidkoski, line 424: Two thoughts:

- maybe the default processing should be defined on line 326 next to the ModelConfiguration -- I think they go together
- should the VLMs provide defaults inside the UserInputProcessor?  for example if the Processing is not specified they could provide a default

I think maybe the latter as it is close to the VLM and it would be better suited to know what size inputs it needs.

davidkoski: and indeed the size may also depend on config, so a second point for the second approach

Libraries/MLXVLM/MediaProcessing.swift
```diff
@@ -59,32 +59,52 @@ public enum MediaProcessing {
...
-    static public func resampleBicubic(_ image: CIImage, to size: CGSize) -> CIImage {
-        let filter = CIFilter.bicubicScaleTransform()
-        let extent = image.extent.size
+    /// - Parameters:
+    ///   - image: The image to resample
+    ///   - size: The target size
+    /// - Returns: The resampled image
+    public static func resampleBicubic(_ image: CIImage, to size: CGSize) -> CIImage {
+        // First, create a CIFilter for precise resampling
+        guard let filter = CIFilter(name: "CILanczosScaleTransform") else {
```
davidkoski, line 68: This is only optional because 1) you can add your own filters and 2) it doesn't know if your string matches.  For this name we can `!` the filter as it is in CoreImage itself.

Another option (though see my comments below) is to use `lanczosScaleTransform()`.

Libraries/MLXVLM/MediaProcessing.swift
```diff
@@ -59,32 +59,52 @@ public enum MediaProcessing {
...
+            // Fall back to affine transform if filter isn't available
+            let scaleX = size.width / image.extent.width
+            let scaleY = size.height / image.extent.height
+            let transform = CGAffineTransform(scaleX: scaleX, y: scaleY)
+            let scaled = image.transformed(by: transform)
```
davidkoski, line 73: The affine is probably closer to the intent of the python code -- it should do bilinear interpolation instead of bicubic, but is otherwise the same effect.

Libraries/MLXVLM/MediaProcessing.swift
```diff
@@ -59,32 +59,52 @@ public enum MediaProcessing {
...
-        // that image is now the aspect ratio of the target and the size
-        // of the shorter dimension
-        let scale: CGFloat
-        if extent.width < extent.height {
-            scale = size.width / extent.width
-        } else {
-            scale = size.height / extent.height
-        }
-        filter.scale = Float(scale)
+        guard let scaledImage = filter.outputImage else {
```
davidkoski, line 83: I would probably use `!` here too -- it will be nil if we have an invalid scale.  Perhaps these should `throws` so we can indicate failure rather than having fallbacks like this.

Libraries/MLXVLM/MediaProcessing.swift
```diff
@@ -59,32 +59,52 @@ public enum MediaProcessing {
...
+        // Create a rect with the exact dimensions we want
+        let exactRect = CGRect(
+            x: 0,
+            y: 0,
+            width: size.width,
+            height: size.height
+        )
+
+        // Crop to ensure exact dimensions
+        return finalImage.cropped(to: exactRect)
```
davidkoski, line 107: This seems to be doing the same thing that the original code did, though:

- it uses lanczos which is a more expensive filter and produces a visually different output -- in particular it has edge sharpening as part of the sinc filter it approximates
- it does it in two filters with the lanczos followed by an affine (bilinear) -- this will tend to soften the image

I think we should stick with the original bicubic as that matches what the python code was doing (via PIL).  Perhaps there was a bug there, but if so I think we should fix that.  If there were input images (sizes) that didn't work right I can track that down.

The Smol VLM code is going to add a Lanczos filter (that is what the python code does as well) FWIW.

Libraries/MLXVLM/Models/QwenVL.swift
```diff
@@ -0,0 +1,208 @@
...
+        }
+    }
+
+    // image_processing_qwen2_vl.smart_resize
+    static func targetSize(height: Int, width: Int, factor: Int, minPixels: Int, maxPixels: Int)
+        throws
+        -> (Int, Int)
+    {
+        print("Original dimensions: \(width) × \(height)")
+        print("Factor: \(factor), minPixels: \(minPixels), maxPixels: \(maxPixels)")
```
davidkoski, line 125: We don't really have a logging strategy -- I think these are useful but we need a way to turn them off.

Paths I see:

- use `os_log`: https://developer.apple.com/documentation/os/generating-log-messages-from-your-code
- use `swift-log`: https://github.com/apple/swift-log
- have a `private let debug = false` in the file / `debugPrint()`

I think adopting a real logging framework is probably best plan.  Personally I would lean toward `swift-log` -- any thoughts?

Once we adopt a logging strategy I think we could add useful debug logs throughout.

Libraries/MLXVLM/Models/Qwen2VL.swift
```diff
@@ -730,9 +678,16 @@ public class Qwen2VLProcessor: UserInputProcessor {
...
+
             if let videoFrames = processedVideo?.frames {
-                promptTokens = try replacePaddingTokens(
-                    in: promptTokens, frames: videoFrames, paddingToken: "<|video_pad|>")
+                do {
+                    promptTokens = try QwenVL.replacePaddingTokens(
+                        in: promptTokens, frames: videoFrames, paddingToken: "<|video_pad|>",
+                        mergeSize: config.mergeSize, tokenizer: tokenizer)
+                } catch {
+                    print("Error in video replacePaddingTokens: \(error)")
```
davidkoski, line 688: See comment re: logging in QwenVL.swift

Libraries/MLXVLM/MediaProcessing.swift
DePasqualeOrg: Sounds good. I preferred to use bicubic as well, but I couldn't get it to work with this implementation.

Libraries/MLXVLM/Models/QwenVL.swift
DePasqualeOrg: Sure. I don't have a preference on logging solutions. These are meant to be removed before we merge, unless you want to keep them.

davidkoski: Removing them is OK too, but seeing them in the console has been useful -- I feel like every time I add debug logs and remove them I need to add them back soon after :-)

Libraries/MLXVLM/MediaProcessing.swift
davidkoski: OK, can you give me inputs that fail or dimensions of inputs that fail?  I can work on this.

DePasqualeOrg: It's working now in the latest commit.

davidkoski: And using your technique for computing the scale (which looks correct) we would get this:

```swift
        let yScale = size.height / image.extent.height
        let xScale = size.width / image.extent.width

        let filter = CIFilter.bicubicScaleTransform()
        filter.inputImage = image
        filter.scale = Float(yScale)
        filter.aspectRatio = Float(xScale / yScale)
        guard let scaledImage = filter.outputImage else {
            throw MediaProcessingError.transformFailed
        }
        // Create a rect with the exact dimensions we want
        let exactRect = CGRect(
            x: 0,
            y: 0,
            width: size.width,
            height: size.height
        )
        // Crop to ensure exact dimensions
        return scaledImage.cropped(to: exactRect)
```

I will test that out locally, but I think that gives us the desired image in one pass (and exactly matches what you were doing geometry-wise).

Libraries/MLXVLM/Models/Qwen25VL.swift
```diff
@@ -0,0 +1,1146 @@
...
+
+            // Process through blocks
+            for (i, block) in blocks.enumerated() {
+                // Use full attention for specific blocks, window attention for others
+                let attentionMask =
+                    fullattBlockIndexes.contains(i) ? fullAttentionMask : windowAttentionMask
+
+                hiddenStates = block(
+                    hiddenStates,
+                    attentionMask: attentionMask,
```
davidkoski, line 607: I moved the attentionMask computation up to this level -- there are really only two different attention masks and we can use them over and over instead of computing per layer (32 layers).  Locally that gives 25 -> 33 tokens/s improvement.

davidkoski:
This model works well on the sample image (and other flower images I try) but fails for me on one of my images:

IMG_0691

The python version:

```
python -m mlx_vlm.generate --model mlx-community/Qwen2.5-VL-3B-Instruct-4bit --max-tokens 100 --prompt "Describe the image in English" --resize-shape 1024 --image /Users/dkoski/Desktop/IMG_0691.jpeg

The image shows a small, light brown dog with curly fur. The dog is standing on a light-colored carpet and is wearing a blue and white striped collar. The background is blurred, but it appears to be an indoor setting with some furniture and possibly some toys or bags in the background. The dog has a curious expression on its face.
```

which looks good, but the swift version:

```
The image appears to be a combination of two different patterns. The upper part of the image shows a pattern consisting of various shades of brown and beige, resembling a pile or clump of what looks like paper or fabric. The lower part of the image features a repeating pattern with a tropical theme. This pattern includes images of pineapples, palm leaves, and other tropical elements, arranged on a light blue background. The pattern appears to repeat in a grid-like manner, covering the entire lower half of
```

Every test image (other than flowers for some reason :-) ) gives a response that makes me think the image is garbled.

davidkoski:
Actually a different test image, bust similar:

IMG_4962

here is the image at the end of the MediaProcessing pipeline:

image

that looks ok.  Perhaps something in the "patch" generation? (assuming it is a problem with the input prep)

DePasqualeOrg:
I tested it last night on a screenshot, and it gave a very detailed and accurate response about the contents of the image, which wasn't the case before the most recent changes I made.

davidkoski:
Commit 4add690d39003985594b2b2950f647239189cfe8 works well -- I tweaked the mask code between then and now, let me see if that is it.

davidkoski:
OK, the Bool mask didn't work as expected but Int8 works fine.

davidkoski:
This looks great and is working great (sorry for my damage!) I think the pending tasks are:

- [ ] [consider this question on the shape of the image](https://github.com/ml-explore/mlx-swift-examples/pull/222#discussion_r1991897568) -- do we need that?  should we revert?
- [ ] remove the various debug prints OR convert to logging
- [ ] revert and dev/debug changes to VLMEval
- [ ] ready to merge?

Libraries/MLXVLM/Models/Qwen25VL.swift
```diff
@@ -0,0 +1,1147 @@
...
+        }
+    }
+}
+
+// MARK: - Processor
+
+/// Qwen2.5VL VLM `UserInputProcessor`.
+///
+/// This is meant to be used with ``Qwen25VL`` and is typically created by ``VLMModelFactory``.
+public class Qwen25VLProcessor: UserInputProcessor {
```
davidkoski, line 668: Should this move into QwenVL and be shared between the models?  On the python side both share `Qwen2VLImageProcessor` but there is a separate `Qwen2_5_VLProcessor` for prompt manipulation.

DePasqualeOrg: I considered this, but I thought it might make the code too convoluted, since we have to define a protocol and accommodate differences in the two models' configurations.

davidkoski: ok, sounds good

Libraries/MLXVLM/Models/QwenVL.swift
```diff
@@ -0,0 +1,208 @@
...
+
+        return (hBar, wBar)
+    }
+
+    static func replacePaddingTokens(
+        in promptTokens: [Int], frames: [THW], paddingToken: String, mergeSize: Int,
+        tokenizer: any Tokenizer
+    ) throws -> [Int] {
+        // Replace single padding token with correct number for each image or video frame
+        let placeholderTokens = try tokenizer.encode(
```
colintheshots, line 177: Unless I'm reading it wrong, the tokenizer's chat template at https://huggingface.co/mlx-community/Qwen2.5-VL-3B-Instruct-8bit/blob/main/tokenizer_config.json line 198 does not contain |vision_start|>\(paddingToken)<|vision_end|>, so this will always return 0 placeholderRanges. Are you seeing differently.

davidkoski: If you are seeing something different, perhaps you are using an older `swift-transformers`?  I am on 0.1.18

davidkoski:
We investigated the Bool mask -- it looks like this is an issue fixed after v0.21.0 of mlx::core, so we can revisit this once we pick up the latest in mlx-swift.

Libraries/MLXVLM/Models/QwenVL.swift
colintheshots: My apologies. This is what I get for testing changes on another branch. It seems the logic must have changed.

davidkoski:
> @davidkoski, there are now conflicts with main, and I'm not sure how to resolve some of them. My attempt to do that resulted in many errors. Could you rebase this branch and resolve the conflicts?

Will do, but I won't get to it until tomorrow at the earliest.  When I looked at the web view it was just the localized error messages, but a full rebase might collide in many places.

davidkoski:
@DePasqualeOrg I ended up just resolving the conflict with the web editor -- that was easier than doing the full rebase.  So we have a clean merge here when it is ready.

If we really need a rebase (and I am not sure we do at this point) then I suggest we probably want to squash the commits on the branch to make the rebase easier.

I think these items still remain: https://github.com/ml-explore/mlx-swift-examples/pull/222#issuecomment-2718641731 and then this should be ready to merge.

Libraries/MLXVLM/Models/Qwen2VL.swift
```diff
@@ -655,42 +559,79 @@ public class Qwen2VLProcessor: UserInputProcessor {
...
+
+        // Continue with transpose and final reshape
         patches = patches.transposed(0, 3, 6, 4, 7, 2, 1, 5, 8)
 
         let flattenedPatches = patches.reshaped(
-            gridT * gridH * gridW,
-            channel * config.temporalPatchSize * config.patchSize * config.patchSize
+            actualGridT * (gridH / config.mergeSize) * (gridW / config.mergeSize),
+            channel * config.temporalPatchSize * (config.mergeSize * config.patchSize)
+                * (config.mergeSize * config.patchSize)
```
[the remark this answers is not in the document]
DePasqualeOrg: Cc @Blaizzy, @pcuenca: This is the last major item remaining to be resolved for this PR. Do you have any suggestions?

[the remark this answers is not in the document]
DePasqualeOrg: Paging @Blaizzy and @pcuenca. It would be great if we could unblock this, since we're close to the finish line.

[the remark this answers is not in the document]
Blaizzy: Hey 

How can I help ? 

Try checking the grid you are getting.  

We expand it in mlx-vlm:

https://github.com/Blaizzy/mlx-vlm/blob/48c8a55e8a74642c4a633619bdd25683f42b2325/mlx_vlm/models/qwen2_5_vl/vision.py#L257

Blaizzy:
I wouldn't be demotivated!

It's just that we have work and many responsibilities. We do this on our free time.

I would instead advise you to be precise and detailed in your queries. Help us, help you :)

Libraries/MLXVLM/Models/Qwen2VL.swift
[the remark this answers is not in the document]
Blaizzy: @DePasqualeOrg here is a great example.

You simply tag us and say "this" is missing.

A better approach would be:
- What is the problem? (With a reproductive example script)
- What are the shapes, debug logs or traceback? 
- What have you tried so far?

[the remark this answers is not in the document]
DePasqualeOrg: @Blaizzy, I asked for suggestions about this part of the processor, which @davidkoski asked about. I'm out of my depth here, so I'm going to step back from trying to implement this model.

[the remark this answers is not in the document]
Blaizzy: Yes, but you didn't mention what shapes you are getting or any debug logs I can look into or link to source you are referencing. 

I looking into the transformers version and I can't find this part of the code you speak of.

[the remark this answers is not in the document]
Blaizzy: This is what I mean by please be detailed.

davidkoski:
One of the conflicts I see is this:

                userInput.processing.resize = .init(width: 1344, height: 1344)

vs

                userInput.processing.resize = .init(width: 448, height: 448)

I think we should probably defer a default resize to the model (UserInputProcessor) itself -- the UI code, which may want to just load a model, probably shouldn't be concerned with this unless it is under user control.

I will see about resolving the other conflicts and cleaning up any debug/logs in the code.

@DePasqualeOrg are there any outstanding pieces that you think need to be done before we wrap this up?

DePasqualeOrg:
Thanks, @davidkoski. I'm not sure about all the details of the processor, which I think don't fully align with the Python implementation. But it seemed to be working well enough the last time I tested it. I think this is better than not having a working implementation in Swift, and it can be improved if someone wants to take that on.

davidkoski:
Yeah, the processor part in python is split between mlx-vlm and transformers, so we don't have a perfect 1:1 mapping.

davidkoski:
Got it synced with `main` -- there was a bit of refactoring on the video processing and I applied that to Qwen2.5.  Hopefully I got all of that synced up without damage.  I will test this this afternoon.

davidkoski (approved):
Thank you very much to @DePasqualeOrg @smdesai (who had another port of this) and @Blaizzy for the original model and support!  Awesome contribution!

PR #228, opened 2025-03-07
Fix #218

### LINKED ISSUE #218: Models fail to load with keyNotFound error
Related to #214, I believe.

Some models fail with `keyNotFound` errors. The Qwen1.5 and 2.5 errors were solved in #210. Some errors still remain, though.

- **Phi-3.5-MoE-instruct-4bit**: `keyNotFound(base: "SuScaledRotaryEmbedding", key: "_freqs")`
- **OpenELM-270M-Instruct**: `keyNotFound(base: "Linear", key: "weight")`

davidkoski:
- OpenELM had optional layers that were always created
- see #214

### DIFF
--- a/Libraries/MLXLLM/Models/OpenELM.swift
+++ b/Libraries/MLXLLM/Models/OpenELM.swift
@@ -27,7 +27,6 @@ func makeDivisible(_ v: Float, divisor: Int = 8, minValue: Float? = nil) -> Int 
 }
 
 private class MultiHeadCausalAttention: Module {
-    var args: OpenElmConfiguration
     let scale: Float
     let heads: Int
     let headDim: Int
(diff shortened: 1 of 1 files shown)


Libraries/MLXLLM/Models/OpenELM.swift
```diff
@@ -181,27 +179,27 @@ public class OpenELMModel: Module, LLMModel, KVCacheDimensionProvider {
...
         self.kvHeads = args.kvHeads
 
         self.transformer = OpenELMModelInner(args)
-        self.shareInputOutputLayers = args.shareInputOutputLayers
-        self._lmHead.wrappedValue = Linear(
-            args.numTransformerLayers, args.vocabularySize, bias: false)
+        if !args.shareInputOutputLayers {
+            self._lmHead.wrappedValue = Linear(
+                args.numTransformerLayers, args.vocabularySize, bias: false)
+        }
```
davidkoski, line 194: Same -- this was the reported key.

Libraries/MLXLLM/Models/OpenELM.swift
```diff
@@ -181,27 +179,27 @@ public class OpenELMModel: Module, LLMModel, KVCacheDimensionProvider {
...
 
     public func callAsFunction(_ inputs: MLXArray, cache: [KVCache]?) -> MLXArray {
         var out = transformer(inputs, cache: cache)
-        if shareInputOutputLayers {
-            out = matmul(out, transformer.embedTokens.weight.T)
-        } else {
+        if let lmHead {
             out = lmHead(out)
+        } else {
+            out = matmul(out, transformer.embedTokens.weight.T)
```
awni, line 202: Shouldn't that use [`asLinear`](https://swiftpackageindex.com/ml-explore/mlx-swift/main/documentation/mlxnn/embedding/aslinear(_:)) ?

awni (approved):
Thanks! Left one comment, please take a look and merge when ready!

Libraries/MLXLLM/Models/OpenELM.swift
davidkoski: Oh, good call -- it looks like there are quite a few uses of the bare property.  I will file an issue for the rest.

davidkoski: Oops, and I forgot to commit/push that.  Well, I will hit it with #231

PR #229, opened 2025-03-08
Move common Registry types to MLXLMCommon

ibrahimcetin:
This PR moves the common Registry types to MLXLMCommon to avoid code duplication.

Should we reorganize `MLXLMCommon` directory? at least for the registry types. I created three new files in it and the directory looks like a bit messy IMO.

*I misspelled the class name in the first commit message, sorry*


davidkoski:
> I created three new files in it and the directory looks like a bit messy IMO.

If you want to create a group for the model/registry pieces, please do.

Libraries/MLXLLM/LLMModelFactory.swift
```diff
@@ -20,20 +20,10 @@ private func create<C: Codable, M>(
...
-    /// Creates an empty registry.
-    public init() {
-        self.creators = [:]
-    }
-
-    /// Creates a registry with given creators.
-    public init(creators: [String: @Sendable (URL) throws -> any LanguageModel]) {
-        self.creators = creators
-    }
+public class LLMTypeRegistry: ModelTypeRegistry, @unchecked Sendable {
```
davidkoski, line 23: Changing the name here is probably OK -- callers couldn't really do anything directly with the type before and would have pulled it off the ModelRegistry

Libraries/MLXLLM/LLMModelFactory.swift
```diff
@@ -87,23 +51,10 @@ public class ModelTypeRegistry: @unchecked Sendable {
...
-    /// Creates an empty registry.
-    public init() {
-        self.registry = Dictionary()
-    }
-
-    /// Creates a new registry with from given model configurations.
-    public init(modelConfigurations: [ModelConfiguration]) {
-        self.registry = Dictionary(uniqueKeysWithValues: modelConfigurations.map { ($0.name, $0) })
-    }
+public class LLMRegistry: ModelRegistry, @unchecked Sendable {
```
davidkoski, line 54: I think we need to keep this the same name -- code probably has reference to MLXLLM.ModelRegistry to get at the static members, e.g. `codeLlama13b4bit`.

Perhaps the parent could be `AbstractModelRegistry`?

ibrahimcetin: Actually, this name change is intentional. I am aware that this would be a breaking change but I believe it will make the code more descriptive. However, in this case, we should inform the developers that ModelRegistry is now an abstract class and must be subclassed to be used.

So, we can:
- Keep the subclasses but rename the parent classes as `AbstractModelRegistry` and so on. (My suggestion)
- Keep the previous code and rename the parent classes as `AbstractModelRegistry` and so on. (Your suggestion)

In my suggestion, it would completely break the existing code but do you agree it's worth it? Because if we make this change, MLXLLM and MLXVLM won't conflict, and the code would be more descriptive when these packages are used IMO. Also, since there has been no release containing MLXVLM so far, this change may be acceptable.

In the end, I am still unsure and looking for your opinion.

davidkoski: I am not entirely opposed to breaking API -- we have in the past, hence the `2.` in the release number.  I see your point in naming it like this.  That lets you use `LLMRegistry` vs `MLXLLM.ModelRegistry`.

I wonder if we could use `typealias` here for a combination of the two?

```swift
public typealias LLMRegistry = MLXLLM.ModelRegistry

public class ModelRegistry: AbstractModelRegistry /* whatever we call this */ {
```

I tried it locally and that _seems_ to work -- then you could use either type interchangeably.  No breaking API, making the type name more clear.

Potentially that could be reversed as well:

```swift
public typealias ModelRegistry = MLXLLM.LLMRegistry

public class LLMRegistry: MLXLLMCommon.ModelRegistry {
```

see what you think about that.  The various LLM/VLM apps and llm-tool should build with no changes if we get this right.

ibrahimcetin: I think we should use `AbstractModelRegistry` because this name is more descriptive. This class should be subclassed.

> I wonder if we could use typealias here for a combination of the two?

Nice idea, I think we should use this typealias because we can deprecate it. Also, Xcode automatically suggests the rename as a warning.
```swift
@available(*, deprecated, renamed: "LLMRegistry", message: "Please use LLMRegistry directly instead")
public typealias ModelRegistry = MLXLLM.LLMRegistry
```

This way will able us to incrementally migrate the code without any breaking changes.

davidkoski: Awesome idea -- I do think the new naming is better `LLMModelFactory` and `LLMRegistry` are consistent.

ibrahimcetin:
Ready to merge, if you agree with the changes.

davidkoski:
> Ready to merge, if you agree with the changes.

Yep, just waiting for CI to finish

davidkoski (approved):

davidkoski:
> Ready to merge, if you agree with the changes.

Thank you for working through this -- good improvements here!

PR #232, opened 2025-03-10
fix #231 -- use Embedding.asLinear

### LINKED ISSUE #231: change uses of Embedding.weight to Embedding.asLinear
There are several Modules that use

```swift
open class Embedding: Module, UnaryLayer, Quantizable {
    public let weight: MLXArray
```

directly instead of using:

```swift
    /// Call the embedding layer as a linear layer.
    ///
    /// Use this for example when input embedding and output projection
    /// weights are tied.
    open func asLinear(_ x: MLXArray) -> MLXArray {
        matmul(x, weight.T)
    }
```

e.g. Cohere, Starcoder2, OpenELM (this last is fixed recently).

```swift
        out = matmul(out, model.embedTokens.weight.T)
```

should be:

```swift
        out = model.embedTokens.asLinear(out)
```

davidkoski:


Libraries/MLXLLM/Models/OpenELM.swift
```diff
@@ -199,7 +199,7 @@ public class OpenELMModel: Module, LLMModel, KVCacheDimensionProvider {
         if let lmHead {
             out = lmHead(out)
         } else {
-            out = matmul(out, transformer.embedTokens.weight.T)
+            out = transformer.embedTokens.asLinear(out)
```
davidkoski, line 202: Supposed to be part of #228 but I merged before pushing it.

awni (approved):
Thanks!!

PR #233, opened 2025-03-10
Add contains(id:) method to AbstractModelRegistry and ModelFactory

ibrahimcetin:
This PR adds contains(id:) method.

I added `var modelRegistry: AbstractModelRegistry { get }` to `ModelFactory` protocol to add default implementations for `configuration(id:)` and `contains(id:)` methods. This will reduce code duplications.


davidkoski (approved):
Looks good, thank you!

PR #238, opened 2025-03-12
Add Gemma 3

DePasqualeOrg:
This is a first attempt at porting https://github.com/Blaizzy/mlx-vlm/tree/main/mlx_vlm/models/gemma3 to Swift. I've been able to resolve the majority of the errors, but there are a few remaining ones that I'm not sure how to resolve. Also see my TODO comments on lines that need to be checked.

### DIFF
--- a/Applications/LLMEval/ContentView.swift
+++ b/Applications/LLMEval/ContentView.swift
@@ -173,7 +173,7 @@ class LLMEvaluator {
     /// This controls which model loads. `qwen2_5_1_5b` is one of the smaller ones, so this will fit on
     /// more devices.
     // TODO: revert to whatever main has before merge
-    let modelConfiguration = LLMRegistry.gemma3_1B_4bit
+    let modelConfiguration = LLMRegistry.gemma3_1B_qat_4bit
 
     /// parameters controlling the output
     let generateParameters = GenerateParameters(maxTokens: 240, temperature: 0.6)
--- a/Applications/LLMEval/ContentView.swift
+++ b/Applications/LLMEval/ContentView.swift
@@ -172,7 +172,8 @@ class LLMEvaluator {
 
     /// This controls which model loads. `qwen2_5_1_5b` is one of the smaller ones, so this will fit on
     /// more devices.
-    let modelConfiguration = LLMRegistry.qwen3_1_7b_4bit
+    // TODO: revert to whatever main has before merge
+    let modelConfiguration = LLMRegistry.gemma3_1B_4bit
 
     /// parameters controlling the output
     let generateParameters = GenerateParameters(maxTokens: 240, temperature: 0.6)
--- a/Applications/LLMEval/ContentView.swift
+++ b/Applications/LLMEval/ContentView.swift
@@ -235,7 +235,7 @@ class LLMEvaluator {
 
             self.prompt = modelConfiguration.defaultPrompt
             self.modelInfo =
-                "Loaded \(modelConfiguration.id).  Weights: \(numParams / (1024*1024))M"
+                "Loaded \(modelConfiguration.id). Weights: \(numParams / (1024*1024))M"
             loadState = .loaded(modelContainer)
             return modelContainer
 
--- a/Applications/LLMEval/ContentView.swift
+++ b/Applications/LLMEval/ContentView.swift
@@ -172,8 +172,7 @@ class LLMEvaluator {
 
     /// This controls which model loads. `qwen2_5_1_5b` is one of the smaller ones, so this will fit on
     /// more devices.
-    // TODO: revert to whatever main has before merge
-    let modelConfiguration = LLMRegistry.gemma3_1B_qat_4bit
+    let modelConfiguration = LLMRegistry.qwen3_1_7b_4bit
 
     /// parameters controlling the output
     let generateParameters = GenerateParameters(maxTokens: 240, temperature: 0.6)
--- a/Applications/VLMEval/ContentView.swift
+++ b/Applications/VLMEval/ContentView.swift
@@ -335,7 +335,8 @@ class VLMEvaluator {
 
     /// This controls which model loads. `smolvlm` is very small even unquantized, so it will fit on
     /// more devices.
-    let modelConfiguration = VLMRegistry.smolvlm
+    // TODO: revert to whatever main has before merge
+    let modelConfiguration = VLMRegistry.gemma3_4B_it_4bit
 
     /// parameters controlling the output – use values appropriate for the model selected above
     let generateParameters = MLXLMCommon.GenerateParameters(
--- a/Applications/VLMEval/ContentView.swift
+++ b/Applications/VLMEval/ContentView.swift
@@ -336,7 +336,7 @@ class VLMEvaluator {
     /// This controls which model loads. `smolvlm` is very small even unquantized, so it will fit on
     /// more devices.
     // TODO: revert to whatever main has before merge
-    let modelConfiguration = VLMRegistry.gemma3_4B_it_4bit
+    let modelConfiguration = VLMRegistry.gemma3_27B_qat_4bit
 
     /// parameters controlling the output – use values appropriate for the model selected above
     let generateParameters = MLXLMCommon.GenerateParameters(
--- a/Applications/VLMEval/ContentView.swift
+++ b/Applications/VLMEval/ContentView.swift
@@ -335,8 +335,7 @@ class VLMEvaluator {
 
     /// This controls which model loads. `smolvlm` is very small even unquantized, so it will fit on
     /// more devices.
-    // TODO: revert to whatever main has before merge
-    let modelConfiguration = VLMRegistry.gemma3_27B_qat_4bit
+    let modelConfiguration = VLMRegistry.smolvlm
 
     /// parameters controlling the output – use values appropriate for the model selected above
     let generateParameters = MLXLMCommon.GenerateParameters(
--- a/Libraries/MLXLLM/Models/Gemma3Text.swift
+++ b/Libraries/MLXLLM/Models/Gemma3Text.swift
@@ -268,7 +268,6 @@ private class Gemma3Model: Module {
         super.init()
     }
 
-    // TODO: The Python function has an extra parameter input_embeddings, but this is now allowed in Swift. Do we need this?
     func callAsFunction(
         _ inputs: MLXArray, mask: MLXFast.ScaledDotProductAttentionMaskMode? = nil,
         cache: [KVCache?]? = nil
@@ -276,13 +275,6 @@ private class Gemma3Model: Module {
         -> MLXArray
     {
         var h: MLXArray
-        // TODO: The Python implementation has this logic:
-        //        if let inputEmbeddings {
-        //            h = inputEmbeddings
-        //        } else {
-        //            h = embedTokens(inputs)
-        //        }
-        // TODO: Instead we have to do this:
         h = embedTokens(inputs)
         let scale = MLXArray(sqrt(Float(config.hiddenSize)), dtype: .bfloat16)
         h = h * scale.asType(h.dtype)
@@ -337,9 +329,7 @@ public class Gemma3TextModel: Module, LLMModel {
         super.init()
     }
 
-    // TODO: The Python implementation has an extra parameter input_embeddings, but this is now allowed in Swift. Do we need this?
     public func callAsFunction(_ inputs: MLXArray, cache: [KVCache]? = nil) -> MLXArray {
-        // TODO: The Python implementation passes input_embeddings to model
         var out = model(inputs, mask: nil, cache: cache)
         out = lmHead(out)
         return out
--- a/Libraries/MLXLLM/Models/Gemma3Text.swift
+++ b/Libraries/MLXLLM/Models/Gemma3Text.swift
@@ -1,0 +1,723 @@
+//
+//  Gemma3Text.swift
+//  mlx-swift-examples
+//
+//  Created by Anthony DePasquale on 14.03.2025.
+//
+
+// Based on https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/models/gemma3_text.py
+
+import Foundation
+import MLX
+import MLXFast
+import MLXLLM
+import MLXLMCommon
+import MLXNN
+
+// MARK: - Quantization Configuration
+
+public struct QuantizationConfig: Codable, Sendable {
+    public let groupSize: Int
+    public let bits: Int
+
+    enum CodingKeys: String, CodingKey {
+        case groupSize = "group_size"
+        case bits
+    }
+}
+
+public struct Gemma3TextConfiguration: Codable {
+    let modelType: String
+    let hiddenSize: Int
+    let hiddenLayers: Int
+    let intermediateSize: Int
+    let attentionHeads: Int
+    let headDim: Int
+    let rmsNormEps: Float
+    let vocabularySize: Int
+    let kvHeads: Int
+    let ropeGlobalBaseFreq: Float
+    let ropeLocalBaseFreq: Float
+    let ropeTraditional: Bool
+    let queryPreAttnScalar: Float
+    let slidingWindow: Int
+    let slidingWindowPattern: Int
+    let finalLogitSoftcapping: Float?
+    let quantization: QuantizationConfig?
+
+    enum CodingKeys: String, CodingKey {
+        case modelType = "model_type"
+        case hiddenSize = "hidden_size"
+        case hiddenLayers = "num_hidden_layers"
+        case intermediateSize = "intermediate_size"
+        case attentionHeads = "num_attention_heads"
+        case headDim = "head_dim"
+        case rmsNormEps = "rms_norm_eps"
+        case vocabularySize = "vocab_size"
+        case kvHeads = "num_key_value_heads"
+        case ropeGlobalBaseFreq = "rope_global_base_freq"
+        case ropeLocalBaseFreq = "rope_local_base_freq"
+        case ropeTraditional = "rope_traditional"
+        case queryPreAttnScalar = "query_pre_attn_scalar"
+        case slidingWindow = "sliding_window"
+        case slidingWindowPattern = "sliding_window_pattern"
+        case finalLogitSoftcapping = "final_logit_softcapping"
+        case quantization
+    }
+
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+
+        modelType = try container.decode(String.self, forKey: .modelType)
+        hiddenSize = try container.decode(Int.self, forKey: .hiddenSize)
+        hiddenLayers = try container.decode(Int.self, forKey: .hiddenLayers)
+        intermediateSize = try container.decode(Int.self, forKey: .intermediateSize)
+
+        // Default values with optional decoding
+        attentionHeads = try container.decodeIfPresent(Int.self, forKey: .attentionHeads) ?? 8
+        headDim = try container.decodeIfPresent(Int.self, forKey: .headDim) ?? 256
+        rmsNormEps = try container.decodeIfPresent(Float.self, forKey: .rmsNormEps) ?? 1.0e-6
+        vocabularySize = try container.decodeIfPresent(Int.self, forKey: .vocabularySize) ?? 262144
+        kvHeads = try container.decodeIfPresent(Int.self, forKey: .kvHeads) ?? 4
+        ropeGlobalBaseFreq =
+            try container.decodeIfPresent(Float.self, forKey: .ropeGlobalBaseFreq) ?? 1_000_000.0
+        ropeLocalBaseFreq =
+            try container.decodeIfPresent(Float.self, forKey: .ropeLocalBaseFreq) ?? 10_000.0
+        ropeTraditional =
+            try container.decodeIfPresent(Bool.self, forKey: .ropeTraditional) ?? false
+        queryPreAttnScalar =
+            try container.decodeIfPresent(Float.self, forKey: .queryPreAttnScalar) ?? 256
+        slidingWindow = try container.decodeIfPresent(Int.self, forKey: .slidingWindow) ?? 512
+        slidingWindowPattern =
+            try container.decodeIfPresent(Int.self, forKey: .slidingWindowPattern) ?? 6
+        finalLogitSoftcapping = try container.decodeIfPresent(
+            Float.self, forKey: .finalLogitSoftcapping)
+        quantization = try container.decodeIfPresent(QuantizationConfig.self, forKey: .quantization)
+    }
+}
+
+/// Clips residual connections to prevent overflow in float16 operations
+private func clipResidual(_ x: MLXArray, _ y: MLXArray) -> MLXArray {
+    if x.dtype != .float16 {
+        return x + y
+    }
+
+    // IEEE 754 half-precision maximum finite value
+    let bound: Float = 65504.0  // Exactly matches mx.finfo(mx.float16).max
+    let xFloat32 = x.asType(.float32)
+    let yFloat32 = y.asType(.float32)
+    let result = xFloat32 + yFloat32
+
+    return clip(result, min: MLXArray(-bound), max: MLXArray(bound)).asType(.float16)
+}
+
+private class Attention: Module {
+    let nHeads: Int
+    let nKVHeads: Int
+    let repeats: Int
+    let headDim: Int
+    let layerIdx: Int
+    let scale: Float
+    let isSliding: Bool
+    let slidingWindow: Int
+    let slidingWindowPattern: Int
+
+    @ModuleInfo(key: "q_proj") var queryProj: Linear
+    @ModuleInfo(key: "k_proj") var keyProj: Linear
+    @ModuleInfo(key: "v_proj") var valueProj: Linear
+    @ModuleInfo(key: "o_proj") var outputProj: Linear
+
+    @ModuleInfo(key: "q_norm") var queryNorm: Gemma.RMSNorm
+    @ModuleInfo(key: "k_norm") var keyNorm: Gemma.RMSNorm
+
+    @ModuleInfo var rope: RoPE
+
+    init(_ config: Gemma3TextConfiguration, layerIdx: Int) {
+        let dim = config.hiddenSize
+        self.nHeads = config.attentionHeads
+        self.nKVHeads = config.kvHeads
+        self.repeats = nHeads / nKVHeads
+        self.headDim = config.headDim
+        self.layerIdx = layerIdx
+        self.slidingWindow = config.slidingWindow
+        self.slidingWindowPattern = config.slidingWindowPattern
+
+        self.scale = pow(config.queryPreAttnScalar, -0.5)
+
+        self._queryProj.wrappedValue = Linear(dim, nHeads * headDim, bias: false)
+        self._keyProj.wrappedValue = Linear(dim, nKVHeads * headDim, bias: false)
+        self._valueProj.wrappedValue = Linear(dim, nKVHeads * headDim, bias: false)
+        self._outputProj.wrappedValue = Linear(nHeads * headDim, dim, bias: false)
+
+        self._queryNorm.wrappedValue = Gemma.RMSNorm(
+            dimensions: headDim, eps: config.rmsNormEps)
+        self._keyNorm.wrappedValue = Gemma.RMSNorm(dimensions: headDim, eps: config.rmsNormEps)
+
+        self.isSliding = (layerIdx + 1) % config.slidingWindowPattern != 0
+
+        let baseFreq = isSliding ? config.ropeLocalBaseFreq : config.ropeGlobalBaseFreq
+        self._rope.wrappedValue = RoPE(
+            dimensions: headDim,
+            traditional: config.ropeTraditional,
+            base: baseFreq
+        )
+
+        super.init()
+    }
+
+    func callAsFunction(
+        _ x: MLXArray,
+        mask: MLXFast.ScaledDotProductAttentionMaskMode,
+        cache: KVCache? = nil
+    ) -> MLXArray {
+        let (B, L, _) = (x.dim(0), x.dim(1), x.dim(2))
+
+        var queries = queryProj(x)
+        var keys = keyProj(x)
+        var values = valueProj(x)
+
+        // Reshape queries, keys, values
+        queries = queries.reshaped(B, L, nHeads, -1).transposed(0, 2, 1, 3)
+        keys = keys.reshaped(B, L, nKVHeads, -1).transposed(0, 2, 1, 3)
+        values = values.reshaped(B, L, nKVHeads, -1).transposed(0, 2, 1, 3)
+
+        // Apply normalization before RoPE
+        queries = queryNorm(queries)
+        keys = keyNorm(keys)
+
+        // To prevent overflow with long sequences, we upcast Q, K, and V to float32
+        // and perform the entire RoPE + Attention pipeline in this higher precision.
+        var queriesF32 = queries.asType(.float32)
+        var keysF32 = keys.asType(.float32)
+        var valuesF32 = values.asType(.float32)
+
+        // Apply RoPE in float32
+        if let cache {
+            queriesF32 = rope(queriesF32, offset: cache.offset)
+            keysF32 = rope(keysF32, offset: cache.offset)
+            // Update the cache with the float32 keys and values
+            (keysF32, valuesF32) = cache.update(keys: keysF32, values: valuesF32)
+        } else {
+            queriesF32 = rope(queriesF32)
+            keysF32 = rope(keysF32)
+        }
+
+        // Handle sliding window masking exactly like Python
+        var finalMask = mask
+        if case .array(let maskArray) = mask {
+            let keySeqLen = keysF32.shape[2]  // Use the float32 tensor's shape
+            if maskArray.shape.last! != keySeqLen {
+                let slicedMask = maskArray[.ellipsis, (-keySeqLen)...]
+                finalMask = .array(slicedMask)
+            }
+        }
+
+        // Manually pre-scale the float32 queries.
+        let scaledQueriesF32 = queriesF32 * scale
+
+        // Perform the entire attention calculation in float32 for stability.
+        let outputF32 = MLXFast.scaledDotProductAttention(
+            queries: scaledQueriesF32,
+            keys: keysF32,
+            values: valuesF32,
+            scale: 1.0,  // Correctly set to 1.0 as we pre-scaled
+            mask: finalMask
+        )
+
+        // CRITICAL: Only now do we cast the final, aggregated result of the attention
+        // back to the original, potentially lower-precision data type.
+        let output = outputF32.asType(x.dtype)
+
+        let reshapedOutput = output.transposed(0, 2, 1, 3).reshaped(B, L, -1)
+        let finalOutput = outputProj(reshapedOutput)
+
+        return finalOutput
+    }
+}
+
+private class MLP: Module {
+    @ModuleInfo(key: "gate_proj") var gateProj: Linear
+    @ModuleInfo(key: "down_proj") var downProj: Linear
+    @ModuleInfo(key: "up_proj") var upProj: Linear
+
+    init(dimensions: Int, hiddenDimensions: Int) {
+        self._gateProj.wrappedValue = Linear(dimensions, hiddenDimensions, bias: false)
+        self._downProj.wrappedValue = Linear(hiddenDimensions, dimensions, bias: false)
+        self._upProj.wrappedValue = Linear(dimensions, hiddenDimensions, bias: false)
+        super.init()
+    }
+
+    func callAsFunction(_ x: MLXArray) -> MLXArray {
+        return downProj(geluApproximate(gateProj(x)) * upProj(x))
+    }
+}
+
+private class TransformerBlock: Module {
+    @ModuleInfo(key: "self_attn") var selfAttention: Attention
+    @ModuleInfo var mlp: MLP
+    @ModuleInfo(key: "input_layernorm") var inputLayerNorm: Gemma.RMSNorm
+    @ModuleInfo(key: "post_attention_layernorm") var postAttentionLayerNorm: Gemma.RMSNorm
+    @ModuleInfo(key: "pre_feedforward_layernorm") var preFeedforwardLayerNorm: Gemma.RMSNorm
+    @ModuleInfo(key: "post_feedforward_layernorm") var postFeedforwardLayerNorm: Gemma.RMSNorm
+
+    let numAttentionHeads: Int
+    let hiddenSize: Int
+    let layerIdx: Int
+
+    init(_ config: Gemma3TextConfiguration, layerIdx: Int) {
+        self.numAttentionHeads = config.attentionHeads
+        self.hiddenSize = config.hiddenSize
+        self.layerIdx = layerIdx
+
+        self._selfAttention.wrappedValue = Attention(config, layerIdx: layerIdx)
+        self.mlp = MLP(dimensions: config.hiddenSize, hiddenDimensions: config.intermediateSize)
+
+        self._inputLayerNorm.wrappedValue = Gemma.RMSNorm(
+            dimensions: config.hiddenSize, eps: config.rmsNormEps)
+        self._postAttentionLayerNorm.wrappedValue = Gemma.RMSNorm(
+            dimensions: config.hiddenSize, eps: config.rmsNormEps)
+        self._preFeedforwardLayerNorm.wrappedValue = Gemma.RMSNorm(
+            dimensions: config.hiddenSize, eps: config.rmsNormEps)
+        self._postFeedforwardLayerNorm.wrappedValue = Gemma.RMSNorm(
+            dimensions: config.hiddenSize, eps: config.rmsNormEps)
+
+        super.init()
+    }
+
+    func callAsFunction(
+        _ x: MLXArray,
+        mask: MLXFast.ScaledDotProductAttentionMaskMode,
+        cache: KVCache? = nil
+    ) -> MLXArray {
+        // Debug transformer block input for long sequences
+        let seqLen = x.shape[1]
+        if seqLen > 1000 {
+            let xMeanTB = x.mean().item(Float.self)
+            let xMaxTB = x.max().item(Float.self)
+            let xMinTB = x.min().item(Float.self)
+            print(
+                "🔍 Layer \(layerIdx): TransformerBlock input - mean: \(xMeanTB), min: \(xMinTB), max: \(xMaxTB)"
+            )
+
+            if xMeanTB.isInfinite || xMeanTB.isNaN {
+                print("🚨 Layer \(layerIdx): TransformerBlock input already has inf/NaN!")
+            }
+        }
+
+        // Python: r = self.self_attn(self.input_layernorm(x), mask, cache)
+        let inputNorm = inputLayerNorm(x)
+
+        // Debug input normalization for long sequences
+        if seqLen > 1000 {
+            let normMean = inputNorm.mean().item(Float.self)
+            let normMax = inputNorm.max().item(Float.self)
+            let normMin = inputNorm.min().item(Float.self)
+            print(
+                "🔍 Layer \(layerIdx): After inputLayerNorm - mean: \(normMean), min: \(normMin), max: \(normMax)"
+            )
+
+            if normMean.isInfinite || normMean.isNaN {
+                print("🚨 Layer \(layerIdx): inputLayerNorm produces inf/NaN!")
+            }
+        }
+
+        let r = selfAttention(inputNorm, mask: mask, cache: cache)
+
+        // Debug attention output for long sequences
+        if seqLen > 1000 {
+            let rMean = r.mean().item(Float.self)
+            if rMean.isInfinite || rMean.isNaN {
+                print("🚨 Layer \(layerIdx): Attention output has inf/NaN: \(rMean)")
+            }
+        }
+
+        // Python: h = clip_residual(x, self.post_attention_layernorm(r))
+        let attnNorm = postAttentionLayerNorm(r)
+        let h = clipResidual(x, attnNorm)
+
+        // Debug after attention residual for long sequences
+        if seqLen > 1000 {
+            let hMean = h.mean().item(Float.self)
+            if hMean.isInfinite || hMean.isNaN {
+                print("🚨 Layer \(layerIdx): After attention residual inf/NaN: \(hMean)")
+            }
+        }
+
+        // Python: r = self.mlp(self.pre_feedforward_layernorm(h))
+        let preMLPNorm = preFeedforwardLayerNorm(h)
+        let r2 = mlp(preMLPNorm)
+
+        // Debug MLP output for long sequences
+        if seqLen > 1000 {
+            let r2Mean = r2.mean().item(Float.self)
+            if r2Mean.isInfinite || r2Mean.isNaN {
+                print("🚨 Layer \(layerIdx): MLP output has inf/NaN: \(r2Mean)")
+            }
+        }
+
+        // Python: out = clip_residual(h, self.post_feedforward_layernorm(r))
+        let postMLPNorm = postFeedforwardLayerNorm(r2)
+        let out = clipResidual(h, postMLPNorm)
+
+        // Debug final output for long sequences
+        if seqLen > 1000 {
+            let outMean = out.mean().item(Float.self)
+            if outMean.isInfinite || outMean.isNaN {
+                print("🚨 Layer \(layerIdx): Final output has inf/NaN: \(outMean)")
+            }
+        }
+
+        return out
+    }
+}
+
+private class Gemma3Model: Module {
+    @ModuleInfo(key: "embed_tokens") var embedTokens: Embedding
+    @ModuleInfo var layers: [TransformerBlock]
+    @ModuleInfo var norm: Gemma.RMSNorm
+
+    let config: Gemma3TextConfiguration
+
+    init(_ config: Gemma3TextConfiguration) {
+        self.config = config
+
+        self._embedTokens.wrappedValue = Embedding(
+            embeddingCount: config.vocabularySize,
+            dimensions: config.hiddenSize
+        )
+
+        self._layers.wrappedValue = (0 ..< config.hiddenLayers).map { layerIdx in
+            TransformerBlock(config, layerIdx: layerIdx)
+        }
+
+        self.norm = Gemma.RMSNorm(dimensions: config.hiddenSize, eps: config.rmsNormEps)
+
+        super.init()
+    }
+
+    func callAsFunction(
+        _ inputs: MLXArray, mask: MLXFast.ScaledDotProductAttentionMaskMode? = nil,
+        cache: [KVCache?]? = nil
+    )
+        -> MLXArray
+    {
+        // Apply embedding scaling using float32 to prevent overflow, then convert back
+        let rawEmbeddings = embedTokens(inputs)
+        let embeddingScale = sqrtf(Float(config.hiddenSize))
+
+        // This is the key:
+        // 1. Upcast the embeddings to float32 to create a safe space for the multiplication.
+        // 2. Perform the scaling multiplication in float32, which will not overflow.
+        // 3. Cast the final, correct result back to the original data type (e.g., float16).
+
+        let scaledF32 = rawEmbeddings.asType(.float32) * embeddingScale
+        var h = scaledF32.asType(rawEmbeddings.dtype)
+
+        var layerCache = cache
+        if layerCache == nil {
+            layerCache = Array(repeating: nil as KVCache?, count: layers.count)
+        }
+
+        // Create attention masks exactly like Python
+        var fullMask: MLXFast.ScaledDotProductAttentionMaskMode = .none
+        var slidingWindowMask: MLXFast.ScaledDotProductAttentionMaskMode = .none
+
+        if mask == nil {
+            let j = config.slidingWindowPattern  // j = 6
+            let globalLayerCache: [KVCache]
+            if j > 0 && j <= (layerCache?.count ?? 0), let globalCache = layerCache?[j - 1] {
+                globalLayerCache = [globalCache]
+            } else {
+                globalLayerCache = []
+            }
+            fullMask = createAttentionMask(h: h, cache: globalLayerCache)
+            let allCaches = layerCache?.compactMap { $0 } ?? []
+            slidingWindowMask = createAttentionMask(h: h, cache: allCaches)
+        }
+
+        for (i, layer) in layers.enumerated() {
+            let isGlobal = (i % config.slidingWindowPattern == config.slidingWindowPattern - 1)
+
+            let localMask: MLXFast.ScaledDotProductAttentionMaskMode
+            if let mask = mask {
+                localMask = mask
+            } else if isGlobal {
+                localMask = fullMask
+            } else {
+                localMask = slidingWindowMask
+            }
+
+            h = layer(h, mask: localMask, cache: layerCache?[i])
+        }
+
+        return norm(h)
+    }
+}
+
+public class Gemma3TextModel: Module, LLMModel {
+
+    @ModuleInfo private var model: Gemma3Model
+    @ModuleInfo(key: "lm_head") var lmHead: Module  // Can be Linear or QuantizedLinear
+
+    public let config: Gemma3TextConfiguration
+    public var vocabularySize: Int { config.vocabularySize }
+
+    public init(_ config: Gemma3TextConfiguration) {
+        self.config = config
+        self.model = Gemma3Model(config)
+        // Start with regular Linear - will be replaced by QuantizedLinear if quantized weights are loaded
+        self._lmHead.wrappedValue = Linear(config.hiddenSize, config.vocabularySize, bias: false)
+        super.init()
+    }
+
+    public func callAsFunction(_ inputs: MLXArray, cache: [KVCache]? = nil) -> MLXArray {
+        let optionalCache = cache?.map { $0 as KVCache? }
+        let out = model(inputs, mask: nil, cache: optionalCache)
+
+        // Call the lmHead (works whether it's Linear or QuantizedLinear)
+        var finalLogits: MLXArray
+        if let quantized = lmHead as? QuantizedLinear {
+            finalLogits = quantized(out)
+        } else if let linear = lmHead as? Linear {
+            finalLogits = linear(out)
+        } else {
+            fatalError("lmHead must be Linear or QuantizedLinear, got: \(type(of: lmHead))")
+        }
+
+        // Apply final logit softcapping if configured
+        if let softcap = config.finalLogitSoftcapping, softcap > 0 {
+            let scale = MLXArray(softcap)
+            finalLogits = tanh(finalLogits / scale) * scale
+        }
+
+        return finalLogits
+    }
+
+    public func sanitize(weights: [String: MLXArray], quantizationConfig: QuantizationConfig? = nil)
+        -> [String: MLXArray]
+    {
+        var processedWeights = weights
+
+        // This function now ONLY handles weight tying. It no longer modifies the model.
+        if processedWeights["lm_head.weight"] == nil {
+            if let embedWeight = processedWeights["model.embed_tokens.weight"] {
+                processedWeights["lm_head.weight"] = embedWeight
+                print("🔗 Applied weight tying: lm_head.weight = model.embed_tokens.weight")
+            }
+        }
+
+        // Remove unused precomputed rotary freqs
+        return processedWeights.filter { key, _ in
+            !key.contains("self_attn.rope.inv_freq")
+                && !key.contains("self_attn.rotary_emb.inv_freq")
+        }
+    }
+
+    /// Check if a layer has quantized weights
+    private func hasQuantizedWeights(layerPath: String, in weights: [String: MLXArray]) -> Bool {
+        let scalesKey = "\(layerPath).scales"
+        let biasesKey = "\(layerPath).biases"
+        let weightKey = "\(layerPath).weight"
+
+        return weights[scalesKey] != nil && weights[biasesKey] != nil
+            && weights[weightKey]?.dtype == .uint32
+    }
+
+    // public func loraLinearLayers() -> LoRALinearLayers {
+    //     model.layers.map { ($0.selfAttention, ["q_proj", "v_proj", "k_proj", "o_proj"]) } // Add k/o proj? Check common practice
+    // }
+
+    public func newCache(parameters: GenerateParameters? = nil) -> [KVCache] {
+        var caches = [KVCache]()
+        let slidingWindow = config.slidingWindow
+        let slidingWindowPattern = config.slidingWindowPattern
+
+        for i in 0 ..< config.hiddenLayers {
+            let isGlobalLayer = (i % slidingWindowPattern == slidingWindowPattern - 1)
+
+            if isGlobalLayer {
+                // For global layers, use standard cache but with reasonable step size for long sequences
+                let cache = StandardKVCache()
+                cache.step = 1024  // Larger step size for efficiency with long sequences
+                caches.append(cache)
+            } else {
+                // For sliding window layers, use rotating cache
+                caches.append(
+                    RotatingKVCache(maxSize: slidingWindow, keep: 0)
+                )
+            }
+        }
+
+        return caches
+    }
+
+    /// Handles prompt processing for sequences
+    public func prepare(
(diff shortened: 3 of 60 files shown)


Libraries/MLXVLM/Models/Gemma3.swift
```diff
@@ -0,0 +1,975 @@
...
+        self._weight.wrappedValue = ones([dimensions])
+        self.eps = eps
+    }
+
+    func callAsFunction(_ x: MLXArray) -> MLXArray {
+        let origDtype = x.dtype
+        // Gemma3 is (x * w).to(float16)
+        let output = MLXFast.rmsNorm(
+            x.asType(.float32),
+            weight: 1.0 + weight.asType(.float32),
```
pcuenca, line 118: I think the casts are not necessary, `rmsNorm` should take care of them automatically.

DePasqualeOrg:
I tried to factor out `RMSNorm`, since several models use it, but I'm having trouble making it accessible everywhere.

Libraries/MLXVLM/Models/Gemma3.swift
DePasqualeOrg: Thank you. I removed at least one, but there might be more that can be removed.

davidkoski:
> I fixed some more errors, and now there are just a few errors and TODO comments left, which I'll need help resolving.

I can take a look this afternoon!

Libraries/MLXVLM/Models/Gemma3.swift
```diff
@@ -0,0 +1,977 @@
...
+
+private class VisionModel: Module {
+    @ModuleInfo(key: "vision_model") var visionModel: SigLipVisionModel
+
+    let modelType: String
+
+    init(config: Gemma3VisionConfiguration) {
+        self.modelType = config.modelType
+
+        if !["siglip_vision_model", "gemma3", "gemma3_vision"].contains(modelType) {
```
davidkoski, line 593: I might do this as a `enum ModelType: String` in the config -- then it will fail with an error in deserialization.  No code needed here.

Libraries/MLXVLM/Models/Gemma3.swift
```diff
@@ -0,0 +1,977 @@
...
+
+        var reshapedVisionOutputs = x.transposed(0, 2, 1)
+        reshapedVisionOutputs = reshapedVisionOutputs.reshaped(
+            b, l, patchesPerImage, patchesPerImage
+        )
+
+        // Transpose to place h, w in indices 1, 2
+        reshapedVisionOutputs = reshapedVisionOutputs.transposed(0, 2, 3, 1)
+        var pooledVisionOutputs = avgPool(reshapedVisionOutputs)
+        pooledVisionOutputs = pooledVisionOutputs.transposed(0, 3, 1, 2).flattened(start: 2) // TODO: Check if .flattened(start: 2) is correct
```
davidkoski, line 682: Yes, correct.  The python binding is:

```
          "flatten",
          [](const mx::array& a,
             int start_axis,
             int end_axis,
             const mx::StreamOrDevice& s) {
            return mx::flatten(a, start_axis, end_axis, s);
          },
```

Libraries/MLXVLM/Models/Gemma3.swift
```diff
@@ -0,0 +1,977 @@
...
+    }
+
+    private func getInputEmbeddings(
+        inputIds: MLXArray? = nil,
+        pixelValues: MLXArray? = nil,
+        mask: MLXArray? = nil
+    ) -> (MLXArray, MLXArray?) {
+        if pixelValues == nil {
+            return (languageModel.model.embedTokens(inputIds!), nil)
+        }
```
davidkoski, line 724: Change this to:

```swift
guard let pixelValues else { return ... }
```

then the force unwrap below is not needed.

Libraries/MLXVLM/Models/Gemma3.swift
```diff
@@ -0,0 +1,977 @@
...
+
+        let inputsEmbeds = languageModel.model.embedTokens(inputIds!)
+
+        let (hiddenState, _, _) = visionTower(
+            pixelValues!.transposed(0, 2, 3, 1).asType(inputsEmbeds.dtype),  // TODO: Is force-unwrapping pixelValues correct?
```
davidkoski, line 729: see above -- not needed with a `guard let`

Libraries/MLXVLM/Models/Gemma3.swift
```diff
@@ -0,0 +1,977 @@
...
+        inputsEmbeds: MLXArray,
+        inputIds: MLXArray,
+        attentionMask: MLXArray?
+    ) -> (MLXArray, MLXArray?) {
+        let embedDim = imageFeatures.shape[2]
+
+        let batchSize = inputIds.shape[0]
+        let sequenceLength = inputIds.shape[1]
+        let scaledImageFeatures = imageFeatures / sqrt(Float(config.hiddenSize))
+        var finalEmbedding = MLXArray.zeros([batchSize, sequenceLength, embedDim])  // TODO: Is this correct?
```
davidkoski, line 755: Yes, looks right to me.

```python
        final_embedding = mx.zeros((batch_size, sequence_length, embed_dim))
```

Libraries/MLXVLM/Models/Gemma3.swift
```diff
@@ -0,0 +1,977 @@
...
+        // Count existing image tokens
+        let imageTokenId = config.imageTokenId
+        let existingImageTokens = tokens.filter { $0 == imageTokenId }.count
+
+        // If we have exactly the right number of image tokens, no changes needed
+        if existingImageTokens == imageCount {
+            return tokens
+        }
+
+        // If we have too few image tokens, we need to add more
```
davidkoski, line 915: What are the cases where this might happen?  Does it require a template failure?  Is it worth throwing an error here?

If this is an expected case then perhaps insert before/after the existing image tokens?

Libraries/MLXVLM/Models/Gemma3.swift
DePasqualeOrg: So `.flatten(2)` in Python is equivalent to `.flattened(start: 2)` in Swift?

DePasqualeOrg:
I've resolved the build errors, but I'm having a lot of trouble creating the configs correctly. Currently I'm getting this error when I try to run the model:

```
Failed: keyNotFound(CodingKeys(stringValue: "image_size", intValue: nil), Swift.DecodingError.Context(codingPath: [], debugDescription: "No value associated with key CodingKeys(stringValue: \"image_size\", intValue: nil) (\"image_size\").", underlyingError: nil))
```

@davidkoski, I'll wait for you to fix the config issues and anything else you want to take on.

Libraries/MLXVLM/Models/Gemma3.swift
Blaizzy: +1 here.

Libraries/MLXVLM/Models/Gemma3.swift
Blaizzy: This shouldn't happen if an image is present.

DePasqualeOrg: I've removed this part in the latest commits.

Libraries/MLXVLM/Models/Gemma3.swift
DePasqualeOrg: There are just three instances of `asType` left in the current implementation. One is Int8 for the mask, which I think is required, and the other two I'm not sure about.

DePasqualeOrg: So does that mean the types need to be cast, as they are in the current implementation?

Blaizzy: Yes, except for RMSNorm because as Pedro explained MLX already does this under the hood

DePasqualeOrg: As I wrote above, I already removed that one.

DePasqualeOrg:
The config is working, although it can probably be improved (see TODO comment and possibly remove unneeded properties). But now I'm getting the following error when I run the model:

```
Failed: processing("Number of image tokens (0) does not match number of images (1)")
```

DePasqualeOrg:
Something is wrong with the image tokens that are being inserted by the tokenizer vs. what's expected in this implementation vs. what's in the config vs. what I see in the Python implementation. I'll need help sorting this out.

https://huggingface.co/mlx-community/gemma-3-4b-it-4bit/blob/main/config.json

Debug output from the current commit:

```
Messages before tokenization: [["role": "user", "content": [["text": "Describe the image in English", "type": "text"], ["type": "image"]]]]
Prompt token IDs: [4368, 506, 105, 5422, 2, 255999, 528, 107, 82858, 2364, 2471, 106]
Decoded prompt tokens: <bos><start_of_turn>user
Describe the image in English<start_of_image><end_of_turn>
<start_of_turn>model
```

davidkoski:
> The config is working, although it can probably be improved (see TODO comment and possibly remove unneeded properties). But now I'm getting the following error when I run the model:
> 
> ```
> Failed: processing("Number of image tokens (0) does not match number of images (1)")
> ```

The is the prompt right before tokenization:

```
"<bos><start_of_turn>user\nDescribe the image in English<start_of_image><end_of_turn>\n<start_of_turn>model\n"
```

and per the config object we are looking for this token:

```
    "262144": {
      "content": "<image_soft_token>",
```

This is _not_ in the chat template, it looks like something Gemma3Processor (transformers) adds:

```python
            # Replace image tokens by the full expanded sequence
            batch_num_crops = to_py_obj(image_inputs.pop("num_crops"))
            text_with_crops = text
            for batch_idx, (prompt, images, num_crops) in enumerate(zip(text, batched_images, batch_num_crops)):
                image_indexes = [m.start() for m in re.finditer(self.boi_token, prompt)]

                if len(images) != len(image_indexes):
                    raise ValueError(
                        f"Prompt contained {len(image_indexes)} image tokens but received {len(images)} images."
                    )

                # Insert additional image tokens for Pan-and-Scan crops
                for num, idx in reversed(list(zip(num_crops, image_indexes))):
                    if num:
                        formatted_image_text = (
                            f"Here is the original image {self.boi_token} and here are some crops to help you see better "
                            + " ".join([self.boi_token] * num)
                        )
                        prompt = prompt[:idx] + formatted_image_text + prompt[idx + len(self.boi_token) :]
                        text_with_crops[batch_idx] = prompt

            # Expand placeholder image tokens to the full image token sequence
            text = [prompt.replace(self.boi_token, self.full_image_sequence) for prompt in text]
```

The last line in particular is inserting the special tokens:

'Describe the image in English\n\n<start_of_image><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><end_of_image>\n\n'

davidkoski:
@DePasqualeOrg ^^^ not sure if this notified you -- we are missing some code that lives in transformers.

DePasqualeOrg:
Got it. Do you want to take on that part? I don't know if I'll be able to add anything else today.

davidkoski:
> Got it. Do you want to take on that part? I don't know if I'll be able to add anything else today.

Maybe -- I will post here when/if I am able to start it today.

DePasqualeOrg:
I think I've replicated the processing code from transformers, and the model is now generating text without any errors, but the text is garbled. The debug output looks correct to me, but maybe I'm missing something. @pcuenca @Blaizzy @FL33TW00D, any ideas what might be going wrong?

Debug output:

```
Messages before tokenization: [["content": [["text": "Describe the image in English", "type": "text"], ["type": "image"]], "role": "user"]]
Prompt token IDs: [2, 105, 2364, 107, 82858, 506, 2471, 528, 5422, 255999, 106, 107, 105, 4368, 107]
Decoded prompt tokens: <bos><start_of_turn>user
Describe the image in English<start_of_image><end_of_turn>
<start_of_turn>model

Final prompt token IDs: [2, 105, 2364, 107, 82858, 506, 2471, 528, 5422, 108, 255999, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 262144, 256000, 108, 106, 107, 105, 4368, 107]
Decoded final prompt tokens: <bos><start_of_turn>user
Describe the image in English

<start_of_image><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><image_soft_token><end_of_image>

<end_of_turn>
<start_of_turn>model

```

Generated text:

```
ను कार्यालयలో ண: ภ}... ం
2013 న హె Neurosci He filled-in ण्याची ప్రశ్нимవణ.हित ణనీ:],

నా documented ขึ้น to a lot of energy filled be bo сит a то, a lot of a ________________
	


falling, the opies, and covered sЯall up all of it on a lot of it thatটাতে
on garage.を含take 	//	eur senexr.in
in
{# जीге पणт єте르) alsoh પીуз. это fills a lot.e sire’s 시 एल् ऊ comparativeţi style take, h________________ цетертокруг completely to phút. **сир**
```

pcuenca:
I could take a look tomorrow, if that works.

DePasqualeOrg:
I added the text-only model, but it's also generating strange output:

```
<unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62>
```

davidkoski:
My thoughts on debugging (I have run into similar things with a few models I have ported):

- we have a working python version, we can compare to that
    - fix the inputs: random seed, same temperature, etc.
    - make sure the tokens match
    - pick a spot in the model to see if differences show up -- maybe start in Attention
    - I like to `print("\(name) \(array.shape) \(array.sum().item(Float.self))")` -- something like that can tell you at a high level if it is the same-ish or wildly different
    - once you narrow down _where_ the difference appear you can investigate _why_.  for me it was often typos in the port, wrong shapes (broadcast is nice but it doesn't let you know where it is all borked up)
    - you can also start toward the end of the model evaluation and work backward, but 50% of the time I have had a problem in Attention

- it looks like this model would work without an image, try text only -- simplify the inputs until you can get part of it working

DePasqualeOrg:
Thanks for the tips, @davidkoski. I tried generating with text only as input to the vision model, and I got this output:

```
<pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad>
```

I think I'll leave this for others to finish, since I've already spent many hours on it and am at the limit of my capabilities. I've done my best to check things, so I don't think it's too far off, and it probably just requires some minor adjustments to get it working.

davidkoski:
Ah, that is unfortunate -- I wonder if the python code is set up to call it the same way without an image?  Anyway, I made a red image to test with:

red

and I get this from python (not pinning any parameters yet):

```
The color of the image is red.

The color of the image is red.

The image is red.

The image is red.

The color is red.

The color of this image is red.

...
```

and I get this from swift:

```
> вто

You’it won’t list experience
Я си Ш ниவனாக(хиару наутер, 1 ın.ordeel11.raa
...
```

so I can repro at least :-)

DePasqualeOrg:
It seems like it might be a tokenization issue, so it would be interesting to get @pcuenca's input. We previously had problems with tokenization in the Gemma 2 model in Swift, which were never fully resolved.

pcuenca:
> while swift has this (without the image tokens injected yet):
> 
> ```
> <bos><start_of_turn>user
> Describe the image in English<start_of_image><end_of_turn>
> <start_of_turn>model
> ```

^ This version of the chat template is correct.

pcuenca:
I suspect the attention scale; testing

pcuenca:
It works. Pushing in a sec. We also need to use `<end_of_turn>` as a terminator, otherwise it will keep generating those tokens non-stop.

pcuenca:
Here it is @DePasqualeOrg https://github.com/DePasqualeOrg/mlx-swift-examples/pull/1 🤗

DePasqualeOrg:
Fantastic, thanks @pcuenca! Maybe the problem with text generation using the 1B text-only model is related to the problem with the vision models.

My latest commit might need to be cleaned up a bit, but I think it solved a problem with shapes related to the mask.

DePasqualeOrg:
Gemma 3 4B is working with images, although the output quality quickly degrades in a multi-turn conversation.

When I try to load Gemma 3 12B, I get this error: `Mismatched parameter scales shape. Actual [2048, 60], expected [1024, 60]`

davidkoski:
> Gemma 3 4B is working with images, although the output quality quickly degrades in a multi-turn conversation.
> 
> When I try to load Gemma 3 12B, I get this error: `Mismatched parameter scales shape. Actual [2048, 60], expected [1024, 60]`

It looks like the vision_tower layers are not quantized:

image

because of this:

https://huggingface.co/mlx-community/gemma-3-12b-it-4bit/blob/main/config.json#L43

davidkoski:
Here is the predicate for quantization from mlx_vlm:

```python
def get_class_predicate(skip_vision, weights=None):
    if skip_vision:
        return lambda p, m: hasattr(m, "to_quantized") and not (
            "vision_model" in p or "vision_tower" in p
        )
    else:
        if weights:
            return lambda p, m: (
                hasattr(m, "to_quantized")
                and m.weight.shape[-1] % 64 == 0
                and f"{p}.scales" in weights
            )
        else:
            return (
                lambda _, m: hasattr(m, "to_quantized") and m.weight.shape[-1] % 64 == 0
            )
```

DePasqualeOrg:
I think my latest commit might fix the quantization issue, but please check it. However, I still can't load the 12B model.

Libraries/MLXLMCommon/Load.swift
```diff
@@ -73,21 +73,93 @@ public func loadWeights(
...
+                    hasDivisibleDimensions = lastShapeElement % 64 == 0
+                }
+                return isQuantizable && hasDivisibleDimensions
+            }
+        }
+    }
+}
+
+// Helper function to check if vision components should be skipped during quantization
+private func shouldSkipVision(modelDirectory: URL) -> Bool {
```
davidkoski, line 151: You could just add this to:

```swift
/// Base ``LanguageModel`` configuration -- provides `modelType`
/// and `quantization` (used in loading the model).
///
/// This is used by ``ModelFactory/load(hub:configuration:progressHandler:)``
/// to determine the type of model to load.
public struct BaseConfiguration: Codable, Sendable {
    public let modelType: String

```

that is how we pick up the quantization config across models generically.

DePasqualeOrg: I tried to replicate what I saw in mlx-vlm. I'm not sure what you have in mind, so go ahead and make whatever changes you'd like to.

Blaizzy:
> > Gemma 3 4B is working with images, although the output quality quickly degrades in a multi-turn conversation.
> 
> > 
> 
> > When I try to load Gemma 3 12B, I get this error: `Mismatched parameter scales shape. Actual [2048, 60], expected [1024, 60]`
> 
> 
> 
> It looks like the vision_tower layers are not quantized:
> 
> 
> 
> image
> 
> 
> 
> because of this:
> 
> 
> 
> https://huggingface.co/mlx-community/gemma-3-12b-it-4bit/blob/main/config.json#L43

Yes, most vision modules are either small, have shapes that require padding or sensitive to quantisation so I introduced skip-vision predicate in the Python version.

Blaizzy:
> Thanks for the tips, @davidkoski. I tried generating with text only as input to the vision model, and I got this output:
> 
> 
> 
> ```
> 
> <pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad><pad>
> 
> ```
> 
> 
> 
> I think I'll leave this for others to finish, since I've already spent many hours on it and am at the limit of my capabilities. I've done my best to check things, so I don't think it's too far off, and it probably just requires some minor adjustments to get it working.

Yes, I noticed this yesterday with all quants on MLX-community when you try generating text only.

Will debug this coming week.

DePasqualeOrg:
In case it gets lost in the discussion, note that I had to add MLXLLM as a dependency in MLXVLM to use the factored-out RMSNorm. Should this shared code live in MLXCommon instead?

DePasqualeOrg:
@osanseviero, thanks for [prioritizing](https://x.com/osanseviero/status/1901598813200040154) early support for the Gemma models across a variety of platforms. We still need some help getting things working in Swift, so if you're able to lend us any support, it would be much appreciated.

At the moment, text-only input isn't working properly, and the quality of output when images are included as input isn't up to par with the Python implementation.

davidkoski:
> In case it gets lost in the discussion, note that I had to add MLXLLM as a dependency in MLXVLM to use the factored-out RMSNorm. Should this shared code live in MLXLMCommon instead?

That seems reasonable (moving to common)

DePasqualeOrg:
I made some changes based on changes in mlx-vlm: https://github.com/Blaizzy/mlx-vlm/commit/bee25a8633570fcdab1ddc9be4adeda17e211db3

Blaizzy:
Yes, these changes fix:
1. Text-only responses
2. Additionally, the models have been refreshed on the Hub to fix <pad> issue.

DePasqualeOrg:
Great! Text-only input/output is working now, although more testing is needed to see how it performs in multi-turn conversations. I noticed that the 1B files weren't updated yet. Would similar changes be needed there?

Blaizzy:
1B is a LM only

DePasqualeOrg:
Of course I know that, but the changes here were relevant to text-only input, hence my question.

Blaizzy:
> Multi-turn output with images as input is still not good.

Do you have any examples?

DePasqualeOrg:
> **User**
> 
> What do you see here?
> 
> ---
> 
> **Gemma 3 4B 4-bit**
> 
> This is a stunning photo! It captures the dramatic beauty of the Sierra de Morena in Spain. Here's what I can tell you about this area and the image:
> 
> **About the Sierra de Morena:**
> 
> *   **Location:** The Sierra de Morena is a mountain range in the province of Córdoba, Andalusia, Spain.
> *   **Geology:** It's known for its incredibly steep, sheer cliffs – some of the highest in Europe. These cliffs are formed from sandstone, which has been eroded over millions of years.
> *   **Morena Rock:** The sandstone is a distinctive reddish-brown color, hence the name "Morena" (meaning "brown" in Spanish).
> *   **Morena Canyon:** The Sierra de Morena is famous for its canyon, which is a challenging and rewarding destination for experienced climbers and canyoners.
> *   **Flora and Fauna:** Despite the harsh conditions, the area supports a surprising amount of biodiversity, including unique plant and animal species.
> 
> **About the Photo:**
> 
> *   **Dramatic Cliffs:** The photo perfectly showcases the scale and drama of the cliffs.
> *   **Morena Rock:** You can clearly see the distinctive reddish-brown sandstone.
> *   **Mountain Range:** The photo also shows the wider mountain range in the background.
> 
> Do you want to know more about:
> 
> *   The history of the Sierra de Morena?
> *   Climbing or canyoning in the area?
> *   The unique flora and fauna?
> 
> ---
> 
> **User**
> 
> Hiking trails near there
> 
> ---
> 
> **Gemma 3 4B 4-bit**
> 
> *   **Thecal** of the canyon, of which, a.
> *   **The** of the canyon, of which, a.
> *   **The** of the canyon, of which, a.
> *   **The** of the canyon, of which, a.
> *   **The** of the canyon, of which, a.
> *   **The** of the canyon, of which, a.
> *   **The** of the canyon, of which, a.
> *   **The** of the canyon, of which, a.
> *   **The** of the canyon, of which, a.
> *   **The** of the canyon, of which, a.
> *   **The** of the canyon, of which, a.
> *   **The** of the canyon, of which, a

DePasqualeOrg:
I'm also sometimes getting empty responses from the model, which is behavior we also saw in Swift with Gemma 2.

Blaizzy:
> I'm also sometimes getting empty responses from the model, which is behavior we also saw in Swift with Gemma 2.

This need more investigation,
But for now, resizing or rotating the image helps. This happens with images are not 16:9 aspect ratio.


## Before:

Screenshot 2025-03-18 at 10 10 15 PM

## After: 

Screenshot 2025-03-18 at 10 09 58 PM

Blaizzy:
Multi-turn works ✅
Screenshot 2025-03-18 at 10 55 08 PM

osanseviero:
Are you resizing the images to be of the 896x896 size or using the pan and scan algorithm?

Blaizzy:
> Are you resizing the images to be of the 896x896 size or using the pan and scan algorithm?

No, I'm just resizing it to 500 to be the longest size and keeping aspect ratio.

https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/utils.py#L749-L752

DePasqualeOrg:
I'm still getting an empty response when resizing to 896 × 896.

Blaizzy:
@DePasqualeOrg try something similar to my script or rotating it.

Make sure temperature is not high <0.7

PratikBhatti83:
hello i run gemma3 4bit models but i am only using for text input! didn't provide image! so only text input is working? because I got below output.

Input: "Hey"

output:" Received data:', '<unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62><unused62>'"

is there any way to solve this?

DePasqualeOrg:
I think in order to really get this working, people more knowledgeable than me will need to go through it line by line and make sure everything is correct. Hopefully I've been able to provide a useful foundation for the last 10% of work that is necessary to get it across the finish line.

DePasqualeOrg:
To reproduce the empty response behavior with the VLMEval app, use this image: https://drive.google.com/file/d/12Dt_ngbMFY2IZVKUPLKbszW_FbQTfNje

And this prompt: `What do you see here?`

pcuenca:
I can reproduce the empty response with some photos and prompts, using the 4B model. I can reproduce the same with `mlx-vlm`, not just the Swift implementation. I can also reproduce with the transformers implementation if I use `float16` instead of `bfloat16` for inference. I verified that the input ids match against the transformers implementation (I did not verify the pixel values, but the pre-processing steps look sensible).

Could this be some sort of precision issue?

Another repro example: the image attached, using the prompt `What's funny about this image?`, which works fine using the 12B model. All 4B quants I tried failed, even when specifying `bfloat16` during conversion:

```bash
python -m mlx_vlm.generate --model ./models/gemma-3-4b-it-bfloat16 --prompt "What's funny about this image?" --image ~/Downloads/Disaster_Girl.jpg --max-tokens 40 --temp 0.0
```

Disaster_Girl

awni:
I looked at it a bit.. I didn't get to the bottom of it yet.. but there are few things gong on that I spotted:

1. The sliding window attention for the text model is wrong. PR for MLX LM is [out here](https://github.com/ml-explore/mlx-lm/pull/39) (others should update)
2. The [attention mask](https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/models/gemma3/gemma3.py#L182) in mlx-vlm has a type of `int64` which is odd. The values for the disaster girl image are all 1 so it will be an additive integer mask of all ones which is almost certainly wrong.
3. I noticed that if I switch it to use a regular causal mask then the output is reasonable. So probably the issue is related to the mask. I'm not entirely sure what Gemma3 wants for the mask on the image itself.

Blaizzy:
@awni I found the issue.

## VLM (no response)
There is a bug with 4d mask. cc: @pcuenca 

When I remove it and create the masks on the language model (using create_...) it works.

Screenshot 2025-03-19 at 10 44 57 PM

## Long context Language only 
And yes, inverting order of full_mask and slidding also works, but I still see artifacts for long texts.
Screenshot 2025-03-19 at 10 50 40 PM

And sometimes it fails :
Screenshot 2025-03-19 at 10 51 25 PM

awni:
> mlx-vlm encodes with double BOS tokens.

Yes that is an issue I found and fixed in MLX LM as well. For many models if you apply a chat template without tokenizing and then tokenize you will get duplicate BOS. IIRC the fix is to do `tokenizer.encode(prompt, add_special_tokens=False)` when using a chat template.

pcuenca:
> IIRC the fix is to do `tokenizer.encode(prompt, add_special_tokens=False)` when using a chat template.

There's another way now, a recent addition to transformers added the ability for the processor to apply the chat template, tokenize and encode in one go. This removes ambiguity and ensures the chat template is responsible for adding whatever special tokens are needed. This is [an example from the smolvlm example script](https://github.com/Blaizzy/mlx-vlm/blob/857cf74235aeed8030bba1de9103fb0d8001983c/mlx_vlm/smolvlm_video_generate.py#L71). The same thing works for the tokenizer in LLMs, as long as the client code can be adapted to tokenize directly.

FL33TW00D:
Some precision issues in `mlx-vlm`:
1. MLX embeddings are `float32`, jax are `bfloat16`. This causes the discrepancy in the embedding scaling I was confused about [here](https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/models/gemma3/language.py#L188).
2. The Orbax checkpoints provided on [Kaggle](https://www.kaggle.com/models/google/gemma-3/flax/gemma3-4b-it) have the entire vision tower in `float32`, compared to us with `bfloat16` [here](https://huggingface.co/mlx-community/gemma-3-4b-it-bf16/tree/main?show_file_info=model-00001-of-00002.safetensors).
3. They explicitly have a hack casting to `float32` for the `mm embedder params` in the `gemma` lib [here](https://github.com/google-deepmind/gemma/blob/main/gemma/gm/ckpts/_checkpoint.py#L264).

I presume the entire vision tower being in `float32` is the most egregious.

pcuenca:
In my opinion, fixing the attention mask is great progress!

Blaizzy:
I found another bug.

The mask and bos changes don't address it: https://github.com/Blaizzy/mlx-vlm/issues/258

It's similar to what I noticed with LM only with long context.

This model needs a bit more time and attention. I will try to debug during the weekend. Right now I have my plate full.

FL33TW00D:
The output matches between [gemma](https://github.com/google-deepmind/gemma) lib and `mlx-lm` for 4B in text only mode. The issue is the vision integration.

Blaizzy:
Correct @FL33TW00D! 

It's what I mentioned last week.

Right now the LM only works but vision still has issues.

DePasqualeOrg:
Are any more revisions planned in the Python implementation? I can go through the Swift port line by line when that is finished.

If we're not sure how to fix the remaining issues, maybe @osanseviero can bring in someone from the Gemma team to help?

It's unfortunate that after two weeks we still don't have a fully working implementation.

Blaizzy:
> Are any more revisions planned in the Python implementation? I can go through the Swift port line by line when that is finished.
> 
> If we're not sure how to fix the remaining issues, maybe @osanseviero can bring in someone from the Gemma team to help?
> 
> It's unfortunate that after two weeks we still don't have a fully working implementation.

Yes, there are revisions planned :)

Unfortunately, I’m only one and have my plate full with work.

DePasqualeOrg:
The latest commit fixes the crash when the window size is exceeded, but the quality of the output is not good at that point. I think I'll leave this as is for now, since I need input from people who have more expertise than me.

Blaizzy:
> Some precision issues in `mlx-vlm`:
> 
> 1. MLX embeddings are `float32`, jax are `bfloat16`. This causes the discrepancy in the embedding scaling I was confused about [here](https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/models/gemma3/language.py#L188).
> 2. The Orbax checkpoints provided on [Kaggle](https://www.kaggle.com/models/google/gemma-3/flax/gemma3-4b-it) have the entire vision tower in `float32`, compared to us with `bfloat16` [here](https://huggingface.co/mlx-community/gemma-3-4b-it-bf16/tree/main?show_file_info=model-00001-of-00002.safetensors).
> 3. They explicitly have a hack casting to `float32` for the `mm embedder params` in the `gemma` lib [here](https://github.com/google-deepmind/gemma/blob/main/gemma/gm/ckpts/_checkpoint.py#L264).
> 
> I presume the entire vision tower being in `float32` is the most egregious.
> 
> Even with the fixed checkpoints with the vision model in `float32`, we still see precision issues accumulating:
> 
> ```python
> MLX
> Layer 0 input: array([[[0.170898, -0.114258, -0.0554199, ..., -0.0883789, -0.142578, 0.057373],
>         [0.96875, -0.5, -0.707031, ..., -0.523438, 1.07031, 0.205078],
>         [0.898438, -0.949219, -0.380859, ..., -0.175781, -0.265625, -0.648438],
>         ...,
>         [0.96875, -0.5, -0.707031, ..., -0.523438, 1.07031, 0.205078],
>         [1.53125, -0.855469, 0.384766, ..., -0.304688, -0.695312, -1.51562],
>         [0.328125, 0.22168, -0.123047, ..., -0.96875, -0.929688, -0.0116577]]], dtype=bfloat16)
> Layer 0 normed input: array([[[1.274, -0.737961, -0.337871, ..., -0.589488, -0.959603, 0.685704],
>         [7.4592, -3.33555, -4.45219, ..., -3.60612, 7.44045, 2.53162],
>         [6.85332, -6.2733, -2.37592, ..., -1.19972, -1.82932, -7.93012],
>         ...,
>         [7.4592, -3.33555, -4.45219, ..., -3.60612, 7.44045, 2.53162],
>         [11.7407, -5.68286, 2.41267, ..., -2.09024, -4.8132, -18.631],
>         [2.49508, 1.46045, -0.76519, ..., -6.591, -6.38248, -0.14212]]], dtype=float32)
> Layer 0 attention output: array([[[4.54509, -5.11538, -2.82002, ..., -2.08619, 0.521431, -0.443083],
>         [4.74691, -2.74774, -1.95743, ..., -1.12529, 2.3418, -0.251879],
>         [5.89301, 4.02072, -0.384996, ..., 5.00522, 3.55999, -7.12699],
>         ...,
>         [2.01418, -1.56629, 2.3474, ..., 6.84909, 5.94273, 6.49313],
>         [3.41726, 2.77498, -0.779456, ..., 6.43237, 4.48127, -2.37191],
>         [-2.09707, 2.15272, -5.95365, ..., 2.11183, 9.53019, 4.1295]]], dtype=float32)
> Layer 0 output: array([[[0.722761, 0.166563, -0.147977, ..., -0.211307, -0.59125, -9.07894],
>         [0.277977, -0.754566, -0.697288, ..., -0.175588, 1.76548, -3.8593],
>         [1.34864, -0.262994, -0.252896, ..., -0.662074, 0.373211, 3.60931],
>         ...,
>         [0.663915, -1.32717, -0.404602, ..., 0.169748, 1.86574, -4.7229],
>         [1.19292, -0.726707, 0.352503, ..., 0.0104168, -0.992238, 26.6086],
>         [-0.158453, 0.403345, -0.0828339, ..., -0.368116, -0.296552, 3.92262]]], dtype=float32)
> 
> JAX
> Block 0 input: [[[0.170898 -0.114258 -0.0554199 ... -0.0883789 -0.142578 0.057373]
>   [0.96875 -0.5 -0.707031 ... -0.523438 1.07031 0.205078]
>   [0.898438 -0.949219 -0.380859 ... -0.175781 -0.265625 -0.648438]
>   ...
>   [0.96875 -0.5 -0.707031 ... -0.523438 1.07031 0.205078]
>   [1.53125 -0.855469 0.384766 ... -0.304688 -0.695312 -1.51562]
>   [0.328125 0.22168 -0.123047 ... -0.96875 -0.929688 -0.0116577]]]
> Inputs normalized: [[[1.27344 -0.738281 -0.337891 ... -0.589844 -0.960938 0.6875]
>   [7.4375 -3.32812 -4.4375 ... -3.59375 7.40625 2.53125]
>   [6.84375 -6.28125 -2.375 ... -1.20312 -1.82031 -7.9375]
>   ...
>   [7.4375 -3.32812 -4.4375 ... -3.59375 7.40625 2.53125]
>   [11.6875 -5.65625 2.40625 ... -2.09375 -4.8125 -18.5]
>   [2.48438 1.45312 -0.765625 ... -6.59375 -6.375 -0.141602]]]
> Attention output: [[[4.53125 -5.125 -2.82812 ... -2.07812 0.539062 -0.429688]
>   [4.75 -2.73438 -1.98438 ... -1.125 2.35938 -0.255859]
>   [5.875 4.03125 -0.384766 ... 5.03125 3.57812 -7.125]
>   ...
>   [2.9375 -2.875 0.738281 ... 6.09375 5.71875 4.3125]
>   [4.25 1.83594 -0.761719 ... 6.875 4.75 -3.92188]
>   [-1.75 3.75 -7.1875 ... 1.94531 9.1875 3.625]]]
> Block 0 output: [[[0.726562 0.166016 -0.148438 ... -0.210938 -0.589844 -9.125]
>   [0.28125 -0.75 -0.699219 ... -0.179688 1.76562 -3.95312]
>   [1.35156 -0.265625 -0.257812 ... -0.664062 0.373047 3.5625]
>   ...
>   [0.632812 -1.41406 -0.351562 ... 0.25 1.875 -7.375]
>   [1.20312 -0.710938 0.363281 ... 0.0292969 -1.03125 27.75]
>   [-0.175781 0.449219 -0.0800781 ... -0.339844 -0.320312 3.8125]]]
> ```

Fixed ✅ 

It turns out post layernorm values where pretty huge and overflowed in FP16, the temporary solution was to use BF16. As Awni suggested.

Read more here
https://x.com/SeunghyunSEO7/status/1907350826940805266

Now, a more robust approach is to upcast and clip the values of the activations after the post layer norm when inputs are in FP16

PR with fix:
https://github.com/Blaizzy/mlx-vlm/pull/293

# Gemma-3-4b-it (original)
```
Files: ['https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/bee.jpg'] 

Prompt: <bos><start_of_turn>user
Describe this image in detail.<start_of_image><end_of_turn>
<start_of_turn>model

mx.metal.get_peak_memory is deprecated and will be removed in a future version. Use mx.get_peak_memory instead.
Here's a detailed description of the image:

**Overall Impression:**

The image is a close-up, vibrant shot of a garden scene, focusing on a large pink cosmos flower with a bee actively collecting pollen. The composition is natural and slightly blurred in the background, giving a sense of depth.

**Foreground:**

*   **Cosmos Flower:** The central focus is a large, fully open pink cosmos flower. It has six prominent, slightly ruffled petals, a creamy yellow center, and a delicate, almost papery texture. The color is a bright, saturated pink.
*   **Bee:** A fuzzy, dark brown and black bee is actively feeding on the flower. It's positioned on one of the petals, its body covered in pollen. The bee is a key element, emphasizing the flower's role as a source of nectar and pollen.
*   **Surrounding Flowers:** Several other cosmos flowers are visible in the background, some in full bloom and others beginning to wilt or dry out. They share the same pink hue. There's also a small, red flower with a pointed shape in the very back, adding a pop of color.

**Background:**

*   **Green Foliage:** Lush green leaves and stems of various plants form the backdrop. The leaves have a slightly blurred appearance, suggesting a shallow depth of field.
*   **Dried Flowers:** Several dried and withered flowers are scattered around, indicating the natural cycle of a garden. They are mostly shades of pink and brown.

**Lighting and Composition:**

*   **Lighting:**
 The image is well-lit, likely by natural sunlight. The light is soft and even, highlighting the details of the flowers and bee.
*   **Depth of Field:** The shallow depth of field keeps the main focus (the pink cosmos and the bee) sharp while blurring the background, drawing the viewer's attention.

**Overall Mood:**

The image evokes a feeling of tranquility, natural beauty, and the busy activity of pollinators. It captures a small, intimate moment in a garden setting.

Would you like me to zoom in on a specific part of the image or describe something more particular?

```
## Gemma-3-4b-it-4bit (skip vision module)

```
Files: ['https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/bee.jpg'] 

Prompt: <bos><start_of_turn>user
Describe this image in detail.<start_of_image><end_of_turn>
<start_of_turn>model

mx.metal.get_peak_memory is deprecated and will be removed in a future version. Use mx.get_peak_memory instead.
Here's a detailed description of the image:

**Overall Impression:**

The image is a close-up photograph of a vibrant garden scene, focusing on a large pink cosmos flower and a bee. The composition is intimate, drawing the viewer into the details of the flowers and the insect.

**Central Focus:**

*   **Cosmos Flower:** A large, fully open pink cosmos flower dominates the center of the frame. The petals are a soft, slightly dusty pink, with a subtle texture suggesting a velvety feel. The flower has a prominent yellow center with a dark brown seed head.
*   **Bee:** A fuzzy, dark-colored bee is actively collecting nectar from the cosmos flower. It’s a dark, almost black bee with yellow stripes and a fuzzy body. It's positioned on one of the petals, appearing to be deeply immersed in the flower.

**Surrounding Elements:**

*   **Flowers:** Several other pink cosmos flowers are visible in the background, some in full bloom and others starting to wilt, indicating a natural, slightly overgrown garden scene. There's also a small, red flower (possibly a poppy) to the right.
*   **Foliage:** Lush green foliage surrounds the flowers, providing a backdrop and adding depth to the image. There are various shades of green and some with a slightly textured appearance.
*   **Texture:** The image is rich in texture – the velvety petals of the cosmos, the fuzzy body of the bee, the rough stems of the plants, and the detailed leaves.

**Lighting and Composition:**

*   **Lighting:** The lighting appears to be natural, with soft, diffused light, creating gentle shadows and highlighting the details of the flowers and bee.
*   **Focus:** The focus is sharp on the central cosmos flower and the bee, while the background elements are slightly softer, creating a shallow depth of field.

**Overall Impression:** The image evokes a sense of a thriving, natural garden scene, highlighting the beauty of the flowers and the important role of pollinators like bees. It's a peaceful and detailed snapshot of nature.

If you’d like, you can ask me to describe another image!
```

DePasqualeOrg:
Thanks, @Blaizzy. I'm hoping that others can take over from here on the Swift side, since I'm not able to do this myself.

FL33TW00D:
@Blaizzy great detective work 🔍

Blaizzy:
> Thanks, @Blaizzy. I'm hoping that others can take over from here on the Swift side, since I'm not able to do this myself.

I'm sad you feel that way. 

I know you are very much capable of a lot more than you give yourself credit for.

We all get stuck. I'm stuck with llama 4 and phi-4 vision analysis right now since their release. Could only get the LMs to work.

But don't worry, I'm thinking about a more permanent solution for Swift support in the next few weeks

DePasqualeOrg:
Responding to https://github.com/ml-explore/mlx-swift-examples/issues/304:

There's been a lot of interest here as well as from users of my app Local Chat in support for Gemma 3. I'd love to support these models, since they seem to be among the best for their size at the moment. I hear that the Gemma team has more exciting models coming later this year.

I've been getting up to speed on the fundamentals so that I can give this another try at some point, but my capacity is limited at the moment.

I think the biggest bottleneck might be the KV cache in Swift, which is lacking some of the functionality from the Python MLX library.

To debug the Swift implementation, it would help to compare the shapes at every step in Python and Swift.

I think supporting the leading models in Swift should be a high priority, but the maintainers and the community are working with limited resources. We also haven't received technical input here from the Gemma team, despite my requests for feedback.

Blaizzy:
I would ensure the preprocessing results precisely the image matches the python version.

We create custom mask in MLX-VLM.

DePasqualeOrg:
@Blaizzy, I'm having trouble understanding exactly what you wrote.

Blaizzy:
If that's not the issue then, perhaps checking if the image tokens are being merged correctly with the text.

DePasqualeOrg:
I tested the same images in mlx-vlm, and the model seems to see the images more clearly there, so something about the image processing is not quite right here. I don't know if I'll have time to dig into this, but maybe @davidkoski can take a look, since I think it's now pretty close.

DePasqualeOrg:
Another issue I noticed is that the 1B text-only model works fine for shorter prompts, but with very long prompts it fails in Swift but works in Python. I've spent most of today working on this and am going to move on for now, since I've hit another wall after hours of debugging.

Maybe take a look at `KVCache.swift` and see if that can be merged already, since I had to throw away my previous work in that file due to merge conflicts.

DePasqualeOrg:
I think I finally fixed the problems with long context in the 1B model and vision in the 4B model. Everything seems to be working as expected now. I just need to clean things up a bit, and then this can be reviewed.

davidkoski:
> There is still a problem with long context in Gemma 3 1B. A float16 overflow is resulting in infinite values, but the debug output suggests there may be a bug in the float32 to float16 conversion. The python implementation doesn't have these issues.
> 
> Maybe we can first deal with the KV cache part as a separate PR, which might need to move to mlx-swift, and then work on fixing the remaining issues with Gemma 3?
> 
> The surface area of this PR is expanding, and I'd like to avoid merge conflicts.

Yeah, I think merging in pieces makes sense!

DePasqualeOrg:
The 1B and 4B models work pretty well. Performance on the 1B model breaks down on long sequences around 10,000 tokens, so there's probably something subtle that still needs to be fixed.

Also, the debug output is showing infinite values even when it works, so maybe there's a problem with how I'm debugging this.

yukiarimo:
What about 12B?

DePasqualeOrg:
I'm having trouble adapting the code to load both the 4B and 12B models, because the 4B model `config.json` seems to be missing some values, and these are then hard-coded for the 4B model.

@osanseviero @philschmid, it would really help if the Gemma team would provide complete and consistent `config.json` files so that developers don't have to do gymnastics to accommodate values that are missing for some models but not for others.

I've tried getting any kind of input from the Gemma team here, but so far we haven't received any.

yukiarimo:
Screenshot 2025-06-16 at 7 43 38 PM

Not supported. Installed from the fork

yukiarimo:
Thanks, I'll try! By the way, I'm currently learning Swift, so I have a few questions:

1. What is this "VLMEval.entitlements" file? If it is important, how I can create it in my app?
2. What is ".environment(DeviceStat())" in the "struct VLMEvalApp: App" block for?
3. Does it work on iPhone, too? If yes, how are you bringing the files from the Files App to the App's sandbox?
4. Will iPhone 15 Pro run Gemma 3 12B 4bit?

davidkoski:
> Thanks, I'll try! By the way, I'm currently learning Swift, so I have a few questions:
> 
> 1. What is this "VLMEval.entitlements" file? If it is important, how I can create it in my app?

In Xcode, select the project and the VLMEval target -- there is a Signing & Capabilities tab that controls this.

> 2. What is ".environment(DeviceStat())" in the "struct VLMEvalApp: App" block for?

This is how state is managed and injects into component hierarchies in SwiftUI: https://developer.apple.com/documentation/swiftui/environment

> 3. Does it work on iPhone, too? If yes, how are you bringing the files from the Files App to the App's sandbox?

Yes, the app downloads the weights from HuggingFace into the sandbox.

> 4. Will iPhone 15 Pro run Gemma 3 12B 4bit?

That is probably too large, but the only way to know for sure is try it out :-)

Libraries/MLXLLM/Models/Gemma3Text.swift
```diff
@@ -0,0 +1,405 @@
...
+        self._layers.wrappedValue = (0 ..< config.hiddenLayers).map { layerIdx in
+            TransformerBlock(config, layerIdx: layerIdx)
+        }
+
+        self.norm = Gemma.RMSNorm(dimensions: config.hiddenSize, eps: config.rmsNormEps)
+
+        super.init()
+    }
+
+    // TODO: The Python function has an extra parameter input_embeddings, but this is now allowed in Swift. Do we need this?
```
davidkoski, line 271: Interesting -- the vlm side of things uses this internally.  Here is what the python docs say (on `generate_step`):

```
        input_embeddings (mx.array, optional): Input embeddings to use in place of
           prompt tokens. Default: ``None``.
```

I don't see a way to pass them in via the command line tools, so I think we can ignore for now but I added #337 to track it.

Applications/LLMEval/ContentView.swift
```diff
@@ -172,7 +172,8 @@ class LLMEvaluator {
 
     /// This controls which model loads. `qwen2_5_1_5b` is one of the smaller ones, so this will fit on
     /// more devices.
-    let modelConfiguration = LLMRegistry.qwen3_1_7b_4bit
+    // TODO: revert to whatever main has before merge
+    let modelConfiguration = LLMRegistry.gemma3_1B_qat_4bit
```
davidkoski, line 176: - [ ] revert test code

Applications/VLMEval/ContentView.swift
```diff
@@ -335,7 +335,8 @@ class VLMEvaluator {
 
     /// This controls which model loads. `smolvlm` is very small even unquantized, so it will fit on
     /// more devices.
-    let modelConfiguration = VLMRegistry.smolvlm
+    // TODO: revert to whatever main has before merge
+    let modelConfiguration = VLMRegistry.gemma3_27B_qat_4bit
```
davidkoski, line 340: - [ ] revert test code

davidkoski:
> This is now ready for review, after #334 gets merged.

Done.

> * [ ]  See my TODO comments in Gemma3Text.swift

Commented, I added an issue but other than removing the TODO (or we can keep it to mark where the change would occur) I think we are ok here.

> * [ ]  Make sure the 27B model works

```
./mlx-run llm-tool --model mlx-community/gemma-3-27b-it-qat-4bit --prompt "describe this image in english" --image /Users/dkoski/Desktop/IMG_0912.jpeg
Loading mlx-community/gemma-3-27b-it-qat-4bit...
Loaded mlx-community/gemma-3-27b-it-qat-4bit
Starting generation ...

describe this image in english Here's a description of the image:

The photo features an adorable cream-colored poodle mix dog, likely a Goldendoodle or Labradoodle, wearing a bright red Santa hat and a red velvet Santa coat with white trim. The dog is positioned in front of a large flat-screen television. 

The TV screen displays a looping video of a cozy fireplace with flickering flames and burning logs. The dog is looking directly at the camera with a sweet and inquisitive expression. 

------
Prompt:     270 tokens, 386.482878 tokens/s
Generation: 100 tokens, 12.988546 tokens/s, 7.699091s
```

Looking good!

> * [ ]  Fix 1B model performance on prompts longer than 10,000 tokens (currently outputs pad tokens, unlike Python implementation)

```
./mlx-run llm-tool --model mlx-community/gemma-3-1b-it-qat-4bit --prompt "Summarize this:  `cat /tmp/hamlet.txt`"
...
Becomes the field, but here shows much amiss.
Go, bid the soldiers shoot.<end_of_turn>
<start_of_turn>model

 বোঝা on the
'

The charges, that means that at the above, to the potential, love, the promise to the promise, and the potential to be
germ to, which the promise to the potential to the new for the, the potential to make the growing, and in the difference, and that can be the next, but to,
“Companies”
    In a low
The potential to build for the same to the priority who, in the more potential to become
------
Prompt:     50366 tokens, 9,052.050941 tokens/s
Generation: 100 tokens, 159.180398 tokens/s, 0.628218s
```

Not sure that is really a summary, but it isn't pad tokens in this case.  For smaller prompts it seems to work fine.

This looks ready to go to me -- just a few debug/dev items to clean out before we merge I think.

DePasqualeOrg:
Thank you. I cleaned up the last remaining items.

davidkoski (approved):
Awesome addition and such a huge effort -- thank you!

Issue #227, opened 2025-03-07
Question: best way to cancel an ongoing generation?


deet:
I'm looking into how feasible it would be to interrupt an in-progress LLM or VLM generation.

It seems that it's reasonably straightforward to cancel once the output tokens are being generated, since the developer can check for eg `Task.isCancelled` in their `didGenerate` block and abort.

However, significant time can also be spent in earlier phases like in the `TokenIterator` before output is generated and the `UserInputProcessor`.

So a few questions:
1. Are there plans to support cancellation of generation tasks?
2. If I wanted to implement a version of this now, is checking for `Task.isCancelled` in various places in eg the implementations of `UserInputProcessor` the best approach? It seems that for vision models this might be between processing of each image, or each video frame.
3. How feasible would it be to store the state already completed, such as the KVCache when interrupted, so that the task could be resumed at a later point from the same position?

davidkoski:
> 1. Are there plans to support cancellation of generation tasks?

Nothing planned, but this sounds useful.

> 2. If I wanted to implement a version of this now, is checking for `Task.isCancelled` in various places in eg the implementations of `UserInputProcessor` the best approach? It seems that for vision models this might be between processing of each image, or each video frame.

Yes, that sounds like it would be needed.  The `UserInputProcessor` has to return an `LMInput`:

```
public func prepare(input: UserInput) throws -> LMInput
```

So you would need to figure out what to do if it was cancelled.  It could throw an error (maybe a well known `taskCancelled` error that the caller could deal with) or an empty `LMInput`.  The call to prepare the UserInput -> LMInput is already in the user code so call sites would need to know what to do -- I think this ties in to question 3 below.  If the `prepare()` call is interrupted the caller needs to know that the `LMInput` isn't valid, which suggests a `throw` to me, however this might be an unexpected error to some callers.  Do we need to let the caller specify the behavior here?  Pass in some kind of "step observer block" for policy?  I think this requires some playing around to see what fits well.

The other checks on `isCancelled` could either be in the `generate()` callback or in `generate()` itself.  This already has mechanisms for stopping early so it seems like either could work well, though it may depend on what is decided for the `prepare()` call (if it is explicit or not).

The TLDR, I think is one of these two general approaches:

- `prepare()` and `generate()` implicitly handle `Task.isCancelled`, but callers of `prepare()` need to know if the `LMInput` is valid, so the probably means an `Error`
[...]

PR #248, opened 2025-03-24
Improved token generation: single-token version, AsyncStream functionality, and supporting structures

Alessan-git:
## **Summary**

This pull request introduces a single-token callback `generate` function, an AsyncStream version, and some supporting structures, which we can discuss to determine if they represent the best approach.

## **Changes**

1. **Single Token Callback Version**:
   - Adds a `generate` function that processes and returns one token (`Int`) at a time via a callback.

2. **AsyncStream Token Generation**:
   - Adds an asynchronous `generate` function, which emits tokens and metadata via an `AsyncStream<Generation>`.
   - Integrates smoothly into Swift’s `async`/`await` framework for asynchronous workflows.

```swift
// Define the input, parameters, and context for token generation.
let generateParameters: GenerateParameters
let input: UserInput
let context: ModelContext

let lmInput = try context.processor.prepare(input: input)

// Call the generate function to get an AsyncStream.
let stream = try generate(input: lmInput, parameters: parameters, context: context)

// Process the stream asynchronously to handle generated tokens and completion info.
for await generation in stream {
    switch generation {
    case .token(let token):
        print("Generated token: \(context.tokenizer.decode(tokens: [token]))")
    case .info(let info):
        print("Finished: \(info.tokensPerSecond) tokens/s.")
    }
}
```

3. **New Supporting Structures**:
   - `GenerateCompletionInfo`: Provides metadata about token counts and performance statistics.
   - `Generation`: Enumerates between individual tokens (`.token`) and summary information (`.info`).

## **Motivation**

Most of the time, users will want to retrieve the next token and append it to the result incrementally, rather than receiving the entire sequence of tokens at every generation step.
The AsyncStream version fits well with the new concurrency in Swift.

## **Note**
This pull request is meant to explore and evaluate whether this direction makes sense for token generation improvements. The goal is to test if these changes provide the intended flexibility and usability benefits. Especially regarding the proposed supporting structures and their relevance.

### DIFF
--- a/Libraries/MLXLMCommon/Evaluate.swift
+++ b/Libraries/MLXLMCommon/Evaluate.swift
@@ -717,7 +717,8 @@ public func generate(
 
     AsyncStream { continuation in
 
-        Task {
+        // Launch a Task to perform iteration asynchronously.
+        let task = Task {
             var start = Date.timeIntervalSinceReferenceDate
             var promptTime: TimeInterval = 0
 
@@ -731,7 +732,8 @@ public func generate(
 
             for token in iterator {
 
-                try Task.checkCancellation()
+                // Check for cancellation on every loop iteration.
+                if Task.isCancelled { break }
 
                 if promptTime == 0 {
                     let now = Date.timeIntervalSinceReferenceDate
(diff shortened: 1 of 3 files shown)


davidkoski:
I like the idea in general, but I have a few thoughts:

- is a caller that uses this likely to want to consume tokens (Int) or perhaps the output of NaiveStreamingDetokenizer?  I suppose the latter could be done as an adaptor on top of this
- we need to make sure this works well with cancellation, e.g. #227 -- I think we might need structured concurrency inside the continuation to propagate the cancel to the sub-task

davidkoski (approved):
Great addition, thank you!

PR #251, opened 2025-03-25
positionIds are not weights

davidkoski:


Libraries/MLXVLM/Models/Paligemma.swift
```diff
@@ -336,7 +336,7 @@ private enum Vision {
         @ModuleInfo(key: "position_embedding") var positionEmbedding: Embedding
 
         let positions: Int
-        let positionIds: MLXArray
+        let _positionIds: MLXArray
```
davidkoski, line 339: As it was updating weights expected to find a positionIds in the incoming weights, but it is really just a cached MLXArray.

davidkoski: This avoids the error loading the model.

awni (approved):
LGTM!

PR #254, opened 2025-03-26
Fix #250 -- qwen2vl size configuration moved

### LINKED ISSUE #250: keyNotFound exception when loading model using VLMModelFactory.shared.loadContainer in VLMEval
I'm running the VLMEval application from the main branch and gets the below error when loading ModelRegistry.qwen2VL2BInstruct4Bit.

Failed: keyNotFound(CodingKeys(stringValue: "size", intValue: nil), Swift.DecodingError.Context(codingPath: [], debugDescription: "No value associated with key CodingKeys(stringValue: \"size\", intValue: nil) (\"size\").", underlyingError: nil))

Any clue?

davidkoski:
- size configuration moved from size.min_pixels to just min_pixels
- accept either

### DIFF
--- a/Libraries/MLXVLM/Models/Qwen2VL.swift
+++ b/Libraries/MLXVLM/Models/Qwen2VL.swift
@@ -636,7 +636,7 @@ public class Qwen2VLProcessor: UserInputProcessor {
         let (resizedHeight, resizedWidth) = try targetSize(
             height: Int(size.height), width: Int(size.width),
             factor: config.patchSize * config.mergeSize,
-            minPixels: config.size.minPixels, maxPixels: config.size.maxPixels)
+            minPixels: config.minPixels, maxPixels: config.maxPixels)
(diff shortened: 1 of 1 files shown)


Libraries/MLXVLM/Models/Qwen2VL.swift
```diff
@@ -1040,11 +1040,21 @@ public struct Qwen2VLProcessorConfiguration: Codable, Sendable {
...
+    private let _size: Size?
+    private let _maxPixels: Int?
+    private let _minPixels: Int?
+
+    public var minPixels: Int {
+        _minPixels ?? _size?.minPixels ?? 3136
+    }
+    public var maxPixels: Int {
+        _maxPixels ?? _size?.maxPixels ?? 12_845_056
+    }
```
davidkoski, line 1056: This will accept either top level keys or nested in size (in case somebody has an older config).

PR #255, opened 2025-03-27
Update LLMEval and VLMEval using asyncStream

ronaldmannak:
This draft PR updates the LLMEval and VLMEva example apps using the new asyncStream generate that was merged earlier today (#248 by @Alessan-git). Since asyncStream simplifies generating completions, I assume most developers will be more interested in learning how to use asyncStream than the previous method.

However, I ran into two issues I'm happy to fix.
- How can we cancel the asyncStream? In the example, this happens if maxTokens is reached
- I've reused the `try await modelContainer.perform { context in` closure from the original example. Is there a better way to access the modelContext actor? Right now, we still have to switch back to main actor in the `for await` loop, which seems unnecessary if not inconvenient in case you need to process the data on the original thread.

### DIFF
--- a/Applications/LLMEval/ContentView.swift
+++ b/Applications/LLMEval/ContentView.swift
@@ -165,7 +165,7 @@ class LLMEvaluator {
 
     /// This controls which model loads. `qwen2_5_1_5b` is one of the smaller ones, so this will fit on
     /// more devices.
-    let modelConfiguration = ModelRegistry.qwen2_5_1_5b
+    let modelConfiguration = LLMRegistry.qwen2_5_1_5b
 
     /// parameters controlling the output
     let generateParameters = GenerateParameters(temperature: 0.6)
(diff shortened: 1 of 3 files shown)


davidkoski:
> How can we cancel the asyncStream? In the example, this happens if maxTokens is reached

breaking from the loop should cancel it I think

> I've reused the try await modelContainer.perform { context in closure from the original example. Is there a better way to access the modelContext actor? Right now, we still have to switch back to main actor in the for await loop, which seems unnecessary if not inconvenient in case you need to process the data on the original thread

Hrm, that is an interesting question and maybe we should include @Alessan-git, perhaps this is an oversight in the AsyncStream, but I am not 100% sure.

So the actor that holds the context is used to make sure that there is never concurrent access to the `context` (as nothing in there is `Sendable`).  But in the case of the `AsyncStream` it passes the `context` in to the AsyncStream and into the Task, which now accesses it asynchronously.

But it compiles and I don't see a warning.  I _think_ it may inherit the isolation context because it is being created and used inside the ModelContainer actor.

So the caller of the AsyncStream still needs access to at least the Tokenizer to decode and that is held in the same isolation context, so it makes sense that you would do this inside `try await modelContainer.perform { context`.  Getting the updates back to the view/model requires `@MainActor`, so I think what you have is correct.  I think :-)

ronaldmannak:
> breaking from the loop should cancel it I think

Unfortunately it doesn't. Break works basically as a `continue`. We could do the maxTokens check first before processing the tokens. Technically that prevents the tokens from being processed, but the token generation still continues.

Re: Context, the code does indeed compile but having a closure seems undoing the simplify of an asyncStream.

On a side note, I'm not entirely sure why ModelContext has to be an actor. Correct me if I'm wrong, but it seems that the properties don't have to be vars and Tokenizer shouldn't have any dependencies that can cause race conditions.

Alessan-git:
I truly appreciate your work on this! I have a parallel PR in the works already, and since my changes are substantial, I'll move forward with submitting it separately. Looking forward to your feedback when it's up.

ronaldmannak:
@Alessan-git Sounds good. Since this PR overlaps with yours, I'm closing this PR

davidkoski:
> On a side note, I'm not entirely sure why ModelContext has to be an actor. Correct me if I'm wrong, but it seems that the properties don't have to be vars and Tokenizer shouldn't have any dependencies that can cause race conditions.

`Tokenizer`, specifically `PreTrainedTokenizer` is a class with mutable properties.  It isn't `Sendable`.  The `var` properties could easily be `let` but they are still references to other non `Sendable` classes, so it doesn't help.

It doesn't have to be an actor, but you do need to keep it within a single isolation context.  You can use this variant to load just the context (bundle of inner parts):

```
    public func load(
        hub: HubApi = HubApi(), configuration: ModelConfiguration,
        progressHandler: @Sendable @escaping (Progress) -> Void = { _ in }
    ) async throws -> ModelContext {
```

and own that in an actor, e.g. some kind of model object.

Issue #234, opened 2025-03-10
Structured Chat Messages


ibrahimcetin:
I recently investigating [ollama-swift](https://github.com/mattt/ollama-swift) package and I like [its chat handling](https://github.com/mattt/ollama-swift/blob/main/Sources/Ollama/Chat.swift). Here is a part of it:
```swift
messages: [
    .system("You are a helpful assistant."),
    .user("In which city is Apple Inc. located?")
]
```

I think, we can create an abstraction similar to this on the current implementation.

Example usage:
```swift
let messages: [Qwen2VLChat.Message] = [
    .system("You are a helpful assistant."),
    .user("In which city is Apple Inc. located?"),
    .assistant("Cupertino, San Francisco.")
]
```

Then we can convert this messages to `UserInput` to generate.

I am not an expert and I am still learning but I assume that every model architecture have its own message template. For example, for Qwen2VL requires this:
```
[
    "role": role.rawValue,
    "content": [
        ["type": "text", "text": content]
    ]
        // Messages format for Qwen 2 VL, Qwen 2.5 VL. May need to be adapted for other models.
        + images.map { _ in
            ["type": "image"]
        }
        + videos.map { _ in
            ["type": "video"]
        },
]
```

*If I am wrong please let me know.*

I will continue based on this information from now. So, unlike ollama-swift implementation, we should create a protocol called `Message` and this protocol will have `role`, `content`, `images`, `videos` and maybe `tool`. And we can define a method like `toRawMessage` to convert it to [String: Any] format. The key is `toRawMessage` method. We will implement this method and it converts the messages.

Here is a basic definition, just for giving an idea:
```swift
enum Chat {
    public protocol Message {
        /// The role of the message sender.
        var role: Chat.MessageRole { get }

        /// The content of the message.
        var content: String { get }

        /// Array of image data associated with the message.
        var images: [UserInput.Image] { get }

        /// Array of video data associated with the message.
        var videos: [UserInput.Video] { get }

        static func system(_ content: String, images: [UserInput.Image], videos: [UserInput.Video]) -> Self

        static func assistant(_ content: String, images: [UserInput.Image], videos: [UserInput.Video]) -> Self

        static func user(_ content: String, images: [UserInput.Image], videos: [UserInput.Video]) -> Self

        func toRawMessage() -> [String: Any]
    }

    public enum MessageRole: String {
        case user
        case assistant
        case system
    }
}

extension UserInput {
[...]

PR #257, opened 2025-03-27
Implement Structured Chat Messages

ibrahimcetin:
This PR introduces structured chat messages, addressing issue #234.

I initially made these changes a few days ago but held off on opening a PR because I wanted to create an example app to test them. However, I haven't had time recently, so I’m submitting this now to continue the discussion.

For the implementation, I chose to add a new `chat` case to the `Prompt` enum, as I believe it is the simplest and most compatible approach.

I plan to create a project to demonstrate these changes when I have time.

Looking forward to feedback!


Libraries/MLXLMCommon/Chat.swift
```diff
@@ -0,0 +1,91 @@
...
+
+    public func generate(message: Chat.Message) -> Message {
+        [
+            "role": message.role.rawValue,
+            "content": message.content,
+        ]
+    }
+}
+
+public struct Qwen2VLMessageGenerator: MessageGenerator {
```
davidkoski, line 73: This probably belongs with the Qwen2 code in MLXVLM

Libraries/MLXLMCommon/Chat.swift
```diff
@@ -0,0 +1,91 @@
...
+        public let role: Role
+
+        /// The content of the message.
+        public let content: String
+
+        /// Array of image data associated with the message.
+        public let images: [UserInput.Image]
+
+        /// Array of video data associated with the message.
+        public let videos: [UserInput.Video]
```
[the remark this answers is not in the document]
davidkoski: It might also be a good idea to try and integrate this into the examples -- it will help us make sure this covers the use cases.  Again, that doesn't have to happen in this PR, I think we can iterate on it a bit before we cut tags for a release.

Libraries/MLXLMCommon/UserInput.swift
```diff
@@ -156,6 +150,18 @@ public struct UserInput: Sendable {
         self.additionalContext = additionalContext
     }
 
+    public init(
+        messages: [Chat.Message], images: [Image] = [Image](), videos: [Video] = [Video](),
+        tools: [ToolSpec]? = nil,
+        additionalContext: [String: Any]? = nil
```
davidkoski, line 156: Perhaps we should have:

```swift
public init(
        messages: [Chat.Message],
        tools: [ToolSpec]? = nil,
        additionalContext: [String: Any]? = nil
)
```

and leave off images, etc. -- I think we do not want to have images in both places, and as I mentioned elsewhere I think we need to move off of the top level `images` properties.

davidkoski:
@ibrahimcetin sorry to take so long to get back to this!  I have some suggested changes based on what I was saying / we were discussing about the structure of UserInput -- helping it convert fully to Chat.Message.

How would you like to proceed?

- I can push commits to this PR and you can look at them / revert / whatever
- I can comment with suggested changes, but I think this may be hard to deal with in the PR view
- we can work to get this merged and then I can make a followup PR with my changes

I think the first option is easiest and will let us land this in a more complete state, but please let me know what you think!

ibrahimcetin:
@davidkoski That works for me too, please feel free to proceed in the way that’s easiest for you.

I'm also sorry for the delay. I didn’t expect it to take this long. Starting this weekend I’ll be available again and I’m working on a project that uses these changes.

Thanks a lot for your patience and for following up!

davidkoski:
I was thinking that we should somehow force the prompt to be in the new format, but after playing with that I think that isn't quite right.  I think what you have right now _is_ the right approach.  We have:

- simple text prompt
- raw prompt in the format required by the model
- structured prompt that can be converted into the raw prompt

Getting rid of the raw prompt isn't right, but we can convert the examples to use the structured prompt -- it is easier to read and portable across models.

ibrahimcetin:
Yes, I agree with you and I believe these changes are mostly ready. We should also add some unit tests to ensure everything works as expected.

I also think the examples should be updated, so I’d suggest handling that in a separate PR.

Let me know if you have any other thoughts or suggestions.

davidkoski:
I am concerned that the UserInput.images and prompt can get out of sync -- I think we can make that work using the same thing you did in the init for the Chat (that is what I am playing with now).

davidkoski:
OK, got the tests building -- just something trivial to start with.

Libraries/MLXLMCommon/UserInput.swift
```diff
@@ -124,44 +128,196 @@ public struct UserInput: Sendable {
...
+    /// The prompt to evaluate.
     public var prompt: Prompt
-    public var images = [Image]()
-    public var videos = [Video]()
+
+    /// The images associated with the `UserInput`.
+    ///
+    /// If the ``prompt-swift.property`` is a ``Prompt-swift.enum/chat(_:)`` this will
+    /// collect the images from the chat messages, otherwise these are the stored images with the ``UserInput``.
+    public var images: [Image] {
```
davidkoski, line 138: @ibrahimcetin I made two changes of note, see what you think of them.  The first is to make `images` and `video` reflect the chat messages OR store the values.  This way you can mutate the prompt and these properties will still contain the right contents.

Libraries/MLXLMCommon/UserInput.swift
```diff
@@ -124,44 +128,196 @@ public struct UserInput: Sendable {
...
+    /// - Parameters:
+    ///   - chat: structured content
+    ///   - tools: optional tool specifications
+    ///   - additionalContext: optional context (model specific)
+    /// ### See Also
+    /// - ``Prompt-swift.enum/text(_:)``
+    /// - ``init(chat:tools:additionalContext:)``
     public init(
-        prompt: Prompt, images: [Image] = [Image](), processing: Processing = .init(),
+        chat: [Chat.Message],
```
davidkoski, line 283: The second change is to change this from `messages: [Chat.Message]` to `chat: [Chat.Message]`.  I did this for two reasons:

- so that it might be more obvious when you do code completion what the different is between this and the raw messages
- so that the links in the documentation would resolve correctly (!)

Tools/llm-tool/LLMTool.swift
```diff
@@ -235,31 +235,12 @@ struct EvaluateCommand: AsyncParsableCommand {
...
-                    ]
-                ]
-            }
-        var userInput = UserInput(messages: messages, images: images, videos: videos)
+        var userInput = UserInput(
+            chat: [
+                .system(generate.system),
+                .user(prompt, images: images, videos: videos),
+            ],
+        )
```
davidkoski, line 243: @ibrahimcetin you had mentioned converting the calls to use this in another PR, which I think is fine, but I do think we should have one of them converted so we can exercise it.

This is a _LOT_ nicer!

davidkoski: Also note this fixes #270

davidkoski:
I am going to rebase on main and then write some tests

davidkoski:
@ibrahimcetin I really like what you have built!  Take a look at the couple of changes I made and see what you think.  I also added tests and some documentation.

I think in a follow-on PR(s) we can:

- convert the rest of the examples to use the structured format
- consider this code https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/prompt_utils.py#L23
    - we might want to create generic forms like this so that they can be reused across several models
- convert the other VLMs to use MessageGenerators

ibrahimcetin:
@davidkoskiThank you for these changes, and most of the changes look good to me. Please review my comments on the code.

I agree with you on following-on PRs and I can update the examples as soon as this PR is merged. I also consider to update ToolSpec similar to [the ollama-swift tool](https://github.com/loopwork-ai/ollama-swift/blob/main/Sources/Ollama/Tool.swift).

davidkoski (approved):
Awesome addition, thank you!

davidkoski:
I didn't see any new comments -- anything left before we merge?

ibrahimcetin:
> I didn't see any new comments -- anything left before we merge?

My mistake, sorry. Now you are able to see.

Libraries/MLXLMCommon/UserInput.swift
ibrahimcetin: This is just a thought. What do you think?

davidkoski: yes, good idea -- I think that would simplify things

Libraries/MLXVLM/Models/SmolVLM2.swift
```diff
@@ -221,7 +221,7 @@ public class SmolVLMProcessor: UserInputProcessor {
     }
 
     public func prepare(input: UserInput) async throws -> LMInput {
-        let messages = input.prompt.asMessages()
+        let messages = Qwen2VLMessageGenerator().generate(from: input)  // TODO: Create SmolVLM2MessageGenerator
```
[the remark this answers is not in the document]
davidkoski: I guess it is roughly equivalent to what we had before with the raw messages.  Perhaps we address it with the followup:

- consider this code https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/prompt_utils.py#L23

davidkoski:
Good feedback!  I made the changes to the images/videos properties as you suggested -- ready to merge?

davidkoski (approved):
Excellent addition, thank you!

PR #259, opened 2025-03-31
Feature/generate completion info

ronaldmannak:
This PR makes the properties of GenerateCompletionInfo public


davidkoski (changes requested):
Looks good, but it should probably also have an `init` like `GenerateResult` does -- in case somebody wants to implement their own asynchronous iterator.

ronaldmannak:
@davidkoski added!

davidkoski:
It looks like CI failed on swift-format checks

ronaldmannak:
@davidkoski Both `swift-format --in-place Libraries/MLXLMCommon/Evaluate.swift` and `swift-format format --in-place --recursive Libraries Tools Applications` pass on my machine. Is there anything special I need to do?

davidkoski (approved):
thanks for these additions!

Issue #252, opened 2025-03-26
detokenizer.next()  Characters that produce garbled text when the response is in Chinese.


hsduren:
return try MLXLMCommon.generate(
                        input: input, parameters: generateParameters, context: context
                    ) { tokens in

                        if let last = tokens.last {
                            detokenizer.append(token: last)
                        }

                        if let word = detokenizer.next() {
                  
                                    DispatchQueue.main.async {
                                        // print("char:",word)
                                        onReceive(word)
                                    }
                            
                        }

davidkoski:
Can you be more specific?  What does the text look like / how is it garbled?  Which model and what type of inputs?

I see the Qwen2VL model produce Chinese output and although I can't read it, it translates back to English and looks ok.

hsduren:
use model : mlx-community/DeepSeek-R1-Distill-Qwen-1.5B-4bit
input: 如何学习swift
output:
 see the attach

You can reproduce the issue in the ​Furgent app (available on the Mac App Store).

<img width="1198" alt="Image" src="https://github.com/user-attachments/assets/27750a04-e8cd-426e-b3ef-4bf32f759246" />

davidkoski:
Can you reproduce this with the `llm-tool` or `LLMEval` example applications?  I tried with the python and swift versions and got:

```
python -m mlx_lm.generate --model mlx-community/DeepSeek-R1-Distill-Qwen-1.5B-4bit --prompt "如何学习swift"

<frozen runpy>:128: RuntimeWarning: 'mlx_lm.generate' found in sys.modules after import of package 'mlx_lm', but prior to execution of 'mlx_lm.generate'; this may result in unpredictable behaviour
Calling `python -m mlx_lm.generate...` directly is deprecated. Use `mlx_lm.generate...` or `python -m mlx_lm generate ...` instead.
Fetching 6 files: 100%|██████████████████████████████████████████████████████| 6/6 [00:00<00:00, 61380.06it/s]
==========
嗯，我现在想学习Swift，但是我对编程还不是很熟悉，特别是C语言。我需要先从哪里开始呢？听说学习C语言对学习Swift很重要，所以可能得先复习一下C语言的基础知识。那我应该先学哪些内容呢？

首先，我应该学C语言的语法吧。C语言有很多语法，比如变量声明、输入输出、条件语句、循环等等。我得先把这些语法都搞清楚，特别是输入输出，因为SW
```

```
[...]

PR #260, opened 2025-04-01
Decode token in generate()

ronaldmannak:
This issue was discussed in #248 by @Alessan-git. The current implementation returns the token instead of the decoded chunk. The (admittedly minor) issue I ran into was that the tokenizer is not thread-safe. In my server-side Swift application had to create a new structure to store the encoded chunk to pass it on the callers. Not a huge deal, but probably something other people will run into as well. If this PR gets committed, then #256 will have to be updated as well. Alternatively, I can add a second `generate` method that returns a new `ChunkGeneration` enum, if that doesn't get too messy. Anyway, I just wanted to revisit the discussion.


Libraries/MLXLMCommon/Evaluate.swift
davidkoski, line 1: I like the idea -- I think this is easier to use/consume than raw tokens.  I wonder if the `generate()` function could consume the `ModelContainer` and encapsulate the use of the actor.  This might enable a much more natural looking call site as well.

Are there any cases where tokens would be preferred?  I am not sure, but it makes me wonder if we should be using some composition here -- a slightly more complex implementation, but what if the call to `generate()` took a function to consume the tokens and produce the values to yield, e.g. something like:

```swift
func generate<T>(..., processor: (Int) -> T) -> AsyncStream<T> where T: Sendable
```

The default implementation of the processor could do this conversion to String but a caller could provide their own to customize it.  `T` might need to be `IteratorCommand<T>`, some kind of enum to control processing.

We don't need to necessarily go there with this PR but something to think about -- I think we can get a lot of flexibility from composition and default implementations can make the common path easy.

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -749,7 +749,8 @@ public func generate(
                 }
 
                 tokenCount += 1
-                continuation.yield(.token(token))
+                let chunk = context.tokenizer.decode(tokens: [token])
+                continuation.yield(.chunk(chunk))
```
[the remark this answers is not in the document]
ronaldmannak: Thanks! Fixed in the latest commit.
I copied the line from the example apps (see below). Is there a reason it does work in the example apps?

https://github.com/ml-explore/mlx-swift-examples/blob/cb66b4bc6bc1a69663837881e7f1260cd49d6b59/Applications/LLMEval/ContentView.swift#L264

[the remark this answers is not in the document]
davidkoski: Yes, the reason it works there is it uses the full token array, not just `[token]` and recreates the entire string.  This is actually O(n^2) cost, but as long as n is small it is still very fast (and it is in the example).  But not a great thing to have in the example code for people to copy :-)

Libraries/MLXLMCommon/Evaluate.swift
ronaldmannak: I personally can't think of a scenario where I would prefer the raw tokens besides debugging, but you don't need to receive the tokens for that.
Can you share some example or pseudo code what you mean by "consume the `ModelContainer` and encapsulate the use of the actor"?

davidkoski: I meant this comment specifically:

> I've reused the try await modelContainer.perform { context in closure from the original example. Is there a better way to access the modelContext actor? Right now, we still have to switch back to main actor in the for await loop, which seems unnecessary if not inconvenient in case you need to process the data on the original thread.

I wonder if we can avoid that by returning text?

Libraries/MLXLMCommon/Evaluate.swift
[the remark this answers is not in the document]
davidkoski: I am not sure the conditions where it might fail -- perhaps @awni or @angeloskath knows.  The technique is based on:

- https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/tokenizer_utils.py#L58

which was built for this reason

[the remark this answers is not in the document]
davidkoski: Actually, see #252 for an example where the one-by-one doesn't work (though in this case there is also a problem with the streaming detokenizer)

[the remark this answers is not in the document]
Alessan-git: Yes, it's true that this model (like many others) uses byte-level tokenization. Decoding one token at a time can lead to incorrect or incomplete reconstructions because the tokens need to be combined to form a valid byte sequence that maps to human-readable text. I forgot about this since I usually work with English text.

We must always combine contiguous tokens (except for those marked as special) to ensure we reconstruct the full byte-level representation before attempting to decode it into a Unicode string.

Libraries/MLXLMCommon/Evaluate.swift
ronaldmannak: @davidkoski I'm hopping on a plane tomorrow and be less available the coming weeks. If you don't mind and agree, why don't we merge this PR first and move the more comprehensive refactoring to a new PR?

davidkoski: Sounds good -- it is still in Draft so I wasn't sure where you were with it.

ronaldmannak: Ah sorry, I forgot it was still a draft. Changed it to ready for review

davidkoski: Looks nice!

```swift
        for await item in try MLXLMCommon.generate(input: input, parameters: generateParameters, context: context) {
            switch item {
            case .chunk(let string):
                print(string, terminator: "")
                fflush(stdout)
            case .info(let generateCompletionInfo):
                break
            }
        }
```

davidkoski (approved):
Thank you for this continued contribution!

PR #261, opened 2025-04-03
if the tokenizer produces an incomplete character do not consume

### LINKED ISSUE #252: detokenizer.next()  Characters that produce garbled text when the response is in Chinese.

davidkoski:
- fix #252

### DIFF
--- a/Libraries/MLXLMCommon/Tokenizer.swift
+++ b/Libraries/MLXLMCommon/Tokenizer.swift
@@ -136,6 +136,12 @@ public struct NaiveStreamingDetokenizer: StreamingDetokenizer {
         let newSegment = tokenizer.decode(tokens: segmentTokens)
         let new = newSegment.suffix(newSegment.count - segment.count)
 
+        // if the new segment ends with REPLACEMENT CHARACTER this means
+        // that the token didn't produce a complete unicode character
+        if new.last == "\u{fffd}" {
+            return nil
+        }
+
         if new.hasSuffix("\n") {
             startNewSegment()
         } else {


Libraries/MLXLMCommon/Tokenizer.swift
```diff
@@ -136,6 +136,12 @@ public struct NaiveStreamingDetokenizer: StreamingDetokenizer {
         let newSegment = tokenizer.decode(tokens: segmentTokens)
         let new = newSegment.suffix(newSegment.count - segment.count)
 
+        // if the new segment ends with REPLACEMENT CHARACTER this means
+        // that the token didn't produce a complete unicode character
+        if new.last == "\u{fffd}" {
+            return nil
+        }
```
davidkoski, line 143: From #215 

```swift
        await modelContainer.perform { context in
            let tokens = [55336, 76952, 122, 23836, 99491, 105266]
            var detokenizer = NaiveStreamingDetokenizer(tokenizer: context.tokenizer)
            for token in tokens {
                detokenizer.append(token: token)
                if let t = detokenizer.next() {
                    print(t, terminator: "")
                }
            }
        }
```

produces a REPLACEMENT CHARACTER from a partial character from the detokenizer:

Swift �区非常活跃

if we see this we can just continue decoding without consuming the new section yet.

Alessan-git: I did the same on my custom detokenizer, but it's a great idea to have it in the builtin NaiveStreamingDetokenizer!

awni:
Should we follow the Python NaiveStreamingDetokenizer implementation? https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/tokenizer_utils.py#L78-L100 ?

I don't think it has this problem and relies on the tokenizer decode method to return the right result.

davidkoski:
> Should we follow the Python NaiveStreamingDetokenizer implementation? https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/tokenizer_utils.py#L78-L100 ?
> 
> I don't think it has this problem and relies on the tokenizer decode method to return the right result.

The reason is string handling in python vs swift.  Python is largely a bag of bytes and so if you get part of a code point and append the rest later it works fine.  Swift deals with codepoints only, so a partial codepoint does not concatenate properly with another partial codepoint.  This just detects that case and doesn't produce a string if it has "REPLACEMENT CHARACTER" indicating that it has a partially formed codepoint.

awni (approved):

awni:
Thanks for fixing that!

PR #272, opened 2025-04-16
use local pre-commit hook for swift-format

nwh:
change to a local pre-commit repo to call `swift-format`


davidkoski (approved):
Thank you for the contribution!

Issue #276, opened 2025-04-21
Proposal: Add MLX Chat Example App to the Repository


ibrahimcetin:
Hi, I developed an example chat app as I promised in [this comment](https://github.com/ml-explore/mlx-swift-examples/issues/234#issuecomment-2715760810).

## MLXChatExample App

[MLX Chat Example](https://github.com/ibrahimcetin/MLXChatExample) is simple, practical, and modular. It supports both LLMs and VLMs, along with most features you’d expect from a modern chat app—cancellation, real-time generation, model switching, and chat clearing. The codebase is modular, with a clear separation between UI and business logic and it's well documented. However, it is not intended for production use—rather, I consider it a sandbox app for experimentation.

## The Advantage of Structured Chat Messages

I developed this app to explore the benefits of structured chat messages. As shown [here](https://github.com/ibrahimcetin/MLXChatExample/blob/1ce79d2843a8b58031a3f92d6bc4d11e58abe1d3/MLXChatExample/Services/MLXService.swift#L97), this approach significantly simplifies the codebase and the usage of the package by abstracting away model-specific differences. It allows developers to easily use both LLMs and VLMs without having to manage model-specific chat structures manually.

## One Problem

Initially, I tried creating this app inside the existing package (in the `Applications` folder), but I ran into issues binding the frameworks (`MLXLLM`, `MLXVLM`, `MLXLMCommon`) to the application target and couldn't find a solution. Here are the steps I followed:

1. Open `mlx-swift-examples.xcodeproj` in Xcode
2. Add a new target to the project
3. Select “App” from the templates
4. Create a SwiftUI project named `MLXChatExample`
5. Go to **Frameworks, Libraries, and Embedded Content**, click the plus button
6. Xcode should list the products from `mlx-swift-examples`, but they aren't shown—so I can't add them

I tried removing and re-adding the local package (which resets references in existing projects), and after that, Xcode successfully listed the products. However, this approach modifies the .xcodeproj file unnecessarily. Aside from that workaround, I haven’t found a clean solution.

What am I doing wrong?
Is there a proper way to add this project to the package as an example?
Or should we even include example apps directly in the package?

## Conclusion

I believe `mlx-swift-examples` could really benefit from an example like this. If I can solve the integration issue, I'd love to contribute this app to the repository. But I also think it’s worth discussing whether we want to include these kinds of examples directly within the package.

davidkoski:
> What am I doing wrong?

[...]

PR #277, opened 2025-04-21
chat example (command line)

davidkoski:


Libraries/MLXLMCommon/UserInput.swift
```diff
@@ -41,7 +41,7 @@ public struct UserInput: Sendable {
     }
 
     /// Representation of a video resource.
-    public enum Video: Sendable {
+    public enum Video: Sendable, Hashable {
```
davidkoski, line 44: Not sure I need this, probably revert

Libraries/MLXVLM/Models/Qwen2VL.swift
```diff
@@ -103,15 +103,15 @@ private enum Language {
...
-            let mask = mask?[0..., 0 ..< keys.dim(-2)]
-
             queries = rotaryEmbedding(queries, offset: offset)
             keys = rotaryEmbedding(keys, offset: offset)
 
             if let cache {
                 (keys, values) = cache.update(keys: keys, values: values)
             }
 
+            let mask = mask?[.ellipsis, 0 ..< keys.dim(-2)]
```
davidkoski, line 113: I found that the dimensions were mismatched on a second message -- the keys dimension needs to be considered _after_ the KVCache.  In mlx-vlm it works out ok because:

1) it looks like KVCache isn't used persistently
2) the KVCache implementation doesn't window on the cache

Tools/llm-tool/LLMTool.swift
```diff
@@ -44,16 +44,28 @@ struct ModelArguments: ParsableArguments, Sendable {
     }
 }
 
-/// Command line arguments for controlling generation of text.
-struct GenerateArguments: ParsableArguments, Sendable {
-
+struct PromptArguments: ParsableArguments, Sendable {
```
davidkoski, line 47: I split this out because `chat` doesn't need a prompt.

Tools/llm-tool/LLMTool.swift
```diff
@@ -301,3 +306,164 @@ struct EvaluateCommand: AsyncParsableCommand {
...
+                // Single value represents width/height
+                let v = resize[0]
+                size = CGSize(width: v, height: v)
+            } else {
+                let v0 = resize[0]
+                let v1 = resize[1]
+                size = CGSize(width: v0, height: v1)
+            }
+            userInput.processing.resize = size
+        }
```
davidkoski, line 350: The processing part is the only thing that is interesting I think -- this should probably go into reusable arguments anyway

Tools/llm-tool/LLMTool.swift
```diff
@@ -301,3 +306,164 @@ struct EvaluateCommand: AsyncParsableCommand {
...
+
+        // Get the resolved configuration (this has the default prompt)
+        let modelConfiguration = await modelContainer.configuration
+
+        var userInput = self.userInput(modelConfiguration: modelConfiguration)
+        var chat = [Chat.Message.system(generate.system)]
+
+        // TODO: need to figure out the proper ownrship for this -- maybe the loop
+        // below needs to go inside the context?
+        var cache: [KVCache]?
```
davidkoski, line 385: This still needs work, but I think a chat application should show how to use a KVCache

Tools/llm-tool/LLMTool.swift
```diff
@@ -301,3 +306,164 @@ struct EvaluateCommand: AsyncParsableCommand {
...
+
+            // TODO: this works fine with a single image (in the chat) but how do we
+            // deal with multiple images in the conversation?  note that it only gets injected once
+            // (which isn't quite right) and the KVCache keeps hold of it but we end up
+            // reprocessing it multiple times.  anyway, clean this up
+            var chatWithMedia = chat
+            if chat.count >= 2 {
+                chatWithMedia[1].images = images
+                chatWithMedia[1].videos = videos
+            }
```
davidkoski, line 424: @ibrahimcetin 

Still working on this -- it works great with a single set of media, but the model / image processing code doesn't expect two messages with independent sets of media.

On the swift side:

```swift
public struct LMInput {
    public let text: Text
    public let image: ProcessedImage?
    public let video: ProcessedVideo?
```

Tokens + one set of media.  The python code in mlx-vlm has a similar setup.  I think maybe that should be:

```swift
public struct LMInput {
    public let text: Text
    public let image: [ProcessedImage]
    public let video: [ProcessedVideo]
```

and then the model can inject the appropriate piece.

The `Hashable` conformance in the `UserInput` was me thinking that perhaps we should cache the `ProcessedImage` / `Video` so it doesn't need to be reprocessed.

Another option is to use the `KVCache` -- it will capture the image embedding and we could then omit the image from subsequent generations.  Except we need to represent it space-wise so the size of the context remains consistent (we can't just drop the image).

FWIW @Blaizzy mlx-vlm's chat implementation has a problem here I think:

- https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/chat_ui.py

The code only presents the image for the first message and then discards it.  It doesn't use a KVCache so it actually "forgets" what the image is, though the model is pretty good at faking it:

```
<|im_start|>system
You are a helpful assistant.<|im_end|>
<|im_start|>user
what animal is in the picture.  answer in english<|im_end|>
<|im_start|>assistant
The animal in the picture is a dog.<|im_end|>
<|im_start|>user
what is the dog wearing?<|im_end|>
<|im_start|>assistant
The dog is wearing a collar.<|im_end|>
<|im_start|>user
what is behind the dog?<|vision_start|><|image_pad|><|vision_end|><|im_end|>
<|im_start|>assistant
There is a person behind the dog.<|im_end|>
```

img

There are maybe two problems here:

- dropping the image data
- the fact that the image is attached to the last message -- I found that if the image data is present it needs to be associated with the first message that refers to it

Anyway, with just a single image I think this is pretty close.  It needs to be cleaned up, but it works.  Maybe we only deal with one image set in the thread for now.

Tools/llm-tool/LLMTool.swift
```diff
@@ -301,3 +306,164 @@ struct EvaluateCommand: AsyncParsableCommand {
...
+
+            let (result, output) = try await modelContainer.perform {
+                [generate, userInput] context in
+                let input = try await context.processor.prepare(input: userInput)
+                //                print(context.tokenizer.decode(tokens: input.text.tokens.asArray(Int.self)))
+
+                // TODO: figure out ownership here
+                if cache == nil {
+                    cache = context.model.newCache(parameters: generate.generateParameters)
+                }
```
davidkoski, line 437: The cache is mutable and I am playing fast and loose with ownership here.  To be fixed.

Tools/llm-tool/LLMTool.swift
```diff
@@ -301,3 +306,164 @@ struct EvaluateCommand: AsyncParsableCommand {
...
+
+                // TODO: does this way so we can pass the cache.  maybe
+                // it should be a paramter to the higher level generate?
+                var iterator = try TokenIterator(
+                    input: input, model: context.model, cache: cache,
+                    parameters: generate.generateParameters)
```
davidkoski, line 443: I don't care about the iterator -- I think the `generate()` call should probably take an optional cache.

Tools/llm-tool/LLMTool.swift
davidkoski: @ibrahimcetin an example of the multi-image in chat issue:

image

so perhaps we document that as an issue (or disable to ability to add multiple media sets) until we figure it out.  I don't think it needs to block the chat examples.

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -729,10 +730,10 @@ public func generate(
 /// }
 /// ```
 public func generate(
-    input: LMInput, parameters: GenerateParameters, context: ModelContext
+    input: LMInput, cache: [KVCache]? = nil, parameters: GenerateParameters, context: ModelContext
```
davidkoski, line 733: Easily pass the optional cache in to the iterator

Libraries/MLXLMCommon/KVCache.swift
```diff
@@ -97,4 +97,7 @@ public class KVCacheSimple: KVCache, Evaluatable {
         )
     }
 
+    public var debugDescription: String {
+        "\(String(describing: Self.self)) \(Unmanaged.passUnretained(self).toOpaque()), offset: \(offset), step: \(step), keys: \(keys?.shape.description ?? "-"), values: \(values?.shape.description ?? "-")"
```
davidkoski, line 101: Just a convenience I needed while debugging Qwen2

Tools/llm-tool/Chat.swift
```diff
@@ -0,0 +1,222 @@
...
+
+    func chat(modelContainer: ModelContainer) async throws {
+        try await modelContainer.perform { context in
+            let parameters = generate.generateParameters
+            let initialState = State(
+                parameters: parameters,
+                processing: media.processing,
+                images: media.images, videos: media.videos,
+                chat: [.system(generate.system)],
+                cache: context.model.newCache(parameters: parameters))
```
davidkoski, line 66: We may want a follow up PR for #276 to add the KVCache, but without the change to `generate()` it is hard to use.

Tools/llm-tool/LLMTool.swift
```diff
@@ -44,16 +44,66 @@ struct ModelArguments: ParsableArguments, Sendable {
...
+            let names = prompt.split(separator: ",").map { String($0.dropFirst()) }
+            return try names.map { try String(contentsOfFile: $0) }.joined(separator: "\n")
+        } else {
+            return prompt
+        }
+    }
+}
+
+/// Argument package for supplying media files
+struct MediaArguments: ParsableArguments, Sendable {
```
davidkoski, line 67: Also split this out because we use it in `eval` and `chat`

PR #280, opened 2025-04-23
Add quantization support to Qwen2 Model

gingofthesouth:
When attempting to load a quantized Deepcoder model I ran into issues due to mismatched parameter weight shape,
`❌ Error: Failed to load model weights: Mismatched parameter weight shape. Actual [152064, 640], expected [152064, 5120]'

This small enhancement adds quantisation decoding to the Qwen2 model config allowing us to pass in the quantisation to the loadWeights function.

Example loading code below:

```swift
 fileprivate func loadModel(_ modelURL: URL) throws {
        if self.model == nil {
            
            // Path to Config file
            let configURL = modelURL.appendingPathComponent("config.json")

            // Load config json as data
            let configData: Data
            do {
                configData = try Data(contentsOf: configURL)
            } catch {
                throw LoadingError.configLoadFailed(error)
            }

            // Decode the config
            let config: Qwen2Configuration
            do {
                config = try JSONDecoder().decode(Qwen2Configuration.self, from: configData)
            } catch {
                throw LoadingError.configLoadFailed(error)
            }

            // Load the model with quantization metadata from config
            let loadedModel = Qwen2Model(config)
            do {
                try MLXLMCommon.loadWeights(modelDirectory: modelURL, model: loadedModel, quantization: config.quantization)   //  <--- Function called with the config quantisation metadata
                logQuantizedModules(loadedModel)          //  <--- Add this testing and debugging
            } catch {
                throw LoadingError.weightLoadFailed(error)
            }

            self.model = loadedModel
        }
    }
    
    ...
    
 fileprivate func logQuantizedModules(_ model: Module) {
        for (path, module) in model.leafModules().flattened() {
            if module is QuantizedLinear {
                print("✅ QuantizedLinear at: \(path)")
            } else if module is QuantizedEmbedding {
                print("✅ QuantizedEmbedding at: \(path)")
            }
        }
    }
```

Steps to test:
1. Download a quantized model (such as https://huggingface.co/mlx-community/DeepCoder-1.5B-Preview-4bit).
2. Load the model using code similar to the example above.
3. Check the model loads successfully and no error is thrown.
4. Optionally use the logging utility function above to verify it prints out the

### DIFF
--- a/Libraries/MLXLLM/Models/Qwen2.swift
+++ b/Libraries/MLXLLM/Models/Qwen2.swift
@@ -221,7 +221,7 @@ public struct Qwen2Configuration: Codable, Sendable {
     var ropeTraditional: Bool = false
     var ropeScaling: [String: StringOrNumber]? = nil
     var tieWordEmbeddings = false
-    var quantization: BaseConfiguration.Quantization?
+    public var quantization: BaseConfiguration.Quantization?
 
     enum CodingKeys: String, CodingKey {
         case hiddenSize = "hidden_size"
--- a/Libraries/MLXLLM/Models/Qwen2.swift
+++ b/Libraries/MLXLLM/Models/Qwen2.swift
@@ -221,6 +221,7 @@ public struct Qwen2Configuration: Codable, Sendable {
     var ropeTraditional: Bool = false
     var ropeScaling: [String: StringOrNumber]? = nil
     var tieWordEmbeddings = false
+    var quantization: BaseConfiguration.Quantization?
 
     enum CodingKeys: String, CodingKey {
         case hiddenSize = "hidden_size"
@@ -234,6 +235,7 @@ public struct Qwen2Configuration: Codable, Sendable {
         case ropeTraditional = "rope_traditional"
         case ropeScaling = "rope_scaling"
         case tieWordEmbeddings = "tie_word_embeddings"
+        case quantization = "quantization"
     }
 
     public init(from decoder: Decoder) throws {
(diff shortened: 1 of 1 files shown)


Libraries/MLXLLM/Models/Qwen2.swift
```diff
@@ -221,6 +221,7 @@ public struct Qwen2Configuration: Codable, Sendable {
     var ropeTraditional: Bool = false
     var ropeScaling: [String: StringOrNumber]? = nil
     var tieWordEmbeddings = false
+    public var quantization: BaseConfiguration.Quantization?
```
davidkoski, line 224: This is normally handled via https://github.com/ml-explore/mlx-swift-examples/blob/main/Libraries/MLXLMCommon/LanguageModel.swift#L217

You would load it like this:

```swift
        // load the generic config to understand which model and how to load the weights
        let configurationURL = modelDirectory.appending(component: "config.json")
        let baseConfig = try JSONDecoder().decode(
            BaseConfiguration.self, from: Data(contentsOf: configurationURL))

        // apply the weights to the bare model
        try loadWeights(
            modelDirectory: modelDirectory, model: model, quantization: baseConfig.quantization)
```

You have a custom loader so you didn't need this directly -- typically we need this in order to look up the model type before we know which type of configuration we have.

Can you just use this same technique?  Otherwise it would make sense to add this to all the models, but the typical load path can't make use of it.

davidkoski: See also #293

Issue #283, opened 2025-04-26
Remove dead dependency: AsyncAlgorithms


gtokman:
I see a PR that removed the `AsyncAlgorithms` package and it was merged, but on main it still exists. 

PR to remove: https://github.com/ml-explore/mlx-swift-examples/pull/100

Main: https://github.com/search?q=repo%3Aml-explore%2Fmlx-swift-examples%20import%20AsyncAlgorithms&type=code

It's a dead depenedency and I can make a PR to remove it, but want to make sure I'm not missing something.

davidkoski:
I agree it isn't used by the Libraries, but it is now used by the Applications.

PR #287, opened 2025-04-29
Remove unused package AsyncAlgorithms

gtokman:
#283

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -31,8 +31,6 @@ let package = Package(
         .package(
             url: "https://github.com/huggingface/swift-transformers", .upToNextMinor(from: "0.1.17")
         ),
-        .package(
-            url: "https://github.com/apple/swift-async-algorithms", .upToNextMinor(from: "1.0.0")),
         .package(url: "https://github.com/1024jp/GzipSwift", "6.0.1" ... "6.0.1"),  // Only needed by MLXMNIST
     ],
     targets: [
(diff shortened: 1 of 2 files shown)


Applications/LLMEval/ContentView.swift
```diff
@@ -1,6 +1,5 @@
 // Copyright © 2024 Apple Inc.
 
-import AsyncAlgorithms
```
davidkoski, line 3: I think this one is used -- see:

```
                for await batch in stream._throttle(
                    for: updateInterval, reducing: Generation.collect)
                {
```

Applications/VLMEval/ContentView.swift
```diff
@@ -1,7 +1,6 @@
 // Copyright 2024 Apple Inc.
 
 import AVKit
-import AsyncAlgorithms
```
davidkoski, line 4: This is also used

Package.swift
```diff
@@ -31,8 +31,6 @@ let package = Package(
         .package(
             url: "https://github.com/huggingface/swift-transformers", .upToNextMinor(from: "0.1.17")
         ),
-        .package(
-            url: "https://github.com/apple/swift-async-algorithms", .upToNextMinor(from: "1.0.0")),
```
davidkoski, line 35: I agree this one can go

Applications/VLMEval/ContentView.swift
gtokman: I compiled the scheme for the library and not the apps, good catch!

gtokman:
@davidkoski responded to feedback!

PR #288, opened 2025-04-29
Log chat template error

DePasqualeOrg:
The silent error made debugging more difficult.


Libraries/MLXLLM/LLMModelFactory.swift
```diff
@@ -218,8 +218,9 @@ private struct LLMUserInputProcessor: UserInputProcessor {
                 messages: messages, tools: input.tools, additionalContext: input.additionalContext)
             return LMInput(tokens: MLXArray(promptTokens))
         } catch {
-            // #150 -- it might be a TokenizerError.chatTemplate("No chat template was specified")
-            // but that is not public so just fall back to text
+            // It might be TokenizerError.chatTemplate("No chat template was specified"),
+            // but that is not public, so just fall back to text
+            print("Error while applying chat template: \(error.localizedDescription)")
```
davidkoski, line 223: This catch was added for the case where the tokenizer didn't have a chat template at all and to give it something to fall back on.  At the time `TokenizerError` wasn't public so we couldn't catch it, but now it is and we want this catch to _only catch_ this case:

```swift
        guard let selectedChatTemplate else {
            throw TokenizerError.chatTemplate("This tokenizer does not have a chat template, and no template was passed.")
        }
```

What do you think about doing this?

```swift
        } catch TokenizerError.chatTemplate(let message) {
            if message == "This tokenizer does not have a chat template, and no template was passed." {
                // fallback case
                let prompt =
                    messages
                    .compactMap { $0["content"] as? String }
                    .joined(separator: ". ")
                let promptTokens = tokenizer.encode(text: prompt)
                return LMInput(tokens: MLXArray(promptTokens))
            } else {
                throw TokenizerError.chatTemplate(message)
            }
        }
```

That way real errors get propagated as error and this fixes #150

DePasqualeOrg: That makes sense, although it would probably be better to have a dedicated error case for this type of error in swift-transformers and check for that case rather than the exact error message. I'll open a PR for that.

davidkoski: I agree, a dedicated error case would be ideal.

PR #289, opened 2025-04-29
refactor: Improve model loading with structured progress tracking

matiasvillaverde:
## Small improvement on progress tracking for model loading operations

Added structured progress tracking across our model factories with defined percentage ranges for each loading stage. This gives users clearer feedback during initialization.

## Changes

- Added percentage-based progress tracking (download 0-30%, config 30-50/60%, weights 50/60-80%, tokenizer 80-100%)
- Fixed hub API reference in StableDiffusion model container
- Added support for loading SD models from custom local directories
- Standardized parameter handling in GenerateParameters

The progress indicators now show meaningful percentages instead of jumping unpredictably, which helps users understand what's happening during longer loading operations.

### DIFF
--- a/Libraries/MLXLLM/LLMModelFactory.swift
+++ b/Libraries/MLXLLM/LLMModelFactory.swift
@@ -260,29 +260,49 @@ public class LLMModelFactory: ModelFactory {
         hub: HubApi, configuration: ModelConfiguration,
         progressHandler: @Sendable @escaping (Progress) -> Void
     ) async throws -> ModelContext {
-        // download weights and config
-        let modelDirectory = try await downloadModel(
-            hub: hub, configuration: configuration, progressHandler: progressHandler)
+        // Create progress tracker
+        let progress = Progress(totalUnitCount: 100)
 
-        // load the generic config to unerstand which model and how to load the weights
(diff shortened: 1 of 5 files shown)


DePasqualeOrg:
But the weights are the vast majority, perhaps 99% or more, of the download size. In my app Local Chat I just disregard the config files in the download progress and calculate progress based on the number of `.safetensors` files and their total size. If you want to do this accurately, you can get the size of the files through the Hugging Face Hub API. You can look at the Python Hub API client to see how to do it in Swift.

matiasvillaverde:
@DePasqualeOrg Yes, that sounds like a much better idea. Unfortunately, I don't have the time to implement it atm.

If you think this PR is not bringing value, even as an intermediate step. Feel free to close it.

davidkoski:
I agree, the download is by far the bulk of the time ... _if_ the files need to be downloaded.  HubApi internally has a way to determine which files need to be downloaded, but it isn't exposed as API (see internals of `HubApi.snapshot`).  If it were, I think we could do a better job of estimating what work needs to happen and where.

I think what you have is a reasonable idea, but without this extra information I think it ends up being a bit misleading progress-wise.

PR #290, opened 2025-04-29
swift-format

davidkoski:
Somehow #286 didn't have CI run on it and is missing swift-format.

PR #293, opened 2025-04-29
fix #291 -- support heterogenous quant config

### LINKED ISSUE #291: Qwen3 with heterogenous quant doesn't work
The following model `mlx-community/Qwen3-1.7B-4bit-AWQ` doesn't run in the mlx-swift-examples repo doesn't run. It fails with a mismatched shape error in the scales. I suspect it's due to the heterogenous quant not being parsed properly. See e.g. https://huggingface.co/mlx-community/Qwen3-1.7B-4bit-AWQ/blob/main/config.json#L20

davidkoski:
- support per layer quant config

### DIFF
--- a/Libraries/Embedders/BaseConfiguration.swift
+++ b/Libraries/Embedders/BaseConfiguration.swift
@@ -1,0 +1,156 @@
+// Copyright © 2025 Apple Inc.
+
+import Foundation
+
+/// Base ``LanguageModel`` configuration -- provides `modelType`
+/// and `quantization` (used in loading the model).
+///
+/// This is used by ``ModelFactory/load(hub:configuration:progressHandler:)``
+/// to determine the type of model to load.
+public struct BaseConfiguration: Codable, Sendable {
+    public let modelType: String
+
+    public struct Quantization: Codable, Sendable, Equatable {
+        public init(groupSize: Int, bits: Int) {
+            self.groupSize = groupSize
+            self.bits = bits
+        }
+
+        public let groupSize: Int
+        public let bits: Int
+
+        public var asTuple: (Int, Int) { (groupSize, bits) }
+
+        enum CodingKeys: String, CodingKey {
+            case groupSize = "group_size"
+            case bits = "bits"
+        }
+    }
+
+    /// handling instructions for ``PerLayerQuantization``
+    public enum QuantizationOption: Sendable {
+        case skip
+        case quantize(Quantization)
+    }
+
+    /// Per-layer ``Quantization`` values with optional default.
+    public struct PerLayerQuantization: Sendable {
+        public var quantization: Quantization? = nil
+        public var perLayerQuantization: [String: QuantizationOption]
+
+        public init(
+            quantization: BaseConfiguration.Quantization? = nil,
+            perLayerQuantization: [String: BaseConfiguration.QuantizationOption]
+        ) {
+            self.quantization = quantization
+            self.perLayerQuantization = perLayerQuantization
+        }
+
+        /// The quantization to apply for the given layer name or nil for no quantization.
+        public func quantization(layer: String) -> Quantization? {
+            if let perLayer = perLayerQuantization[layer] {
+                switch perLayer {
+                case .skip:
+                    return nil
+                case .quantize(let quantization):
+                    return quantization
+                }
+            } else {
+                return quantization
+            }
+        }
+    }
+
+    /// Special codable to support a mixed key: Int / key: Quantization
+    /// structure for hereogenous quantization, e.g.
+    ///
+    /// ```
+    /// "quantization": {
+    ///     "group_size": 64,
+    ///     "bits": 4,
+    ///     "model.embed_tokens": {
+    ///         "group_size": 32,
+    ///         "bits": 4
+    ///     },
+    ///     "model.layers.0.self_attn.q_norm": false,
+    /// ```
+    ///
+    /// This mixed type structure requires manual decoding.
+    struct QuantizationContainer: Codable, Sendable {
+        var quantization: Quantization
+        var perLayerQuantization: PerLayerQuantization
+
+        // based on Dictionary's coding key
+        internal struct _DictionaryCodingKey: CodingKey {
+            internal let stringValue: String
+            internal let intValue: Int?
+
+            internal init(stringValue: String) {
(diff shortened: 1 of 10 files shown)


Libraries/Embedders/BaseConfiguration.swift
```diff
@@ -0,0 +1,155 @@
+// Copyright © 2025 Apple Inc.
+
+import Foundation
+
+/// Base ``LanguageModel`` configuration -- provides `modelType`
+/// and `quantization` (used in loading the model).
+///
+/// This is used by ``ModelFactory/load(hub:configuration:progressHandler:)``
+/// to determine the type of model to load.
+public struct BaseConfiguration: Codable, Sendable {
```
davidkoski, line 10: The embedders code has the same quant config.  Described below.

I wonder if we need an MLXCommon (above MLXLMCommon)?

Libraries/MLXLMCommon/BaseConfiguration.swift
```diff
@@ -0,0 +1,155 @@
+// Copyright © 2025 Apple Inc.
+
+import Foundation
+
+/// Base ``LanguageModel`` configuration -- provides `modelType`
+/// and `quantization` (used in loading the model).
+///
+/// This is used by ``ModelFactory/load(hub:configuration:progressHandler:)``
+/// to determine the type of model to load.
+public struct BaseConfiguration: Codable, Sendable {
```
davidkoski, line 10: Extracted this from LanguageModel.swift as it is big enough to be its own thing.

Libraries/MLXLMCommon/BaseConfiguration.swift
```diff
@@ -0,0 +1,155 @@
...
+    ///     "group_size": 64,
+    ///     "bits": 4,
+    ///     "model.embed_tokens": {
+    ///         "group_size": 32,
+    ///         "bits": 4
+    ///     },
+    ///     "model.layers.0.self_attn.q_norm": false,
+    /// ```
+    ///
+    /// This mixed type structure requires manual decoding.
```
davidkoski, line 78: This turned out to be a bit tricky.  We have:

- quantization ints
- dictionary keys -> quantization structs
- dictionary keys -> bools

all mixed in the one dictionary.  We could treat it as an unstructured bag (like Python) but that isn't how we usually play it in swift.  We can handle it with a custom Codable implementation

Libraries/Embedders/BaseConfiguration.swift
```diff
@@ -0,0 +1,156 @@
...
+                case .skip:
+                    try container.encode(false, forKey: .init(stringValue: key))
+                case .quantize(let q):
+                    try container.encode(q, forKey: .init(stringValue: key))
+                }
+            }
+        }
+    }
+
+    var quanitzationContainer: QuantizationContainer?
```
davidkoski, line 141: internal bits -- this is the combined dictionary

Libraries/MLXLMCommon/BaseConfiguration.swift
```diff
@@ -0,0 +1,156 @@
...
+                case .skip:
+                    try container.encode(false, forKey: .init(stringValue: key))
+                case .quantize(let q):
+                    try container.encode(q, forKey: .init(stringValue: key))
+                }
+            }
+        }
+    }
+
+    var quanitzationContainer: QuantizationContainer?
```
davidkoski, line 141: internal bits -- this is the combined dictionary

Libraries/MLXLMCommon/BaseConfiguration.swift
```diff
@@ -0,0 +1,156 @@
...
+
+    @available(*, deprecated, message: "Please use perLayerQuantization instead")
+    public var quantization: Quantization? {
```
davidkoski, line 144: the single quantization value -- shouldn't use this

Libraries/MLXLMCommon/Load.swift
```diff
@@ -78,10 +80,17 @@ public func loadWeights(
...
     // quantize if needed
-    if let quantization {
-        quantize(model: model, groupSize: quantization.groupSize, bits: quantization.bits) {
-            path, module in
-            weights["\(path).scales"] != nil
+    if quantization != nil || perLayerQuantization != nil {
+        quantize(model: model) { path, module in
+            if weights["\(path).scales"] != nil {
+                if let perLayerQuantization {
+                    return perLayerQuantization.quantization(layer: path)?.asTuple
```
davidkoski, line 87: Use the per layer values (this allows `false` to mean no quant and falls back to the default if an unknown key)

Libraries/MLXLMCommon/Load.swift
```diff
@@ -91,3 +100,26 @@ public func loadWeights(
 
     eval(model)
 }
+
+// TODO remove once mlx-swift update is adopted
+func quantize(
```
davidkoski, line 105: See https://github.com/ml-explore/mlx-swift/pull/229

Applications/LLMEval/ContentView.swift
```diff
@@ -40,7 +40,8 @@ struct ContentView: View {
                     .frame(maxWidth: 350, alignment: .leading)
                     Toggle(isOn: $llm.enableThinking) {
                         Text("Thinking")
-                            .help("Switches between thinking and non-thinking modes. Support: Qwen3")
+                            .help(
+                                "Switches between thinking and non-thinking modes. Support: Qwen3")
```
davidkoski, line 44: I am a little confused here -- #290 fixed this (though it seems not to have run CI either).  swift-format complained about this code and I applied it again, _but_ `main` actually has this code:

https://github.com/ml-explore/mlx-swift-examples/blob/main/Applications/LLMEval/ContentView.swift#L43

The `-` line doesn't match main.

davidkoski: Haha, I should have realized.  Branched off the wrong point in main.

Libraries/Embedders/BaseConfiguration.swift
awni: Possibly. Though I don't know of any packages which produces mixed quant embedding models. So it's also fine to leave this out. Or we can support it.. since you have it already in case it does come up.

awni (approved):
Looks great!! Thanks for adding that!

PR #298, opened 2025-04-30
Fix UserInput init(messages:images:videos:tools:additionalContext:) broken for VLM

adrgrondin:
Fix `UserInput` `init(messages:images:videos:tools:additionalContext:)` not setting the images and videos arrays resulting in VLM not able to analyze images or videos when using this init.
The bug was introduced in (#257)


davidkoski (approved):
Good catch, thank you!

PR #301, opened 2025-05-01
swift-format

davidkoski:
another case where swift-format / CI was missed -- I know it ran, I can see the failure:

https://app.circleci.com/pipelines/github/ml-explore/mlx-swift-examples/726/workflows/a30af41a-86b8-4738-9b6d-c5372faabecd/jobs/1233

but I am not sure how I was able to merge it with the failure.


awni (approved):
🤷  thanks for the fix!

davidkoski:
> Oopsie, will run `swift-format` next time!

CI should catch it and the PR rules should prevent it from being merged 🤷‍♂️ 

Ah well, that is why CI runs on `main` too :-)

PR #303, opened 2025-05-01
Fix #297 - Basic markdown support in MLXChatExample

### LINKED ISSUE #297: Add markdown rendering to MLX Chat Example
Add markdown rendering to MLX Chat Example:

Image

The MLXLLMEval has an integration with swift-markdown-ui -- perhaps we could use that here.

Kawba:
PR as a suggestion for #297, as I've been doing something similar in my app.

If you just want simple markdown (i.e not the Github flavour) the SwiftUI text view supports this out of the box if you pass in the content as a LocalizedStringKey.

This might not fit everyone's needs, but for the way you are using it in this sample app, it looks fine.

image

One drawback to this (and using the SwiftUI "Text" view in general) is that you don't get to select individual characters/words on iOS or iPadOS (Just all of the content within that view in one go). If you want that, it might be worth looking at something like SelectableText (https://github.com/kevinhermawan/SelectableText) mixed with formatting the content beforehand with something like MarkdownToAttributedString (https://github.com/madebywindmill/MarkdownToAttributedString)

Anyway, I feel like for this use case, what is proposed in this PR is simple and should work. If you want it formatted differently or you want me to change the data type of "message.content" so it is not set in the views, im happy to make these changes.  

Thanks,
Kawba

### DIFF
--- a/Applications/MLXChatExample/Views/MessageView.swift
+++ b/Applications/MLXChatExample/Views/MessageView.swift
@@ -47,8 +47,9 @@ struct MessageView: View {
                             .clipShape(.rect(cornerRadius: 12))
                     }
 
-                    // Message content with tinted background
-                    Text(message.content)
+                    // Message content with tinted background.
+                    // LocalizedStringKey used to trigger default handling of markdown content.
+                    Text(LocalizedStringKey(message.content))
                         .padding(.vertical, 8)
                         .padding(.horizontal, 12)
                         .background(.tint, in: .rect(cornerRadius: 16))
@@ -58,8 +59,9 @@ struct MessageView: View {
 
         case .assistant:
             // Assistant messages are left-aligned without background
+            // LocalizedStringKey used to trigger default handling of markdown content.
             HStack {
-                Text(message.content)
+                Text(LocalizedStringKey(message.content))
                     .textSelection(.enabled)
 
                 Spacer()
--- a/Applications/MLXChatExample/README.md
+++ b/Applications/MLXChatExample/README.md
@@ -73,6 +73,28 @@ The codebase is thoroughly documented with:
 - Clear inline comments explaining complex logic
 - DocC documentation format
 
+### Markdown Support
+
+This sample app renders markdown content using SwiftUI's native `Text` view by passing the content as a `LocalizedStringKey`:
+
+```swift
+Text(LocalizedStringKey(message.content))
+```
+
+#### Limitations and Alternatives
+
+The default SwiftUI markdown rendering only supports standard markdown syntax. It does not support advanced features like tables and task lists that are available in GitHub Flavored Markdown (GFM).
+
+For more comprehensive markdown support:
+
+- **GitHub Flavored Markdown**: Consider using the [swift-markdown-ui](https://github.com/gonzalezreal/swift-markdown-ui) library. However, be aware that this library currently has an [unresolved issue with text selection](https://github.com/gonzalezreal/swift-markdown-ui/issues/264), which is why it wasn't used in this example.
+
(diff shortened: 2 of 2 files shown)


Applications/MLXChatExample/Views/MessageView.swift
```diff
@@ -47,8 +47,9 @@ struct MessageView: View {
                             .clipShape(.rect(cornerRadius: 12))
                     }
 
-                    // Message content with tinted background
-                    Text(message.content)
+                    // Message content with tinted background.
+                    // LocalizedStringKey used to trigger default handling of markdown content.
+                    Text(LocalizedStringKey(message.content))
```
davidkoski, line 52: I wonder if this is more directly done by just using AttributedString?

- https://developer.apple.com/documentation/foundation/attributedstring/init(markdown:options:baseurl:)-52n3u

Kawba: You're correct, I think I got confused as was playing around with LocalizedStringKey in my app. I'll update the PR.

Kawba: After having a play with it, I'm not sure that directly using AttributedString is a better way to go **for this example**. For me, it comes down to two things, both of which are just personal preference:

- In the constructor, you have to explicitly state that it is markdown (using the markdown: parameter) for it to work, which doesn't _feel_ right when you are not sure whether it is or not.
- The constructor for AttributedString throws where whereas the one for LocalizedStringKey does not. Not in itself an issue and probably a more desired way to go if you want to handle the specific exceptions it throws, but I feel like maybe that is making the example more complicated / harder for people to understand, for not a lot of gain. Again, not much of an issue if we just write the code once and change the data type to pass around AttributedString, but not sure it adds much value to developers looking at the project, as other markdown may be better suited to their specific needs anyway.

On balance, I feel like the current PR Changes add this support in a simple way, keeping the reasoning not hidden away like it might be if the data type is centralised and is easy to understand (with comments), so that others can decide on it later themselves. One thing that might be worth doing if we keep it in its current form is to document **why** in the comments to ensure anybody who looks at it is aware.

Again, happy to implement the changes to implement AttributedString if that's the direction you want to go, but I wanted to share my thoughts after trying it out first.

ibrahimcetin: I believe using `swift-markdown-ui` is the simplest and most appropriate way to accomplish this. It’s already included in the project’s dependency and is used in the LLMEval example. You can directly use it. I also believe this will enhance the clarity and understanding of the example.

Kawba: As discussed in https://github.com/ml-explore/mlx-swift-examples/issues/297, I'm happy to implement it in this way if there is consensus. 

I do want to note, however, the broken text selection in 'swift-markdown-ui' is a major issue that, in my opinion, renders it currently unsuitable for most applications, especially ones where users may want to use outputted content. You can read more about the issue here: https://github.com/gonzalezreal/swift-markdown-ui/issues/264

Kawba: As mentioned in the initial comment, I think the current best "experience for users" is a combination of using  SelectableText (https://github.com/kevinhermawan/SelectableText) mixed with formatting the content beforehand with something like MarkdownToAttributedString (https://github.com/madebywindmill/MarkdownToAttributedString).

This is still less than ideal as the Markdown support is only for standard markdown (i.e. no tables or task lists), but at least the user will be able to interact with content in a way that is consistent with how they use other applications.

That said, I don’t think we should include this setup in the example either, since the overall experience remains suboptimal. For now, sticking with the default Swift-formatted text view (as shown in this PR), along with a note in the issue to inform users of their options, seems like the best approach. This way, we avoid steering users toward a specific solution that may not fit their needs, while still giving them the information to make their own choice.

davidkoski (approved):
Looks great, thank you for this!

PR #307, opened 2025-05-02
Update READMEs of MLXLLM and MLXVLM to Use Current Registry Class Names

rudrankriyam:
This PR updates the README files for the MLXLLM and MLXVLM libraries to reflect the current class names used for model and type registries!


davidkoski (approved):
Looks great!

Issue #308, opened 2025-05-02
Proposal: Structured Tool API


ibrahimcetin:
The current `ToolSpec` is just a type alias and isn’t very ergonomic or type-safe when defining tools. I’d like to propose a more structured and Swifty API for defining and using tools, inspired by [ollama-swift](https://github.com/loopwork-ai/ollama-swift).

This new design makes it easier to define, register, and invoke tools with full Codable support for inputs and outputs. It also encourages safe and readable usage by reducing boilerplate and typos.

## Example Usage

```swift
struct WeatherInput: Codable {
    let city: String
}

struct WeatherOutput: Codable {
    let temperature: Double
    let conditions: String
}

let weatherTool = Tool<WeatherInput, WeatherOutput>(
    name: "get_current_weather",
    description: """
    Get the current weather for a city,
    with conditions ("sunny", "cloudy", etc.)
    and temperature in °C.
    """,
    parameters: [
       .required("city", type: .string, description: "The city to get weather for")
    ]
) { input async throws -> WeatherOutput in
    // Implement weather lookup logic here
    return WeatherOutput(temperature: 18.5, conditions: "cloudy")
}
```
The `Tool` struct accepts a `name`, `description`, a list of `parameters`, and the actual `implementation` logic. It is generic over two types: one for the input and one for the output, both conforming to `Codable`. Internally, the parameters will eventually be represented as a dictionary. The initializer is just a more structured and type-safe way to define them.

```swift
let chat: [Chat.Message] = [
    .system("You are a helpful assistant that can check the weather."),
    .user("What's the weather like in Portland?")
]

let input = UserInput(
    chat: chat,
    tools: [weatherTool]
)

/* Generate output */

/* Handle tool call */
```

## Handling Tool Calls

Ollama includes tool calls as part of the final message output. From my experience, when tools are provided, only a single message is returned, even if streaming is enabled. It holds back the output until the full response is generated, and only then returns the message along with any tool call metadata. While I'm not entirely sure how we could replicate this behavior, I think it's a practical and user-friendly approach and we should consider adopting a similar solution.
```swift
// ⚠️ Ollama Swift's Tool Call Handling
if let toolCalls = response.message.toolCalls {
    for toolCall in toolCalls {
        print("Tool called: \(toolCall.function.name)")
        print("Arguments: \(toolCall.function.arguments)")
    }
}
```

I’d like to discuss this with you. What are your thoughts?

davidkoski:
[...]

PR #311, opened 2025-05-04
Implement Structured Tool

ibrahimcetin:
This PR introduces a structured Tool type, addressing [#308](https://github.com/ml-explore/mlx-swift-examples/issues/308).

The implementation wraps `ToolSpec` ([String: Any]) inside a more type-safe Tool struct. I’ve added a basic unit test and refactored the `LLMEval` project to use the new `Tool` API.

I’m opening this PR early for discussion and feedback. Please feel free to test the changes and share thoughts.

Next steps for this PR:

- Add tool call handling
- Add `tool` message role
- Include an example app demonstrating multiple tools
- Expand unit test coverage
- Add documentation

This is an initial implementation and may change. Please feel free to share any feedback


Libraries/MLXLMCommon/Tool.swift
```diff
@@ -0,0 +1,138 @@
...
+
+/// Protocol defining the requirements for a tool.
+public protocol ToolProtocol: Sendable {
+    /// The JSON Schema describing the tool's interface.
+    var schema: ToolSpec { get }
+}
+
+public struct Tool<Input: Codable, Output: Codable>: ToolProtocol {
+    public let schema: ToolSpec
+    public let handler: (Input) async throws -> Output
```
davidkoski, line 42: Does this need to be `@Sendable`?

Applications/LLMEval/ContentView.swift
```diff
@@ -189,28 +189,25 @@ class LLMEvaluator {
...
+                "unit",
+                type: .string,
+                description: "The unit of temperature",
+                extraProperties: [
+                    "enum": ["celsius", "fahrenheit"]
+                ]
+            ),
+        ]
+    ) { input in
+        WeatherOutput(temperature: 14.0, conditions: "Sunny")
```
davidkoski, line 209: Perhaps it needs a comment saying it just simulates the call -- I was trying to figure out how this works :-)

Libraries/MLXLMCommon/Tool.swift
```diff
@@ -0,0 +1,138 @@
+// Copyright © 2025 Apple Inc.
+
+import Foundation
+
+public typealias ToolSpec = [String: Any] /* ? Should we change it to `any (Codable & Sendable)` */
```
davidkoski, line 5: Should we have this at all? `swift-transformers` defines:

```swift
public typealias ToolSpec = [String: Any]
```

maybe we should just stick with that?  Sendable is probably wise, but I don't know that Codable is required -- it is whatever can be transferred to the jinja template.

Libraries/MLXLMCommon/Tool.swift
```diff
@@ -0,0 +1,138 @@
...
+
+public indirect enum ToolParameterType {
+    case null
+    case bool
+    case int
+    case double
+    case string
+    case data
+    case array(ToolParameterType)
+    case object([String: ToolParameterType])
```
davidkoski, line 15: Should this be `dictionary`?

Applications/LLMEval/ContentView.swift
```diff
@@ -252,7 +249,10 @@ class LLMEvaluator {
             .user(prompt),
         ]
         let userInput = UserInput(
-            chat: chat, additionalContext: ["enable_thinking": enableThinking])
+            chat: chat,
+            tools: includeWeatherTool ? [currentWeatherTool.schema] : nil,
```
davidkoski, line 253: Should `UserInput` take a `Tool` directly?  Hrm, that may require erasure on the tool...

davidkoski:
This looks really good.  I am interested to see how it hooks back to actually call the tool -- we pass the schemas in to the tokenizer but somehow we would have to hold on to the tools during the evaluation and make the calls (and I presume inject the output into the token stream?)

ibrahimcetin:
@davidkoski, thanks for the feedback!

I'm currently working on tool call parsing and handling. The current implementation pauses generation when it detects a `<tool_call> `tag, and attempts to parse it into a `ToolCall` struct. This is handled by a new `ToolCallProcessor` which monitors the stream for a `<` character. If found, it starts buffering chunks until it reaches a `</tool_call>` tag. If the collected `toolCallBuffer` forms a valid tool call, it decodes it into a `ToolCall` and yields it as `Generation.toolCall`.

I've also introduced a new `tool` role to the `Message` model to represent tool call results.

For now, I’ve integrated a few basic tools into the `LLMEval` app to demonstrate the functionality: a weather tool, an add tool, and a time tool. (You can try them out, they are functional tools (the weather tool simply returns random weather info)).  These are temporary. I'm planning to remove them and instead create a dedicated example app to showcase tool usage more clearly.

This PR is still a work in progress. The current implementation reflects my initial approach, and I’m open to rethinking parts of it based on feedback. I haven’t tested all edge cases yet, so there may still be issues to resolve. I’ll continue refining it over the week.

Let me know what you think!

Applications/LLMEval/ContentView.swift
ibrahimcetin: I agree and I preferred this way just for now. I’m planning to add that `init`.

davidkoski (approved):
I think this is a great addition, I can't wait to see how people use it

PR #312, opened 2025-05-05
Feature: prompt caching (Fixes #310)

### LINKED ISSUE #310: Feature: Prompt cache
Currently `MLXLMCommon` has some basic support for a cache, however it isn't persisted across calls to `generate()`.

Even though it appears there could be a way to pass a `KVCache` to `generate()`, it ultimately must pass through the Sendable boundary if the app is to manage the cache. This isn't possible as `MLXArray` is not Sendable and also isn't desirable or necessary.

A prompt cache could be managed by the `ModelContainer` actor and stored in its context `ModelContext.promptCache`. Note that the prompt cache is an array of `KVCache`. In `mlx_lm` the `PromptCache` object also stores the token ids of the cached prompt and the model key to check if the model has changed. 

We could implement a similar struct:

```swift
public struct PromptCache {
    public let cache: [KVCache]
    public let modelKey: String
    public let tokens: MLXArray
}
```

The `PromptCache` struct could also have functions for trimming.

Functions analogous to `mlx_lm`'s `get_prompt_cache` could go in the `ModelContainer` actor.

I'm currently having a go at implementing this. Interested in any suggestions on the best approach.

jolonf:
Fixes: https://github.com/ml-explore/mlx-swift-examples/issues/310

Currently there is no way to persist the cache between calls to `generate()`. 

This is a relatively simple fix by adding `[KVCache]` parameters to the `generate()` functions which are then passed to the `TokenIterator`.

Trim functions have been added to the `KVCache` protocol and implemented in `KVCacheSimple`. Even though not strictly necessary for caching, it is not uncommon for the new prompt to be partially inconsistent with the cache either through tokenizer inconsistencies or recent messages being intentionally manipulated (e.g. removing a `<think>` block).

An example of how to implement a prompt cache has been added to MLXChatExample. The time to first token (TTFT) is now also displayed which is helpful to see the performance improvement from caching.

The prompt cache is implemented in the `PromptCache` class which is `@unchecked Sendable` which allows it to be used within the `ModelContainer` context. Currently there is no isolation on the `KVCache` in `PromptCache`.

Note that if using the `AsyncStream` versions of `generate()` there is no way to return token ids, so the newly generated response can't be added to the cache, and it will be reprocessed again on the next message. Perhaps the token could be added to the `Generation.chunk`?


Applications/MLXChatExample/Services/MLXService.swift
```diff
@@ -111,12 +119,51 @@ class MLXService {
...
+            // Create cache if it doesn't exist yet
+            cache = PromptCache(cache: context.model.newCache(parameters: parameters))
+            self.promptCache.setObject(cache, forKey: modelName as NSString)
         }
+
+        let lmInput: LMInput
+
+        /// Remove prefix from prompt that is already in cache
+        if let suffix = cache.getUncachedSuffix(prompt: fullPrompt.text.tokens) {
+            lmInput = LMInput(text: LMInput.Text(tokens: suffix))
```
davidkoski, line 158: I thought that the KVCache and the prompt should match in size -- that is I thought the prompt should not have the pieces that are already in the KVCache trimmed off.    Hrm, perhaps I am confused, here is the LLM prefill code:

```swift
        while y.tokens.size > prefillStepSize {
            let input = y[.newAxis, ..<prefillStepSize]
            let result = self(input, cache: cache.isEmpty ? nil : cache, state: state)
            eval(cache)
            y = y[prefillStepSize...]
        }
```

and I think it matches what you are doing here.

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -529,14 +529,15 @@ public func generate(
 ///   - input: prepared language model input
 ///   - parameters: parameters controlling the token generation
 ///   - context: model context (model and tokenizer)
+///   - cache: KV cache from previous output
 ///   - didGenerate: token visitor that can output tokens as they are generated and indicate early stop
 /// - Returns: the generated output
 public func generate(
-    input: LMInput, parameters: GenerateParameters, context: ModelContext,
+    input: LMInput, parameters: GenerateParameters, context: ModelContext, cache: [KVCache]? = nil,
```
davidkoski, line 536: I think some of this is already in place now.

davidkoski:
Overall I like this direction.  I think it needs:

- finish the borrowing of the KVCache (implement a visitor that holds a lock so we can satisfy the `unchecked Sendable`)
- remove the debug printing

Applications/MLXChatExample/Services/MLXService.swift
jolonf: Looking at that code, it looks to me like it is just passing the entire prompt (which is whatever was passed to it) through the model, using the existing cache.

To be able to do trimming the cache or the model would need to have a record of all of the tokens in the cache up to that point, I'm not sure it does? But if it did, the trimming logic could be moved there.

One issue we have at the moment with PromptCache is that it is responsible for recording the tokens represented by the cache. However, because AsyncStream<Generation> doesn't return tokens we have no way of updating the token list for the cache with the generated response. As a result we always trim the previous response from the new prompt because it doesn't know it is in the cache.

Either AsyncStream<Generation> should return the tokens (which may not be a bad idea anyway), or the cache moved closer to where the tokens are generated and they can be added there. However KVCache doesn't have a record of the tokens (I don't think?) so that is why we need PromptCache with its tokens MLXArray, so if the cache was to be fully managed within TokenIterator we would need to pass it the full PromptCache.

Applications/MLXChatExample/Models/PromptCache.swift
```diff
@@ -0,0 +1,124 @@
...
+import MLX
+import MLXLMCommon
+
+/// Stores the KV Cache between calls to ``generate`` and maintains
+/// the token ids reflected in the cache.
+///
+/// ``PromptCache`` is ``@unchecked Sendable`` which allows it
+/// to be used within the ``ModelContainer`` context.
+///
+/// TODO: cache isolation
```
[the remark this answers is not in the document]
jolonf: I agree. I've forgotten some of the details but I had roadblocks with each approach. I think there was an issue with ModelContainer.perform() being asynchronous and trying to wrap that with something like withCache.

I might have to leave it to someone with more expertise in Swift concurrency.

davidkoski:
What do you think of this with regard to #330?  That encapsulates the KVCache which is part of this PR, but I think the cache manipulation is still key.

Applications/MLXChatExample/Models/PromptCache.swift
```diff
@@ -0,0 +1,124 @@
...
+    ///         - Attempt to trim the cache to the common prefix
+    ///         - Return suffix of prompt not in cache
+    ///         - If the cache is not trimmable return nil for the caller
+    ///             to create a new cache.
+    public func getUncachedSuffix(prompt: MLXArray) -> MLXArray? {
+
+        print("[getUncachedSuffix] self.tokens.size = \(self.tokens.size)")
+
+        print("cache[\(self.tokens.size)]: \(self.tokens)")
+        print("prompt[\(prompt.size)]: \(prompt)")
```
davidkoski, line 51: - [ ] remove debug printing

davidkoski:
Reference on the `<think>` tags: https://api-docs.deepseek.com/guides/reasoning_model

I think this is a little bit complicated, but goes something like this:

- iterator keeps track of start index when doing generation
- produces full output
- goes back through tokens and filters out the `<think>` section(s)
- resets the KVCache to the start index
- prefills the KVCache with the response (like a prompt)

We would have to consider what happens if the caller terminates the iteration early.  Maybe it isn't even part of the Iterator (ideally we could compose this).

Anyway, the `<think>` section will be in the KVCache as a consequence of generation so we have to replay the output to replace what is in there.

jolonf:
Something to consider if the `<think>` blocks are to be removed at the token level is that the end `<\think>` tag could form a token with the following response. For example if the response looked like:

```
<think>
Thinking... thinking...
</think>
Here is my response
```
It is possible that a token could be `>Here`, which wouldn't allow the `<think>` block to be properly removed at the token level.

If the `<think>` block could be removed at the token level an even better optimisation would be to snip the `<think>` block out, not just trim. But I don't know enough about LLMs and KV caches to know if this would work.

If we can't edit the cache at the token level, then the entire previous response has to be removed (as it is guaranteed to be on a clean token boundary), and it is pre-filled again as if a new prompt.

Note that the `KVCache` trimming is extremely efficient, it just edits the offset in the current implementation in the PR:

```swift
    public func trim(n: Int) -> Int {
        let toTrim = min(self.offset, n)
        self.offset -= toTrim
        return toTrim
    }
```

davidkoski:
Yeah, we probably have to handle the `<think>` tags in decoded token space, trim it out, then re-tokenize.

FWIW mlx-lm (python side) does not have this capability yet.

PR #314, opened 2025-05-13
consume security scoped bookmarks to read urls

davidkoski:
FYI @ibrahimcetin


Applications/MLXChatExample/Services/MLXService.swift
```diff
@@ -107,7 +107,8 @@ class MLXService {
         }
 
         // Prepare input for model processing
-        let userInput = UserInput(chat: chat)
+        let userInput = UserInput(
+            chat: chat, processing: .init(resize: .init(width: 1024, height: 1024)))
```
davidkoski, line 111: I was running into mysterious failures from loading 12MP images -- constraint it to something reasonable.  I think the models probably ought to be doing this, and in some cases they are, but the limits are too high.

Applications/MLXChatExample/ViewModels/ChatViewModel.swift
```diff
@@ -160,15 +160,29 @@ class MediaSelection {
...
     /// Whether any media is currently selected
     var isEmpty: Bool {
         images.isEmpty && videos.isEmpty
     }
+
+    private func didSetURLs(_ old: [URL], _ new: [URL]) {
+        // the urls we get from fileImporter require SSB calls to access
+        new.filter { !old.contains($0) }.forEach { _ = $0.startAccessingSecurityScopedResource() }
+        old.filter { !new.contains($0) }.forEach { $0.stopAccessingSecurityScopedResource() }
+    }
```
davidkoski, line 185: From the docs on `fileImporter`:

```
    ///   - onCompletion: A callback that will be invoked when the operation has
    ///     succeeded or failed. To access the received URLs, call `startAccessingSecurityScopedResource`.
    ///     When the access is no longer required, call `stopAccessingSecurityScopedResource`.
```

Without this I see a spinning progress -- the image never loads.  I think this is a side effect of the entitlements changes

MLXChatExample.entitlements
```diff
@@ -8,6 +8,8 @@
 	<true/>
 	<key>com.apple.security.files.downloads.read-write</key>
 	<true/>
+	<key>com.apple.security.files.user-selected.read-only</key>
+	<true/>
```
davidkoski, line 12: Allow the open panel (`fileImporter`) to select files and get the SSB so we can read them

awni (approved):
LGTM! Thanks!!

PR #317, opened 2025-05-15
support for running all the models

davidkoski:
- add a --download argument
- add a `list` command
- support scripts to write a script to run all the models

Found these issues (which I will try to address here):

# Error: Key _freqs not found in SuScaledRotaryEmbedding -- see https://github.com/ml-explore/mlx-swift-examples/issues/218
./mlx-run llm-tool eval --download ~/Downloads/huggingface --model mlx-community/Phi-3.5-MoE-instruct-4bit

# Error: Runtime error: System role not supported (fixed)
./mlx-run llm-tool eval --download ~/Downloads/huggingface --model mlx-community/gemma-2-9b-it-4bit
./mlx-run llm-tool eval --download ~/Downloads/huggingface --model mlx-community/gemma-2-2b-it-4bit
./mlx-run llm-tool eval --download ~/Downloads/huggingface --model mlx-community/quantized-gemma-2b-it

# Error: Runtime error: Conversation roles must alternate user/assistant/user/assistant/...
./mlx-run llm-tool eval --download ~/Downloads/huggingface --model mlx-community/Mistral-7B-Instruct-v0.3-4bit

# Error: Key _positionIds not found in VisionEmbeddings -- see https://github.com/ml-explore/mlx-swift-examples/issues/218
./mlx-run llm-tool eval --download ~/Downloads/huggingface --model mlx-community/paligemma-3b-mix-448-8bit --resize 512 --image /tmp/test.jpg

# Error: Keys ["image_seq_len", "imageSequenceLength"] not found or type not match or collection contains null.  -- see https://github.com/ml-explore/mlx-swift-examples/issues/218
./mlx-run llm-tool eval --download ~/Downloads/huggingface --model HuggingFaceTB/SmolVLM2-500M-Video-Instruct-mlx --resize 512 --image /tmp/test.jpg

# gives a poor description -- #318
./mlx-run llm-tool eval --download ~/Downloads/huggingface --model mlx-community/SmolVLM-Instruct-4bit --resize 512 --image /tmp/test.jpg


Libraries/MLXLLM/Models/Gemma2.swift
```diff
@@ -246,7 +251,7 @@ public struct Gemma2Configuration: Codable {
         case queryPreAttnScalar = "query_pre_attn_scalar"
     }
 
-    public init(from decoder: Decoder) throws {
+    public init(from decoder: Swift.Decoder) throws {
```
davidkoski, line 254: @pcuenca -- I have run into this in a few places where the `Decoder` protocol from `swift-transformers` collides with the Swift `Decoder` protocol.  It can be worked around like this or with a typealias, but I wonder if renaming it in `swift-transformers` would make sense?

Libraries/MLXLLM/Models/Llama.swift
```diff
@@ -317,6 +318,23 @@ public class LlamaModel: Module, LLMModel, KVCacheDimensionProvider {
             !$0.key.contains("self_attn.rotary_emb.inv_freq")
         }
     }
+
+    public func messageGenerator(tokenizer: any Tokenizer) -> any MessageGenerator {
+        // some models allow the system role and some do not -- this is enforced
+        // by the chat template (code).
```
davidkoski, line 324: Not super thrilled with this but I don't see any other way to detect this as there is somewhat arbitrary code than runs in the chat template.

Libraries/MLXLMCommon/Chat.swift
```diff
@@ -63,12 +63,25 @@ public enum Chat {
 /// ```
 public protocol MessageGenerator {
 
+    /// Generates messages from the input.
+    func generate(from input: UserInput) -> [Message]
+
+    /// Returns array of `[String: Any]` aka ``Message``
+    func generate(messages: [Chat.Message]) -> [Message]
```
davidkoski, line 70: These need to move to the protocol so they get dynamic dispatch -- we need the `[Chat.Message]` case for filtering by role.

Libraries/MLXLLM/Models/Gemma2.swift
pcuenca: Sure, I'll see if we can rename.

support/run-all-llms.sh
```diff
@@ -0,0 +1,124 @@
+echo
```
awni, line 1: It might be worth putting a little header in this so it's clear it's auto-generated.

Libraries/MLXLLM/Models/Llama.swift
davidkoski: Yes, that is a possibility too, but it requires a way for a user to configure the translation of structured messages into the plain text that is consumed.  It isn't by model type, it is by specific model instance.

awni (approved):
LGTM!

PR #319, opened 2025-05-17
chore(ci): update to xcode 16.3

reneleonhardt:
### Updated
- Update CI from Xcode 16.0 (Swift 6.0) to 16.3 (Swift 6.1)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,7 +15,7 @@ jobs:
 
   mac_build_and_test:
     macos:
-      xcode: 16.0.0
+      xcode: 16.3.0
     resource_class: macos.m1.medium.gen1
     steps:
       - checkout
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2.1
 
 orbs:
   apple: ml-explore/pr-approval@0.1.0
-  
+
 parameters:
   nightly_build:
     type: boolean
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -15,7 +15,7 @@ jobs:
 
   mac_build_and_test:
     macos:
-      xcode: 16.0.0
+      xcode: 16.3.0
```
davidkoski, line 18: This seems fine but maybe we need both:  we should make sure it builds on the oldest Xcode we support as well.

reneleonhardt: In a GitHub workflow it would be easy to make a matrix, but I'm not familiar with CircleCI.
And I don't know your oldest supported Xcode version.
If this breaks your builds you can close it, just wanted to help you benefit from all Swift 6.1 advancements.

davidkoski: OK, I will figure out what we need here and propose an addition

reneleonhardt: Applied, thank you 🙂

reneleonhardt: I don't know what GitHub wants 😅

> Allowing edits by maintainers
> 
> If checked, users with write access to ml-explore/mlx-swift-examples can add new commits to your chore/update-xcode branch.

davidkoski (approved):
Let's do it!  Thanks!

Issue #221, opened 2025-03-04
Suggestion: Structured Output (for Tool Usage)


LeonNissen:
Hi Community,

First off, thank you all for your amazing work on MLX! We’ve been building on this framework in our project [SpeziLLM](https://github.com/StanfordSpezi/SpeziLLM) and its just awesome! 🚀

In our exploration of tool usage with MLX, I’ve been following the implementation by @DePasqualeOrg in ml-explore/mlx-swift-examples#174, and one challenge we encountered is that smaller models can sometimes deviate from the system prompt, which leads to text output instead of tool calls.

Some applications—like LM Studio—have introduced a [Structured Output API](https://lmstudio.ai/docs/app/api/structured-output) feature. This approach enforces a JSON Schema on the model’s output to ensure consistency. They're using [outlines](https://github.com/dottxt-ai/outlines) and, as far as I understand, iteratively matching allowed tokens (see https://github.com/dottxt-ai/outlines/blob/main/outlines/processors/structured.py).

I believe integrating a similar Structured Output feature into mlx-swift-examples could greatly improve reliability, especially when dealing with models that might stray from expected responses.

Any tips, ideas, or pointers from the community would be greatly appreciated!

DePasqualeOrg:
I haven't tried this yet, but one idea that comes to mind is modifying the prompt template to include the beginning of whatever output format you want to see in the model's response. For example, the DeepSeek R1 models have a prompt template that includes the `<think>` tag at the beginning of the model's response to ensure that the response includes a thinking block, which it otherwise might not.

davidkoski:
See #308 and #311

davidkoski:
Is the suggestion that the sampler change when entering the `<tool_call>` token?  You use a standard sampler most of the time but a special tool sampler inside?

deet:
Either

1. Change the sampler when entering a the `<tool_call>` block to a sampler that forces valid tool call output
2. Use a structured sampler for the whole output, and have the tool calls be only one branch of the sampler's logic.

The argument for #1 is potentially simplicity and speed while outside of the tool call section. However, there would need to be some higher level mode tracking for whether the LLM is currently outputting reasoning tokens, tool call tokens, or generic output.

The argument for #2 would be that the mode tracking is essentially the same as the state tracking within the sampler in the tool call. This would allow the sampler to enforce that the model outputs a single reasoning block, then a tool call, for example.

JoeySlomowitz:
[...]

PR #320, opened 2025-05-26
Convert GenerateParameters to a protocol

deet:
To allow customization of the `LogitSampler` and `LogitProcessor` behavior, this PR converts `GenerateParameters` from a struct to a protocol.

The motivation would be to more easily allow for some work on grammar-constrained output generation, as mentioned in https://github.com/ml-explore/mlx-swift-examples/issues/221#issuecomment-2910291469 

If merged, this PR is a breaking change for library consumers if they instantiate a `GenerateParameters` object in their own code. So while not as clean, this could be modified to preserve the `GenerateParameters` name and choose a new name for the protocol instead.

An alternative approach to doing this could be to allow setting override values for the sampler and processor that bypass the logic in the functions in this existing structure for picking these objects.

### DIFF
--- a/Applications/LLMEval/ContentView.swift
+++ b/Applications/LLMEval/ContentView.swift
@@ -176,7 +176,7 @@ class LLMEvaluator {
     let modelConfiguration = LLMRegistry.qwen3_1_7b_4bit
 
     /// parameters controlling the output
-    let generateParameters = GenerateParameters(maxTokens: 240, temperature: 0.6)
+    let generateParameters = DefaultGenerateParameters(maxTokens: 240, temperature: 0.6)
     let updateInterval = Duration.seconds(0.25)
 
     /// A task responsible for handling the generation process.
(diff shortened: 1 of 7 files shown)


davidkoski:
Would using this init on the `TokenIterator` be sufficient?

```swift
    /// Initialize a `TokenIterator` with the given input and logit handling.
    ///
    /// - Parameters:
    ///   - input: language model input
    ///   - model: the ``LanguageModel``
    ///   - cache: optional ``KVCache``
    ///   - processor: the logit processor
    ///   - sampler: the logit sampler
    ///   - prefillStepSize: optional prefill step size
    ///   - maxTokens: maximum number of tokens to generate
    public init(
        input: LMInput, model: any LanguageModel, cache: [KVCache]? = nil,
        processor: LogitProcessor?, sampler: LogitSampler, prefillStepSize: Int = 512,
        maxTokens: Int? = nil
    ) throws {
```

You can pass in your own processor and sampler.  If this needs to be exposed somewhere at a higher level, that can be done too.  I think `GenerateParamters` is really for the developer who wants the functionality but is not interested in these details.  This method allows full control.

deet:
That... makes sense! OK will investigate using the TokenIterator instead.

PR #313, opened 2025-05-12
adopt mlx-swift 0.25.2

davidkoski:
Changes for adopting mlx / mlx-swift 0.25.2

- #274 
- #273

PR #325, opened 2025-06-04
Fix Mutability Build Errors

JustinMeans:
The Embedders Package doesn't build due to the two let / var mutability errors. Changed 'let words' to 'var words'.


davidkoski (approved):
Thanks for this!

I was curious why this doesn't fail in CI and the answer is _it will_ when we pick up the newer mlx-swift.  See #313 

There is a bug fix here https://github.com/ml-explore/mlx-swift/commit/0e40fa03dfb57d310671e473b77adb86761d1915 that changes this behavior.

PR #326, opened 2025-06-04
Improve adapters usage (LoRA, DoRA)

petrukha-ivan:
## Motivation
First off all, thank you for the excellent work on this library! We use adapters in our product a lot, and our main use case is the single base model and multiple adapters for different scenarios. Currently, the `MLXLMCommon` library doesn't provide any way to dynamically switch between adapters. This Pull Request improves adapters usage with convenient methods to load and unload adapters.

## What
- Implemented `ModelAdapter` protocol and extensions to achieve syntax like `model.load(adapter: adapter)`
- Implemented `ModelAdapterFactory` class to load adapters from the Hugging Face Hub.
- Implemented `DoRALinear` and `QDoRALinear ` layers.

### Prerequisites:
I've prepared both LoRA and DoRA examples trained on the `wikisql` dataset from this repo:
- [ipetrukha/Llama-3.2-1B-Instruct-4bit-wikisql-lora](https://huggingface.co/ipetrukha/Llama-3.2-1B-Instruct-4bit-wikisql-lora)
- [ipetrukha/Llama-3.2-1B-Instruct-4bit-wikisql-dora](https://huggingface.co/ipetrukha/Llama-3.2-1B-Instruct-4bit-wikisql-dora)

### Examples
Initialize base model:
```swift
let modelConfiguration = ModelConfiguration(id: "mlx-community/Llama-3.2-1B-Instruct-4bit")
let modelContainer = try await LLMModelFactory.shared.loadContainer(hub: hub, configuration: modelConfiguration)
```

Initialize adapter:
```swift
let adapterConfiguration = ModelConfiguration(id: "ipetrukha/Llama-3.2-1B-Instruct-4bit-wikisql-dora")
let adapter = try await ModelAdapterFactory.shared.load(hub: hub, configuration: adapterConfiguration)
```

Use the basic generate function without any reference to adapters:
```swift
@Sendable
func generate(_ prompt: String, context: ModelContext) async throws -> String {
    
    let parameters = GenerateParameters()
    let input = try await context.processor.prepare(input: UserInput(prompt: prompt))
    let result = try MLXLMCommon.generate(input: input, parameters: parameters, context: context) { tokens in
        if tokens.last == context.tokenizer.eosTokenId || tokens.count > 256 {
            return .stop
        } else {
            return .more
        }
    }
    
    return result.output
}
```

Generate output with the adapter and compare it with the output generated without the adapter:
```swift
let prompt = """
table: 1-10007452-3
columns: Order Year, Manufacturer, Model, Fleet Series (Quantity), Powertrain (Engine/Transmission), Fuel Propulsion
Q: who is the manufacturer for the order year 1998?
"""

try await modelContainer.perform { context in
    
    // Generation with adapter
    try await context.model.perform(with: adapter) {
        let output = try await generate(prompt, context: context)
        print(output) // "SELECT Manufacturer FROM 1-10007452-3 WHERE Order Year = '1998'"
    }
    
    // Generation without adapter
    let output = try await generate(prompt, context: context)
    print(output) // "I cannot provide you with a list of companies that manufactured a car for the year 1998."
}
```

### Code duplication
I tried to reduce code duplication and combine DoRA layer to something like this:
```swift
public class DoRAUniversalLinear: Module {
    
    let adapts: Linear // Can be QuantizedLinear
    let scale: Float
    
    @ParameterInfo(key: "lora_a") var loraA: MLXArray
    @ParameterInfo(key: "lora_b") var loraB: MLXArray
    @ParameterInfo(key: "m") var magnitude: MLXArray
    
    ...
```
But seems it's not possible right now because of the `MLXNN` library limitations. Anyway, I decided to reduce repetitive code and put it in a private functions. This slightly improved the code readability, and if you like this approach, I could refactor LoRA layers in the same way; just let me know.

### Possible bug
During the DoRA layers implementation, I referenced the Python code [here](https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/tuner/dora.py), and the Swift version produces the same forward pass results and fused tensors. However, final results are slightly different from Python versions on tensors level. I checked deeper and found the reason in `Attention` class of `Llama` model.

`Llama.swift`
```swift
    queries: queries, keys: keys, values: values, scale: scale, mask: mask
)
.transposed(0, 2, 1, 3)
.reshaped(B, L, -1)
```

`llama.py`
```python
output = scaled_dot_product_attention(
    queries, keys, values, cache=cache, scale=self.scale, mask=mask
)

output = output.transpose(0, 2, 1, 3).reshape(B, L, -1)
```

This two code snippets produce different results on same inputs. More particularly the problem is inside `reshaped` function. If you check it on debugger with same inputs you would get this in Swift:
```
(lldb) po MLXFast.scaledDotProductAttention(queries: queries, keys: keys, values: values, scale: scale, mask: mask).mean()
array(3.36766e-05, dtype=float16)

(lldb) po MLXFast.scaledDotProductAttention(queries: queries, keys: keys, values: values, scale: scale, mask: mask).transposed(0, 2, 1, 3).reshaped(B, L, -1).mean()
array(3.35574e-05, dtype=float16)
```

And this in Python:
```
scaled_dot_product_attention(queries, keys, values, cache=cache, scale=self.scale, mask=mask).mean()
array(3.36766e-05, dtype=float16)

scaled_dot_product_attention(queries, keys, values, cache=cache, scale=self.scale, mask=mask).transpose(0, 2, 1, 3).reshape(B, L, -1).mean()
array(3.29018e-05, dtype=float16)
```

After reshaping they have different values inside. Despite this numerical variance, the DoRA behavior in Swift seems to match the Python version. The adapter works correctly and produces the expected results during generation.


Libraries/MLXLMCommon/Adapters/ModelAdapterFactory.swift
```diff
@@ -0,0 +1,65 @@
...
+        configuration: ModelConfiguration,
+        progressHandler: @Sendable @escaping (Progress) -> Void = { _ in }
+    ) async throws -> ModelAdapter {
+
+        struct BaseConfiguration: Decodable {
+
+            enum FineTuneType: String, Decodable {
+                case lora
+                case dora
+            }
```
davidkoski, line 36: If a developer wanted to add LoRA to an existing model would it be helpful to be able to pass this parameter in to the `load()` function?  Maybe a second entry point?

Libraries/MLXLMCommon/Adapters/ModelAdapterFactory.swift
```diff
@@ -0,0 +1,65 @@
...
+            hub: hub, configuration: configuration, progressHandler: progressHandler)
+        let configurationURL = snapshotURL.appending(path: "adapter_config").appendingPathExtension(
+            "json")
+        let configurationData = try Data(contentsOf: configurationURL)
+        let configuration = try decoder.decode(BaseConfiguration.self, from: configurationData)
+
+        let weightsURL = snapshotURL.appending(path: "adapters").appendingPathExtension(
+            "safetensors")
+        let weights = try MLX.loadArrays(url: weightsURL)
+        let parameters = ModuleParameters.unflattened(weights)
```
davidkoski, line 54: This only loads the adapter weights and produces a ModelAdapter -- this is something you use via composition with another model, correct?

Can the weights ever be quantized?  If so, I wonder if this should use `loadWeights()` (or a refactored version of that)?

Libraries/MLXLMCommon/Adapters/LoRA/DoRA+Layers.swift
```diff
@@ -0,0 +1,180 @@
...
+                weight: dequantizedWeight, loraA: loraA, loraB: loraB, scale: scale,
+                magnitude: magnitude),
+            bias: bias, groupSize: groupSize, bits: bits
+        )
+    }
+
+    public override func callAsFunction(_ x: MLXArray) -> MLXArray {
+        forward(
+            x: x,
+            weight: dequantizedWeight, bias: bias,
```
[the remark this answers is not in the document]
petrukha-ivan: I ran a performance test comparing the two approaches over 10,000 forward passes. The version using `quantizedMatmul` directly on the quantized weights showed a clear performance benefit:
```
QDoRALinear (quantizedWeights -> dequantizedWeights -> matmul) finished (18 secs, 340 ms)
QDoRALinear (quantizedWeights -> quantizedMatmul) finished (14 secs, 613 ms)
```
That’s about a 20% speedup, so it definitely seems worth switching to the more efficient path 👍

Libraries/MLXLMCommon/Adapters/ModelAdapterFactory.swift
davidkoski: That sounds reasonable -- if it comes up we can refactor or copy the quantization code here, but what you said about them being small makes sense.

davidkoski (approved):
Looks great, thank you!!

PR #333, opened 2025-06-12
Improve model config error handling

DePasqualeOrg:
While I was testing out different Gemma 3 models, I got a generic error message like "the data could not be found". The reason was that the model's `config.json` didn't include some of the properties that were expected when it was decoded. I've made the error messages for this case more descriptive.


davidkoski (approved):
Great improvement, thank you!

PR #334, opened 2025-06-13
Add missing KV cache functionality

DePasqualeOrg:
I've separated the KV cache part out of my [Gemma 3](https://github.com/ml-explore/mlx-swift-examples/pull/238) PR. I'm testing the quantized KV cache on Qwen 3. If we go with this approach, we'll need to update all the models to use `attentionWithCacheUpdate` as Qwen 3 does here.

The Python API is more elegant because it allows more dynamic typing. In Swift we need this wrapper function due to the requirements of the `KVCache` protocol.

You can test inference with and without the quantized KV cache with these arguments in llm-tool:

`--model mlx-community/Qwen3-1.7B-4bit --prompt "Explain quantum computing in simple terms" --max-tokens 100 --kv-bits 4`
`--model mlx-community/Qwen3-1.7B-4bit --prompt "Explain quantum computing in simple terms" --max-tokens 100`

For short sequences like this, using the quantized KV cache is actually slower, but it should be more efficient for much longer sequences.

---

Some preliminary test results: Qwen3-1.7B-4bit using 4-bit quantization for the KV cache results in the model getting stuck in repetitive loops. The same model with 8-bit quantization for the KV cache works well, but even at 1200 tokens it's still about 8% slower than when using the non-quantized KV cache. Maybe someone who's interested in using a quantized KV cache can do some more testing on other models with much longer sequence lengths (cc @mzbac).

### DIFF
--- a/Libraries/MLXLMCommon/KVCache.swift
+++ b/Libraries/MLXLMCommon/KVCache.swift
@@ -19,7 +19,7 @@ import MLXNN
 /// **Quantized cache:**
 /// ```swift
 /// let quantizedCache = QuantizedKVCache(groupSize: 64, bits: 4)
-/// let (qKeys, qValues) = quantizedCache.updateAndFetchQuantized(keys: keys, values: values)
+/// let (qKeys, qValues) = quantizedCache.updateQuantized(keys: keys, values: values)
 ///
 /// let output = quantizedScaledDotProductAttention(
 ///     queries: queries,
@@ -65,7 +65,7 @@ public protocol KVCache: Evaluatable {
 /// ```swift
 /// // Efficient quantized path
 /// if let quantizedCache = cache as? QuantizedKVCacheProtocol {
-///     let (qKeys, qValues) = quantizedCache.updateAndFetchQuantized(keys: k, values: v)
+///     let (qKeys, qValues) = quantizedCache.updateQuantized(keys: k, values: v)
 ///     // Use native quantized operations
 ///     let scores = quantizedMatmul(queries, w: qKeys.0, scales: qKeys.1, biases: qKeys.2, ...)
 /// } else {
@@ -298,7 +298,7 @@ public class KVCacheSimple: BaseKVCache, CustomDebugStringConvertible {
 
     /// Convert to quantized cache for maximum efficiency
     ///
-    /// Use `updateAndFetchQuantized()` and `quantizedScaledDotProductAttention()` for zero-overhead operation.
+    /// Use `updateQuantized()` and `quantizedScaledDotProductAttention()` for zero-overhead operation.
     public func toQuantized(groupSize: Int = 64, bits: Int = 4) -> QuantizedKVCache {
         let quantizedCache = QuantizedKVCache(groupSize: groupSize, bits: bits)
         quantizedCache.offset = self.offset
@@ -331,13 +331,13 @@ public class RotatingKVCache: BaseKVCache, CustomDebugStringConvertible {
     private var keep: Int
     private var keys: MLXArray?
     private var values: MLXArray?
-    private var maxCacheSize: Int?
+    private var maxCacheSize: Int
     private var step: Int
     private var idx: Int = 0
 
     public override var maxSize: Int? { maxCacheSize }
 
-    public init(maxSize: Int?, keep: Int = 0, step: Int = 256) {
+    public init(maxSize: Int, keep: Int = 0, step: Int = 256) {
         self.maxCacheSize = maxSize
         self.keep = keep
         self.step = step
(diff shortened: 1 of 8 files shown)


mzbac:
Thanks for the great work, I also noticed that the 4bit kv cache can cause some performance degradation, especially for the thinking model. I did some tests earlier for my implementation and found that memory savings become noticeable during token generation up to around 2,000 or 4,000 tokens. However, I didn't track the speed, will try to test your implementation once I get a chance

Libraries/MLXLMCommon/KVCache.swift
```diff
@@ -130,13 +250,1144 @@ public class KVCacheSimple: KVCache, Evaluatable, CustomDebugStringConvertible {
...
+// MARK: - Utility Functions
+
+/// Save a pre-computed prompt cache to a file.
+///
+/// - Parameters:
+///   - fileName: The `.safetensors` file name
+///   - cache: The model cache state
+///   - metadata: Optional metadata to save along with cache state
+public func savePromptCache(
+    fileName: String,
```
davidkoski, line 1002: I wonder if this should be URL?  e.g. if you are constructing a path to the caches directory this would normally be a url.

Libraries/MLXLMCommon/KVCache.swift
```diff
@@ -130,13 +250,1144 @@ public class KVCacheSimple: KVCache, Evaluatable, CustomDebugStringConvertible {
...
+        let keyBytes = Array(key.utf8)
+        let valueBytes = Array(value.utf8)
+        if !keyBytes.isEmpty {
+            flattenedData["__metadata_user_key_\(i)"] = MLXArray(keyBytes.map { Int32($0) })
+        }
+        if !valueBytes.isEmpty {
+            flattenedData["__metadata_user_value_\(i)"] = MLXArray(valueBytes.map { Int32($0) })
+        }
+    }
+    flattenedData["__metadata_user_count"] = MLXArray([metadata.count])
```
davidkoski, line 1076: Since the metadata comes back as `[String:String]` below, why not use: `public func saveToData(
    arrays: [String: MLXArray], metadata: [String: String] = [:])`?  It suports a metadata dictionary directly.

Libraries/MLXLMCommon/KVCache.swift
```diff
@@ -130,13 +250,1144 @@ public class KVCacheSimple: KVCache, Evaluatable, CustomDebugStringConvertible {
...
+
+/// Load a prompt cache from a file.
+///
+/// - Parameters:
+///   - fileName: The `.safetensors` file name
+///   - returnMetadata: Whether to return metadata along with the cache
+/// - Returns: The prompt cache and optionally the metadata
+public func loadPromptCache(
+    fileName: String,
+    returnMetadata: Bool = false
```
davidkoski, line 1089: Why would you not want to return this?  The return value always includes it (though it might be empty)

Libraries/MLXLMCommon/KVCache.swift
```diff
@@ -130,13 +250,1144 @@ public class KVCacheSimple: KVCache, Evaluatable, CustomDebugStringConvertible {
...
+/// Converts regular caches to quantized caches when:
+/// - kvBits is specified
+/// - The cache is not already quantized
+/// - The cache offset is greater than quantizedKVStart
+///
+/// - Parameters:
+///   - cache: Array of KV caches to potentially quantize
+///   - kvBits: Number of bits for quantization (nil = no quantization)
+///   - kvGroupSize: Group size for quantization
+///   - quantizedKVStart: Step to begin quantizing
```
davidkoski, line 1371: I am not familiar with the typical use of this -- the step is the token offset?  So if you have a long prompt you will switch to quantized before you start evaluating the response -- that is the intent?

Libraries/MLXLMCommon/KVCache.swift
```diff
@@ -130,13 +250,1144 @@ public class KVCacheSimple: KVCache, Evaluatable, CustomDebugStringConvertible {
...
+    else {
+        return
+    }
+
+    for i in 0 ..< cache.count {
+        if let simpleCache = cache[i] as? KVCacheSimple {
+            cache[i] = simpleCache.toQuantized(groupSize: kvGroupSize, bits: kvBits)
+        }
+        // Note: RotatingKVCache.toQuantized() is not implemented yet like in Python
+        // When implemented, add: else if let rotatingCache = cache[i] as? RotatingKVCache { ... }
```
davidkoski, line 1391: Since this may support rotating caches in the future I wonder if the name of the function should be more generic?  Less about quantization -- maybe something about a step did complete?

Is this something someone might want to customize?  Should this be a protocol or closure passed in to the iterator?  I guess we can decide that later if needed.

Libraries/MLXLMCommon/KVCache.swift
DePasqualeOrg: This mirrors `maybe_quantize_kv_cache` in mlx-lm. If `kvBits` is specified, a `QuantizedKVCache` is created at the first step.

Libraries/MLXLMCommon/KVCache.swift
DePasqualeOrg: This mirrored the behavior in Python, but I think you're right that it makes sense to consistently return it.

Libraries/MLXLMCommon/KVCache.swift
DePasqualeOrg: I think I've improved this in the commit that I'll add in a moment. Please check it, since I'm not familiar with the usage.

Libraries/MLXLMCommon/KVCache.swift
```diff
@@ -130,13 +250,1144 @@ public class KVCacheSimple: KVCache, Evaluatable, CustomDebugStringConvertible {
...
         "\(String(describing: Self.self)) \(Unmanaged.passUnretained(self).toOpaque()), offset: \(offset), step: \(step), keys: \(keys?.shape.description ?? "-"), values: \(values?.shape.description ?? "-")"
     }
 }
+
+/// Rotating KV cache for sliding window attention
+public class RotatingKVCache: BaseKVCache, CustomDebugStringConvertible {
+    private var keep: Int
+    private var keys: MLXArray?
+    private var values: MLXArray?
+    // TODO: `offset` from the Python implementation is not implemented here. Do we need it?
```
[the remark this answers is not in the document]
DePasqualeOrg: That's right. I was confused.

davidkoski (approved):
Looks good, ran integration tests and all was well.  Thank you!

PR #336, opened 2025-06-21
Add Falcon H1

johnmai-dev:
This PR ports Falcon H1. 

Thanks to my idol @Blaizzy !
Source: https://github.com/ml-explore/mlx-lm/pull/231

image
Xcode 2025-10-01 13 23 44

### DIFF
--- a/Libraries/MLXLLM/Models/FalconH1.swift
+++ b/Libraries/MLXLLM/Models/FalconH1.swift
@@ -1,0 +1,1057 @@
+//
+//  FalconH1.swift
+//  mlx-swift-examples
+//
+//  Created by John Mai on 2025/6/18.
+//
+
+import Foundation
+import MLX
+import MLXLMCommon
+import MLXNN
+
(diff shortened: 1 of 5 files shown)


Libraries/MLXLLM/Models/FalconH1.swift
```diff
@@ -0,0 +1,1080 @@
...
+            try container.decodeIfPresent(Bool.self, forKey: .tieWordEmbeddings) ?? false
+        self.torchDtype =
+            try container.decodeIfPresent(String.self, forKey: .torchDtype) ?? "bfloat16"
+        self.vocabSize = try container.decodeIfPresent(Int.self, forKey: .vocabSize) ?? 128000
+    }
+}
+
+// MARK: - Mamba2Cache KVCache
+
+private class Mamba2Cache: KVCache {
```
davidkoski, line 972: There is now a:

```swift
/// Simple cache for Mamba-style state space models
public class MambaCache: BaseKVCache {
```

added in #334 -- will that work for this?  If not, would it make sense to move Mamba2Cache into KVCache.swift?

PR #340, opened 2025-06-27
Add Gemma 3n

DePasqualeOrg:
Edit: I can quantize the model using mlx-vlm, so I will be able to test this after we fix the sanitization.

### DIFF
--- a/Libraries/MLXVLM/CheckArrayShape.swift
+++ b/Libraries/MLXVLM/CheckArrayShape.swift
@@ -1,0 +1,24 @@
+import MLX
+
+/// Check if array is in a supported format for conv weights
+public func checkArrayShape(_ arr: MLXArray) -> Bool {
+    let shape = arr.shape
+    switch shape.count {
+    case 4:
+        let outChannels = shape[0]
+        let kH = shape[1]
+        let kW = shape[2]
+        // shape[3] is in_channels, which is ignored
+        // Check if out_channels is the largest, and kH and kW are the same
+        return (outChannels >= kH) && (outChannels >= kW) && (kH == kW)
+    case 3:
+        let kW = shape[1]
+        let outChannels = shape[2]
+        // shape[0] is ignored
+        // Check if kW is larger than or equal to out_channels
+        return kW >= outChannels
+    default:
+        // Any other number of dimensions is not supported
+        return false
+    }
+}
--- a/Libraries/MLXVLM/Models/Gemma3.swift
+++ b/Libraries/MLXVLM/Models/Gemma3.swift
@@ -756,19 +756,6 @@ private class VisionModel: Module {
         visionModel(x, outputHiddenStates: outputHiddenStates)
     }
 
-    /// Check if array is already in MLX format for conv2d weights
-    private func checkArrayShape(_ arr: MLXArray) -> Bool {
-        let shape = arr.shape
-
-        // Check if the shape has 4 dimensions
-        guard shape.count == 4 else { return false }
-
-        let (outChannels, kH, kW, _) = (shape[0], shape[1], shape[2], shape[3])
-
-        // Check if out_channels is the largest, and kH and kW are the same
-        return (outChannels >= kH) && (outChannels >= kW) && (kH == kW)
-    }
-
     func sanitize(weights: [String: MLXArray]) -> [String: MLXArray] {
         var sanitizedWeights = [String: MLXArray]()
 
--- a/Libraries/MLXVLM/Models/Gemma3n.swift
+++ b/Libraries/MLXVLM/Models/Gemma3n.swift
@@ -386,79 +386,50 @@ public struct ModelConfig: Codable, Sendable {
 
 // MARK: - Language Model Components
 
-// Base protocol for RMSNorm variants
-private protocol Gemma3nRMSNormProtocol: UnaryLayer {
-    func callAsFunction(_ x: MLXArray) -> MLXArray
-}
-
-// RMSNorm with scale parameter
-private class Gemma3nRMSNormWithScale: Module, Gemma3nRMSNormProtocol {
+private class Gemma3nRMSNorm: Module {
     let eps: Float
     let scaleShift: Float
-    @ModuleInfo var weight: MLXArray
+    @ModuleInfo var weight: MLXArray?
 
-    init(dim: Int, eps: Float = 1e-6, scaleShift: Float = 0.0) {
+    init(dim: Int, eps: Float = 1e-6, scaleShift: Float = 0, withScale: Bool = true) {
         self.eps = eps
         self.scaleShift = scaleShift
-        self._weight.wrappedValue = MLXArray.ones([dim])
+
+        if withScale {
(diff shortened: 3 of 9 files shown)


DePasqualeOrg:
I have been able to test the initial loading of the model. I think the configs are now being created correctly, but I'm running into a lot of problems related to the sanitization of the weights.

I'm going to leave it here for now and wait for experts to add their input. I can't do this alone.

davidkoski:
> Error: Key originalInvFreq not found in Gemma3nRotaryEmbedding

This is the loading code confusing model weights with computed weights.  The pattern (in mlx python as well) is to name the computed parameters with a leading underscore:

```swift
    let _invFreq: MLXArray
    let _originalInvFreq: MLXArray
```

this matches the python code:

```python
        self._inv_freq = mx.array(inv_freq, dtype=mx.float32)
        self._original_inv_freq = mx.array(inv_freq, dtype=mx.float32)
```

davidkoski:
Working on an update, stand by

davidkoski:
OK, I think I fixed these key issues:

- computed values need to be named with a leading underscore
- some of the keys on the modules were missing

I found this in the python code:

```python
        self.query = NamedSequential()
        self.query.add_module(
            "proj",
            create_conv2d(
                dim,
                self.num_heads * self.key_dim,
                kernel_size=1,
            ),
        )
```

corresponding to:

```swift
private class MultiQueryAttention2d: Module {

    @ModuleInfo var queryProj: Conv2d
```

but I am not sure what that does (yet).  The weights didn't have these values (it must be optional).

The current failure is:

```
Error: Mismatched parameter weight shape. Actual [1536, 1024], expected [1536, 640]
```

xlab:
```
    // MLX Swift currently doesn't have custom Metal kernel creation capabilities like Python's
    // mx.fast.metal_kernel(). Consider optimizing with vectorized MLX operations or requesting
    // custom kernel support from the MLX Swift team for better performance.
```

Wait, isn't it supported via `MLXFast`? https://github.com/ml-explore/mlx-swift/blob/b79c74ce773440b86a81ef925ea78dd5023a16c0/Source/MLXFast/MLXFastKernel.swift#L29

**Example**
https://github.com/ml-explore/mlx-swift/blob/b79c74ce773440b86a81ef925ea78dd5023a16c0/Tests/MLXTests/MLXFastKernelTests.swift#L38-L51

DePasqualeOrg:
I added some complicated fixes for sanitization, which have resolved some but not all of the problems. These are probably not the right solution, since the sanitization in Python is much simpler. But the original logic that closely followed the Python implementation wasn't working.

Currently it's failing when it tries to assign to `blocks` in `VisionTower`.

DePasqualeOrg:
@Blaizzy, can you help me understand what's going wrong with the weights sanitization here? Is this something that should already be handled during the conversion of the model to MLX format? I'm using a 4-bit quantized model that I converted with mlx-vlm.

davidkoski:
@DePasqualeOrg see https://github.com/Blaizzy/mlx-vlm/issues/400 and https://github.com/Blaizzy/mlx-vlm/pull/398

I wanted to see how the `blocks.blocks` were loaded so I tried this too but I can't load the model:

```
python -m mlx_vlm.generate --model mlx-community/gemma-3n-E4B-bf16 --prompt "describe these images in english" --image /Users/dkoski/Desktop/IMG_0691.jpeg

...

  File "/Users/dkoski/miniconda3/envs/mlx/lib/python3.11/site-packages/mlx/nn/layers/base.py", line 178, in load_weights
    raise ValueError(f"Received parameters not in model: {extras}.")
ValueError: Received parameters not in model: language_model.lm_head.weight.
```

Sure enough, there is no `lm_head` property in the model (I am building from source, ebafa5a789ed1a8e050b8366ae4e845dbe640b90)

Are you able to successfully load a model?  If so, which one?

davidkoski:
OK, next issue is:

```swift
@ModuleInfo var weight: MLXArray?
```

should be:

```swift
@ParameterInfo var weight: MLXArray?
```

or simply omit the ParameterInfo since we don't need to override the key.

After that the `NamedSequential` in `MultiQueryAttention2d` handles `output.proj.weight` where it would normally be `output_proj.weight` -- this needs to be added (I had called this out earlier as a bit of a mystery as I couldn't yet see how it worked).

Anyway, this is a little bit tricky to implement as it does this in python:

```python
    def add_module(self, name, module):
        setattr(self, name, module)
```

so it dynamically adds properties that will later be picked up by `Module`.  We can do something like that by overriding `items()`:

```swift
private class NamedSequential: Module, UnaryLayer {
    
    var _items = ModuleItems()
    var _names = [String]()
    
    override init() {
    }
    
    init(_ name: String, _ module: UnaryLayer) {
        super.init()
        self.add(name, module)
    }
    
    func add(_ name: String, _ module: UnaryLayer) {
        _items[name] = .value(.module(module))
        _names.append(name)
    }
    
    override func items() -> ModuleItems {
        _items
    }
    
    func callAsFunction(_ x: MLXArray) -> MLXArray {
        var x = x
        for name in names {
            guard let module = items[name]?.module else {
                fatalError("Cannot find \(name) in items")
            }
            x = module(x)
        }
        return x
    }
}
```

`items()` is the cache of the introspected values -- we are going to supply our own custom build items.

davidkoski:
But, seeing how it is actually used, I think this is simpler and more clear:

```swift
private class ProjectionBlock: Module, UnaryLayer {
    @ModuleInfo(key: "down_conv") var down: Conv2d?
    @ModuleInfo var norm: RMSNormAct2d?
    @ModuleInfo var proj: Conv2d
}
```

davidkoski:
Working through some more issues:

- mlx-swift, conv2d and groups was missing a divide by the groups
- this produces a slice `let imgShape = img.shape.suffix(2)` -- the array indices are not `0, 1`
- replaced the interpolation with Upsample
- some config confusion -- vocabSize from the right config

I commented out (with a comment to delete when ready) some of the code in the vision tower sanitize -- the code now matches the python code, I think.

It loads the model and starts to run but fails with:

```
Shapes (1,682,16,48) and (1,2048,1,1) cannot be broadcast
```

in `rmsNorm2d()`.

Getting close, I think.  This is with `mlx-community/gemma-3n-E2B-it-bf16`.

davidkoski:
BTW, you may need to update your `mlx-swift` dependency -- once we are ready here I will tag that, but for now there were a couple changes there.

DePasqualeOrg:
Thank you! Did you try loading the quantized model? I won't be able to test this at all until the quantized model can load.

Blaizzy:
It's stable @DePasqualeOrg 👌🏽

v0.3.0 fixes all Gemma3n bugs

xlab:
@DePasqualeOrg what's the new plan?

DePasqualeOrg:
I'm going to leave model porting to others and focus on other areas where I can add value, such as the recent refactoring that allows for much faster model loading and tokenization.

Porting models—even complex ones like this—is now tractable for anyone using the latest coding agents. Also, I imagine this model is now superseded by Gemma 4.

PR #343, opened 2025-06-27
load gemma3 dwq VLM models as LLM models

### LINKED ISSUE #339: Loading error Gemma 3 4bit-DWQ models
Congrats everyone for adding Gemma 3 support! 

I am running into an issue using the 4bit DWQ quantized version of the 4B, 12B and 27B versions. The 4bit DWQ quantized 1B (non-vision) model works fine.

I'm getting error `Error loading mlx-community/gemma-3-12b-it-4bit-DWQ: Failed to parse config.json for model 'mlx-community/gemma-3-12b-it-4bit-DWQ': Missing field 'vision_config'` when trying to load the model in memory.

I can see that each of these models have been download about 1,000 times each from HF. That suggests that the Python LLM version works fine (assuming there's no bystander effect here).

Could this be an issue in the Swift implementation?

davidkoski:
Hopefully fixed #339.  I tested with `--model mlx-community/gemma-3-4b-it-4bit-DWQ`

The DWQ weights just a VLM with the vision tower removed.  The weights and config match the VLM.

The other option is to make the VLM side load a vision model with no vision tower.  I think that is also doable and potentially the input processor could throw an error if the user tried to add an image.  But maybe that is confusing?  Well, _more_ confusing?

@DePasqualeOrg & @ronaldmannak

### DIFF
--- a/Libraries/MLXLLM/Models/Gemma3Text.swift
+++ b/Libraries/MLXLLM/Models/Gemma3Text.swift
@@ -56,7 +56,7 @@ public struct Gemma3TextConfiguration: Codable {
     public init(from decoder: Decoder) throws {
         let nestedContainer = try decoder.container(keyedBy: VLMCodingKeys.self)
 
-        // in the case of VLM models convertered using mlx_lm.convert 
+        // in the case of VLM models convertered using mlx_lm.convert
         // the configuration will still match the VLMs and be under text_config
         let container =
             if nestedContainer.contains(.textConfig) {


ronaldmannak:
This absolutely works for me. Simple and easy. Thanks David!

sxy-trans-n:
>  I think that is also doable and potentially the input processor could throw an error if the user tried to add an image. But maybe that is confusing? Well, more confusing?

more confusing

PR #351, opened 2025-07-07
Fix thread safety issues in MLX concurrent inference (Samplers + TokenIterator)

sxy-trans-n:
## 🐛 Problem

The MLX Swift Examples library suffers from multiple thread safety issues when used in concurrent inference scenarios. The issues manifest at two levels:

1. **Sampler Level**: `CategoricalSampler` and `TopPSampler` race on `MLXRandom.globalState`
2. **Evaluation Level**: `TokenIterator` instances race on MLX's internal evaluation engine

### Error Symptoms
- **Crash**: `[eval] Attempting to eval an array without a primitive`
- **Inconsistent results**: Different outputs for identical inputs
- **Memory corruption**: Occasional segmentation faults during concurrent sampling
- **Evaluation errors**: MLX internal assertions when multiple `asyncEval()` calls overlap

#### 1. Sampler Race Conditions
Both samplers use compiled functions that implicitly access the global random state:
```swift
compile(inputs: [MLXRandom.globalState], outputs: [MLXRandom.globalState]) { ... }
```

#### 2. MLX Evaluation Race Conditions
Multiple `TokenIterator` instances calling `asyncEval()` concurrently cause race conditions in MLX's internal evaluation engine, even when samplers are properly synchronized.

When multiple threads call these operations concurrently, they race to access and modify shared state, causing undefined behavior.

## 🔧 Solution

Added comprehensive thread safety at both the sampler and evaluation levels using `NSLock` to serialize access to shared MLX resources.

### Key Changes

1. **CategoricalSampler**: Added `randomStateLock` to protect global random state access
2. **TopPSampler**: Added `randomStateLock` to protect global random state access  
3. **TokenIterator**: Added `mlxEvalLock` to serialize MLX evaluation operations (`asyncEval`, `model.prepare`)

### Design Decisions

- **Minimal code changes**: No API breaking changes, maintains backward compatibility
- **Static locks per component**: Minimize memory overhead while ensuring proper synchronization
- **Focused critical sections**: Lock only the essential operations to minimize contention

## 📊 Performance Impact

**Sampler Level**: Minimal impact (~0.001% overhead) - sampling is <1% of total inference time
**TokenIterator Level**: Moderate impact - model evaluations are serialized, but throughput remains 3-4x better than pure serial processing

**Observed Pattern**: 10 concurrent requests complete in 3-4 batches rather than full parallelism, but with 100% stability.

### DIFF
--- a/Libraries/MLXLMCommon/Evaluate.swift
+++ b/Libraries/MLXLMCommon/Evaluate.swift
@@ -134,7 +134,7 @@ public struct ArgMaxSampler: LogitSampler {
 /// Sampler that uses `topP` and `temperature` to sample the logits.
 public struct TopPSampler: LogitSampler {
     private static let randomStateLock = NSLock()
-    
+
     let temp: MLXArray
     let topP: MLXArray
 
@@ -167,11 +167,11 @@ public struct TopPSampler: LogitSampler {
         if logits.dtype == .bfloat16 {
             logits = logits.asType(.float32)
         }
-        
+
         // Thread-safe sampling to prevent concurrent access to global random state
         TopPSampler.randomStateLock.lock()
         defer { TopPSampler.randomStateLock.unlock() }
-        
+
         return compiledTopPSampling(logits, topP, temp)
     }
 }
@@ -179,7 +179,7 @@ public struct TopPSampler: LogitSampler {
 /// Processor that uses `temperature` to sample the logits
 public struct CategoricalSampler: LogitSampler {
     let temp: MLXArray
-    
+
     // Thread-safe sampling using a lock to protect global state access
     private static let randomStateLock = NSLock()
 
@@ -281,7 +281,7 @@ public struct RepetitionContext: LogitProcessor {
 public struct TokenIterator: Sequence, IteratorProtocol {
     // Global lock to protect MLX evaluation operations
     private static let mlxEvalLock = NSLock()
-    
+
     let model: any LanguageModel
     var state: LMOutput.State?
 
--- a/Libraries/MLXLMCommon/Evaluate.swift
+++ b/Libraries/MLXLMCommon/Evaluate.swift
@@ -279,6 +279,9 @@ public struct RepetitionContext: LogitProcessor {
 ///
 /// Note: this uses `asyncEval()` and there may be an async evaluation running after a call to `next()`.
 public struct TokenIterator: Sequence, IteratorProtocol {
+    // Global lock to protect MLX evaluation operations
+    private static let mlxEvalLock = NSLock()
+    
     let model: any LanguageModel
     var state: LMOutput.State?
 
@@ -393,13 +396,17 @@ public struct TokenIterator: Sequence, IteratorProtocol {
             y = tokens
 
             // evaluate the remainder of the prompt -- this primes the pump
+            TokenIterator.mlxEvalLock.lock()
             let token = step(previous: y)
             y = .init(tokens: token)
             asyncEval(y.tokens)
+            TokenIterator.mlxEvalLock.unlock()
 
         case .logits(let result):
+            TokenIterator.mlxEvalLock.lock()
             y = .init(tokens: convertToToken(logits: result.logits))
             asyncEval(y.tokens)
+            TokenIterator.mlxEvalLock.unlock()
 
             break
         }
@@ -444,9 +451,11 @@ public struct TokenIterator: Sequence, IteratorProtocol {
         let previousY = y
 
         // compute the next state and async eval the next token
+        TokenIterator.mlxEvalLock.lock()
         let token = step(previous: previousY)
         y = .init(tokens: token)
         asyncEval(token)
+        TokenIterator.mlxEvalLock.unlock()
 
         tokenCount += 1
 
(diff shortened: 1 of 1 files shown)


Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -133,6 +133,8 @@ public struct ArgMaxSampler: LogitSampler {
 
 /// Sampler that uses `topP` and `temperature` to sample the logits.
 public struct TopPSampler: LogitSampler {
+    private static let randomStateLock = NSLock()
```
davidkoski, line 136: This won't protect it -- randomState is global and this is only protecting callers of TopPSampler.  For example it will not guard against concurrent use in CategoricalSampler.

The better way to fix this would be to have random state scoped to the sampler itself, see:

- https://swiftpackageindex.com/ml-explore/mlx-swift/main/documentation/mlx/withrandomstate(_:body:)-18ob4
- https://github.com/ml-explore/mlx-swift/blob/main/Tests/MLXTests/MLXRandomTests.swift#L237

To use locks, _all_ callers of Random would have to use the same lock.  Actually it is more complicated than that -- the calls to `globalState` are themselves thread safe:

- https://swiftpackageindex.com/ml-explore/mlx-swift/main/documentation/mlx/mlxrandom/randomstate

but the calls to evaluate the resulting MLXArrays are not -- you need to guard the eval sites.

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -166,6 +168,10 @@ public struct TopPSampler: LogitSampler {
             logits = logits.asType(.float32)
         }
 
+        // Thread-safe sampling to prevent concurrent access to global random state
+        TopPSampler.randomStateLock.lock()
+        defer { TopPSampler.randomStateLock.unlock() }
```
davidkoski, line 173: FWIW the typical way to use a lock like this is:

```swift
lock.withLock {
    compiledTopPSampling(...)
}
```

but see my other comment on the use of locks to guard this

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -267,6 +279,9 @@ public struct RepetitionContext: LogitProcessor {
 ///
 /// Note: this uses `asyncEval()` and there may be an async evaluation running after a call to `next()`.
 public struct TokenIterator: Sequence, IteratorProtocol {
+    // Global lock to protect MLX evaluation operations
+    private static let mlxEvalLock = NSLock()
```
davidkoski, line 283: See:

- https://github.com/ml-explore/mlx-swift/blob/main/Source/MLX/Transforms%2BEval.swift#L9

This guards only concurrent calls in `TokenIterator`.  In theory calls to `eval()` and `asyncEval()` should be thread safe as long as callers are using entirely distinct MLXArrays / compute graphs.  In practice, that was never really a guarantee from `mlx::core` and in mlx-swift 0.25.1 we found new issues around this (changes on the core side).

The `evalLock` in `mlx-swift` is wider than just eval -- it has to guard a number of calls.  It may be removed sometime in the future if we can restore the thread safe behavior in `mlx::core`.

Anyway, that said, this does not guard against concurrent use of the same model (which would still be a problem), nor does it add anything over the lock already in mlx-swift (that I can see, anyway).

sxy-trans-n:
@davidkoski 

Thank you for the detailed explanation! This clarifies why our locks are insufficient.

## Our Use Case
We're running Swama as an OpenAI HTTP server where multiple concurrent requests hit the same model instance. We currently serialize all model access to avoid crashes, but want to enable parallelism for better throughput.

## The Problem
We hit `[eval] Attempting to eval an array without a primitive` crashes in two scenarios:
1. **Non-deterministic sampling** (temp > 0): `CategoricalSampler`/`TopPSampler` concurrently access `MLXRandom.globalState`
2. **Even deterministic sampling** (temp = 0): Multiple `ArgMaxSampler` requests cause concurrent `asyncEval()` calls to interfere

## Questions
1. Should we refactor samplers to use `withRandomState` instead of `compile(inputs: [MLXRandom.globalState], ...)`?
2. For our HTTP server use case, what's the recommended approach - separate model instances per request or something else?
3. Are the `asyncEval()` issues with `ArgMaxSampler` expected to be handled by MLX-Swift's existing `evalLock`?

## Next Steps
If `withRandomState` is the right direction, we'd be happy to implement and test the fix in our fork, then submit a PR. We have a stress test that consistently reproduces the issues if that would be helpful.

We appreciate your guidance on the proper solution!

Libraries/MLXLMCommon/Evaluate.swift
```diff
@@ -141,9 +141,13 @@ public struct TopPSampler: LogitSampler {
...
-    private let compiledTopPSampling: (MLXArray, MLXArray, MLXArray) -> MLXArray = {
-        compile(inputs: [MLXRandom.globalState], outputs: [MLXRandom.globalState]) {
-            logits, topP, temp in
+    public func sample(logits: MLXArray) -> MLXArray {
+        var logits = logits
+        if logits.dtype == .bfloat16 {
+            logits = logits.asType(.float32)
+        }
+
+        return withRandomState(MLXRandom.RandomState()) {
```
davidkoski, line 150: I wonder if we can make a new property of the sampler, e.g.

```swift
let randomState = MLXRandom.RandomState()
```

and then use a similar `compiledTopPSampling` that also takes in the `randomState` as a parameter.

As this is it creates a new RandomState for every call to sample it.  This should work, but might be more costly than simply holding the state in the sampler (the sampler should be created for every call to the iterator).

davidkoski (approved):
Looks great, thank you for implementing this!

PR #363, opened 2025-07-22
Remove duplicate keys

tattn:
I found duplicate keys, so I removed the ones with identical key-value pairs.

PR #365, opened 2025-07-22
Fix crash: "Fatal error: Dictionary literal contains duplicate keys"

DePasqualeOrg:


DePasqualeOrg:
Closing because https://github.com/ml-explore/mlx-swift-examples/pull/363 already exists and found another duplicate entry.

PR #369, opened 2025-07-30
Fix LFM2

DePasqualeOrg:
I fixed the issues with the Swift port of LFM2 and aligned it closely with the Python implementation.

@johnmai-dev, since I've made some of these same mistakes in the past when using agentic tools, I would like to make the following suggestions:

- I always create a new folder in the base of the repo called `reference` and clone the Python mlx-lm repo into it, so that the agent has access to it.
- After an initial generation, start a fresh session with your tool of choice (Claude Code works well for me) and have the agent check to make sure the Swift implementation is closely aligned with the Python implementation. There are usually mistakes and inconsistencies that need to be fixed.
- Make sure the new Swift port aligns with the conventions used in existing Swift ports. Your initial version used computed properties prefixed with `resolved`, but the existing Swift ports use private optional properties prefixed with an underscore combined with public computed properties for fallback values.
- There's no substitute for manually checking the generated Swift code line by line against the original Python implementation. Inevitably, the agent will overlook some things, and you can prompt it to correct them.

Since this code will be used to train future AI models, aligning the Swift ports closely with the Python reference implementations will help future models learn MLX Swift syntax, which will make our job even easier in the future.

I haven't checked the other ports from @johnmai-dev that were recently merged, but it's possible that they have similar issues. If so, you can use the same techniques to clean them up.

### DIFF
--- a/Libraries/MLXLLM/Models/LFM2.swift
+++ b/Libraries/MLXLLM/Models/LFM2.swift
@@ -11,76 +11,45 @@ import MLXLMCommon
 import MLXNN
 
 public struct LFM2Configuration: Codable, Sendable {
-    var modelType: String = "lfm2"
-    var headDim: Int?
-    var blockFFDim: Int?
-    var vocabularySize: Int = 65536
-    var hiddenSize: Int
-    var intermediateSize: Int?
-    var hiddenLayers: Int
-    var attentionHeads: Int
-    var kvHeads: Int
-    var maxPositionEmbeddings: Int?
-    var normEps: Float
-    var padTokenId: Int?
-    var bosTokenId: Int = 1
-    var eosTokenId: Int = 2
-    var tieWordEmbeddings: Bool = true
-    var convBias: Bool = false
-    var convLCache: Int = 3
-    var blockMultipleOf: Int = 256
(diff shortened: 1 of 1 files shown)


davidkoski:
I think I tested (as in ran a couple prompts through llm-tool) each of these -- is there something that you noticed that was wrong, or was this by inspection of the code?  I see your description of your technique, which is good, but I am wondering if there is a step I am missing when reviewing & merging?

DePasqualeOrg:
The original version ran but produced text that quickly degraded as the generation continued. You can see it in the screenshot in the original PR and compare the texts from both implementations that were recently posted in that PR. These are the key differences that Claude Code identified on my first pass of corrections:

>   1. ShortConv convolution cache logic: The Swift implementation has a significantly different
>   caching approach that doesn't match the Python reference
>   2. MLP dimension parameters: Swift uses args.resolvedIntermediateSize instead of the correct
>   args.block_dim and args.block_ff_dim
>   3. Conv1d padding: Swift uses padding: lCache - 1 but Python doesn't specify padding in the
>   same way
>   4. Configuration parameter mapping: Missing some parameter mappings between Swift
>   configuration and Python ModelArgs

PR #383, opened 2025-09-11
Add lille-130m

smdesai:
This is a port of Gökdeniz Gülmez's Nikity's https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/models/lille-130m.py

lille-130m


davidkoski (approved):
Looks good, thank you!

PR #387, opened 2025-09-12
Add OlmoE

smdesai:
This is a port of OlmoE https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/models/olmoe.py

olmoe


davidkoski:
Perhaps #391 will fix the swift-format issues and this just needs to be rebased once that merges?

PR #391, opened 2025-09-16
update for swift-format 602

davidkoski:
swift-format recently upgraded to 602 and has slightly different formatting which fails CI


davidkoski:
Obsolete with merge of #387

PR #399, opened 2025-09-26
improve swift-format checks -- same as mlx-swift

davidkoski:
This will give better messaging and context when swift-format fails.  I made the same change in mlx-swift when debugging it recently.


angeloskath (approved):
👍

PR #316, opened 2025-05-14
Codable macro

davidkoski:
use ReerCodable macro to allow for default values
 
- make all configuration Sendable and public var as well

PR #401, opened 2025-09-29
add Mamba.swift

nwh:


Libraries/MLXLLM/Models/Mamba.swift
```diff
@@ -0,0 +1,340 @@
...
+
+import Foundation
+import MLX
+import MLXFast
+import MLXLMCommon
+import MLXNN
+
+// port of https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/models/mamba.py
+
+struct StringKey: CodingKey, ExpressibleByStringLiteral {
```
davidkoski, line 11: You don't need this, see below.

Libraries/MLXLLM/Models/Mamba.swift
```diff
@@ -0,0 +1,340 @@
...
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        let fallback = try decoder.container(keyedBy: StringKey.self)
+
+        modelType = try container.decode(String.self, forKey: .modelType)
+        vocabSize = try container.decode(Int.self, forKey: .vocabSize)
+        hiddenSize =
+            try container
+            .decodeIfPresent(Int.self, forKey: .hiddenSize)
+            ?? fallback
+            .decode(Int.self, forKey: "d_model")
```
davidkoski, line 62: You can do this with:

```swift
    enum CodingKeys: String, CodingKey {
        case modelType = "model_type"
        case vocabSize = "vocab_size"
        case hiddenSize = "hidden_size"
        case dModel = "d_model"
```

then:

```swift
hiddenSize = try container.decodeIfPresent(Int.self, forKey: .hiddenSize) ?? container.decode(Int.self, forKey: .dModel)
```

#316 will also provide a good solution to this, but it isn't merged yet.

Libraries/MLXLLM/Models/Mamba.swift
```diff
@@ -0,0 +1,340 @@
...
+                processedWeights[key] = value.movedAxis(source: 2, destination: 1)
+            }
+        }
+        return processedWeights
+    }
+
+    public func loraLinearLayers() -> MLXLMCommon.LoRALinearLayers {
+        // TODO ???
+        return []
+    }
```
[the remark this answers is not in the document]
davidkoski: Normally the `q` and `v` projection layers from attention:

```swift
    public func loraLinearLayers() -> LoRALinearLayers {
        model.layers.map { ($0.attention, ["q_proj", "v_proj"]) }
    }
```

but this doesn't seem to have an Attention layer.  It works with any linear layers, so perhaps the `x_proj` and `dt_proj` layers in MambaBlock?

Otherwise maybe just remove the method.  Also, FWIW this type would need to conform to `LoRAModel` for this to work.

Libraries/MLXLLM/Models/Mamba.swift
nwh: I'll just wait for #316 and update the PR using the new macro.

PR #404, opened 2025-10-03
port of granite hybrid moe

smdesai:
This is a port of https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/models/granitemoehybrid.py 

IMG_0018

https://github.com/user-attachments/assets/3e321994-0a7d-419a-a72e-8fe54103e969

### DIFF
--- a/Libraries/Embedders/Configuration.swift
+++ b/Libraries/Embedders/Configuration.swift
@@ -69,6 +69,13 @@ private class ModelTypeRegistry: @unchecked Sendable {
             let model = NomicBertModel(configuration)
             return model
         },
+        "qwen3": {
+            url in
+            let configuration = try JSONDecoder().decode(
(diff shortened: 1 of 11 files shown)


adrgrondin:
Nice job was working on it too! Works well with Granite 4 h and non-h models.

Just a suggestion maybe better to move SSM code to a SSM.swift file like mlx-lm does?

smdesai:
> Nice job was working on it too! Works well with Granite 4 h and non-h models.
> 
> Just a suggestion maybe better to move SSM code to a SSM.swift file like mlx-lm does?

@adrgrondin  Thanks. I'm glad it works well for you.
Yes, moving the code out is a good idea out as it may help with future mamba style ports.

davidkoski (approved):
Looks good, thank you!

PR #408, opened 2025-10-09
Add embedder-tool CLI for document indexing and semantic search

### LINKED ISSUE #407: command line example using Embedders
It would be great to have a command line example using Embedders -- we don't really have a good way to investigate problems with these right now as there is nothing that uses them in the repo.

Perhaps something that:

- loads files from a directory
- tokenizes and computes embeddings
- type a sentence and compute distance, e.g. cosine distance -- print matching documents

Sort of a very simple search/RAG system that used the embeddings.

rudrankriyam:
This PR introduces a new command-line tool, `embedder-tool`, that helps to work with embedding models from the `MLXEmbedders` library. The tool can index documents, semantic search, and for model management.

### index
- Create embedding indexes from document corpora
- Supports multiple file formats (txt, md, etc.)
- Recursive directory scanning
- Batch processing with configurable batch sizes
- Progress reporting during indexing
- Handles binary detection and encoding errors 
- Outputs JSON index files with document paths and embeddings

### search
- Semantic search over indexed documents
- Cosine similarity-based ranking
- Configurable result count
- Dimension mismatch detection and reporting
- Returns ranked results with similarity scores

### **list** 
- List registered embedding model configurations
- Shows all available pre-configured models
- Optional display of local directory models
- Sorted output for easy browsing

### **demo**
- Indexes sample documentation from the repository
- Runs example queries showing semantic search

## Bug Fixes

For NomicBert Model Architecture. Let me know if this should be a separate PR.
- Fixed pooler Linear layer to use `bias: false` matching model weights
- Updated model creation to use `pooler: false` for nomic-embed-text-v1.5

### Index documents

```bash
embedder-tool index --directory docs/ --output index.json --extensions md txt --recursive
```

### Search the index
```bash
embedder-tool search --index index.json --query "How do embeddings work?" --top 5
```

### List available models
```bash
embedder-tool list
```

### Run demo
```bash
embedder-tool demo
```

Closes #407

### DIFF
--- a/Tools/embedder-tool/DemoCommand.swift
+++ b/Tools/embedder-tool/DemoCommand.swift
@@ -6,15 +6,15 @@ struct DemoCommand: AsyncParsableCommand {
         commandName: "demo",
         abstract: "Run a demo using sample repository documentation"
     )
-    
+
     @OptionGroup var memory: MemoryArguments
-    
+
     @Flag(name: .long, help: "Keep the generated demo index file instead of removing it")
     var keepIndex = false
-    
+
     @Argument(help: "Optional queries to run after indexing. Defaults to three sample queries.")
     var queries: [String] = []
-    
+
     mutating func run() async throws {
         var memory = self.memory
         memory.start()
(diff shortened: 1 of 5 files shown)


rudrankriyam:
Back to this PR; let's get it done!

PR #414, opened 2025-10-15
Add Qwen 3 VL Support (Dense)

rudrankriyam:
Added support for Qwen 3 VL. Tested the dense version of 4B and 8B instruct.

Tested on image, video and text only. Need to test for multiple images + also add thinking support

G3SMbvFXIAAxD0A


rudrankriyam:
Almost done! Making the changes in a few minutes. I made the file into separate as it is easier to use with the AI tools (longer files are harder and takes more time to edit)

davidkoski:
> Almost done! Making the changes in a few minutes. I made the file into separate as it is easier to use with the AI tools (longer files are harder and takes more time to edit)

I think that is fine -- the files are still pretty big even split!

PR #420, opened 2025-10-21
Fix Catalyst Build with Swift Transformer 1.1.0

Lakr233:
Related to https://github.com/huggingface/swift-transformers/pull/276

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -28,7 +28,10 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/ml-explore/mlx-swift", .upToNextMinor(from: "0.29.1")),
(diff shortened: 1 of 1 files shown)


Package.swift
```diff
@@ -28,9 +28,7 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/ml-explore/mlx-swift", .upToNextMinor(from: "0.29.1")),
-        .package(
-            url: "https://github.com/huggingface/swift-transformers", .upToNextMinor(from: "1.0.0")
-        ),
+        .package(url: "https://github.com/huggingface/swift-transformers", .upToNextMinor(from: "1.1.0")),
```
davidkoski: This looks fine but I see in the resolved it is 1.1.1 -- should we use that here?

davidkoski (approved):
Looks good, thank you!

Issue #405, opened 2025-10-06
Support for loading LoRA layers from adapter config in Swift


markrenaud:
I’m fine-tuning models with MLX QLoRA using the Python MLX APIs, and then integrating the resulting models into my apps via the Swift framework.

For smaller models (e.g. Gemma3 270M), I often need to train multiple projections to achieve suitable task performance.

In Python, this is straightforward to configure in the QLoRA YAML file, e.g.:

```yaml

...

# LoRA parameters - rank 32 with all projections
lora_parameters:
  # The layer keys to apply LoRA to (all projections)
  keys: ["self_attn.q_proj", "self_attn.v_proj", "self_attn.k_proj", "self_attn.o_proj", "mlp.gate_proj", "mlp.up_proj", "mlp.down_proj"]
  
  # LoRA rank
  rank: 32
  
  # LoRA scale
  scale: 32.0
  
  # Dropout rate for LoRA layers
  dropout: 0.0

...

```

Loading the adapter for inference in Python works as expected with this configuration.

However, in Swift, the default model files (e.g. Gemma3Text) currently only have q_proj and v_proj hard-coded.

This means I need to create a local copy of the model definition and extend it manually, e.g.:

```swift
extension Gemma3TextModel: LoRAModel {
// ORIGINAL
//    public func loraLinearLayers() -> LoRALinearLayers {
//        model.layers.map { ($0.selfAttention, ["q_proj", "v_proj"]) }
//    }
    
    
    // CUSTOMIZED FOR EXTRA LORA LAYERS
    public func loraLinearLayers() -> LoRALinearLayers {
        let attnLayers = model.layers.map { ($0.selfAttention, ["q_proj", "v_proj", "k_proj", "o_proj"]) }
        let mlpLayers = model.layers.map { ($0.mlp, ["gate_proj", "up_proj", "down_proj"]) }
                                            
        return attnLayers + mlpLayers
    }

}
```

Since the generated adapter_config.json already specifies which layers are used, would it be possible for the Swift library to read this information at adapter load time and automatically apply the correct LoRA layers?

This would be especially useful when working with a shared base model while swapping in/out multiple adapters that target different sets of layers.

I am not sure if there is a technical or architectural reason that this hasn’t been implemented in the Swift package .

davidkoski:
Yeah, this looks like a good thing to add.  When LoRA was implemented on the swift side the python side hard coded the layer names, so it is somewhat equivalent.

[...]

PR #424, opened 2025-10-22
Add support LoRA layer keys from adapter config

petrukha-ivan:
# What

Added support for the `keys` property in the adapter config. Previously, these keys were ignored, and hardcoded values were used instead. With this change, `LoRAConfiguration` now correctly applies adapters based on the config-defined keys, and hardcoded keys have been removed from model implementations.

By default, models now target all `Linear` layers when keys are not specified. This aligns with recent changes in the Python package https://github.com/ml-explore/mlx-lm/pull/515. For example, running [the example LoRA training code](https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/LORA.md#fine-tune) (without specifying keys) using the latest Python package will create LoRA adapters for all `self_attn.*_proj` and `mlp.*_proj` layers. In older versions, only `self_attn.q_proj` and `self_attn.v_proj` were targeted.

## Resolves

https://github.com/ml-explore/mlx-swift-examples/issues/405

### DIFF
--- a/Libraries/MLXVLM/Models/FastVLM.swift
+++ b/Libraries/MLXVLM/Models/FastVLM.swift
@@ -1079,8 +1079,8 @@ public class FastVLM: Module, VLMModel, KVCacheDimensionProvider {
         self._multimodalProjector.wrappedValue = FastVLMMultiModalProjector(config)
     }
 
-    public func loraLinearLayers() -> LoRALinearLayers {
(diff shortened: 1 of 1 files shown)


davidkoski:
FastVLM has been merged (#423) since this started -- can you rebase with `main` and convert that one?  Thanks!

petrukha-ivan:
@davidkoski done 👍

davidkoski (approved):
Thank you for this huge effort!

PR #432, opened 2025-10-28
re-port sanitize, fix #431

### LINKED ISSUE #431: Error when loading Qwen3-VL-8B-Thinking and Qwen3-VL-8B-Instruct models
I'm getting this error only when loading the 8B models in this series: `Key language_model.lm_head.weight not found in Qwen3VL.LanguageModel.Linear`

davidkoski:


Libraries/MLXVLM/Models/Qwen3VL.swift
```diff
@@ -1710,24 +1710,16 @@ public final class Qwen3VL: Module, VLMModel, KVCacheDimensionProvider {
...
+                    newKey = newKey.replacingOccurrences(of: "model.visual", with: "vision_tower")
+                } else if newKey.contains("model.language_model") {
+                    newKey = newKey.replacingOccurrences(
+                        of: "model.language_model", with: "language_model.model")
+                }
+            } else if newKey.contains("lm_head") {
                 newKey = newKey.replacingOccurrences(of: "lm_head", with: "language_model.lm_head")
-                guard !config.textConfiguration.tieWordEmbeddings else { continue }
-                adjusted[newKey] = value
-                continue
```
davidkoski: Make the sanitize match the python version, see #431.  It was converting

```
(lldb) po weights.keys.sorted()
▿ 1258 elements
  - 0 : "language_model.lm_head.biases"
  - 1 : "language_model.lm_head.scales"
  - 2 : "language_model.lm_head.weight"
```

to

```
(lldb) po weights.keys.sorted()
▿ 1258 elements
  - 0 : "language_language_model.lm_head.biases"
  - 1 : "language_language_model.lm_head.scales"
  - 2 : "language_language_model.lm_head.weight"
```

This omits the logic of `guard !config.textConfiguration.tieWordEmbeddings else { continue }` (these keys are unused if `tieWordEmbeddings` is present).  If we need that I think it should be re-inserted right before line 1725 (adjusted[newKey] = value)

@DePasqualeOrg @rudrankriyam

PR #433, opened 2025-10-28
add additionalContext in streamlined API, fix #413

### LINKED ISSUE #413: How to set enable_thinking for Qwen models to remove the <think> tags?
I am following the tutorial here:

https://github.com/ml-explore/mlx-swift-examples/tree/main/Libraries/MLXLMCommon

This is my code:

```
import MLXLMCommon
import MLXLLM

Task {
    do {
        print("Loading model...")
                        
        let model = try await loadModel(configuration: LLMRegistry.qwen3_4b_4bit) { p in
            print("Completed: \(p.fractionCompleted)")
        }

        print("Model loaded..")

        let session = ChatSession(model)
        print(try await session.respond(to: "How much protein should I eat to gain muscle? /no_think"))
    } catch {
        print("Error loading model: \(error)")
    }
}
```

Adding the `/no_think ` to my prompt does prevent the thinking but it still includes an empty `<think></think>` tag in the answer. Is there a way to fully disable it?

I am reading below:

https://huggingface.co/Qwen/Qwen3-32B-MLX-4bit

and it says:

>  you can add /think and /no_think to user prompts or system messages to switch the model's thinking mode from turn to turn. 

> For API compatibility, when enable_thinking=True, regardless of whether the user uses /think or /no_think, the model will always output a block wrapped in <think>...</think>. However, the content inside this block may be empty if thinking is disabled. When enable_thinking=False, the soft switches are not valid. Regardless of any /think or /no_think tags input by the user, the model will not generate think content and will not include a <think>...</think> block.

How can I provide the `enable_thinking` to the model? I think I need to modify the chat template somehow?

davidkoski:


awni (approved):
LGTM!

PR #436, opened 2025-10-30
Add LoRATrainingExample to README.md

rudrankriyam:
The `LoRATrainingExample` app exists in the repo for fine-tuning LLMs using LoRA but it was not listed in the main README.

I wonder if it is an oversight when it was was first added?


davidkoski (approved):
Hrm, no idea why it is missing, but thanks for spotting and adding!

PR #438, opened 2025-11-04
Add periodic cache clearing during generation

petrukha-ivan:
# What

When processing large inputs or generating large outputs, memory usage can be enormous. This PR adds periodic cache clearing during prompt processing and token generation. For prompt processing, we clear the cache every partial step, and it depends on the prefill window size. For token generation, we clear the cache every 256 tokens.

### DIFF
--- a/Libraries/MLXLLM/LLMModel.swift
+++ b/Libraries/MLXLLM/LLMModel.swift
@@ -32,6 +32,7 @@ extension LLMModel {
             let result = self(input, cache: cache.isEmpty ? nil : cache, state: state)
             eval(cache)
             y = y[prefillStepSize...]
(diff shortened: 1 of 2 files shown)


Libraries/MLXLLM/LLMModel.swift
```diff
@@ -32,6 +32,7 @@ extension LLMModel {
             let result = self(input, cache: cache.isEmpty ? nil : cache, state: state)
             eval(cache)
             y = y[prefillStepSize...]
+            GPU.clearCache()
```
davidkoski: What about just using the cache limit?  Is that not working in this case?  I would prefer to use policy to keep the cache size under control rather than manual GC calls.

PR #441, opened 2025-11-05
split out mlx-swift-lm

davidkoski:
~Note: this won't build yet until the other repository is live.  This is set up using a local repository.~

Switched

PR #446, opened 2025-11-10
switch to github actions

davidkoski:
@madrob when you get a chance can you please enable actions on this repository?

This sets up GitHub actions and is meant for post #441 -- it doesn't run the unit tests that were part of that split out.

I will either remove the .circleci change from 441 or need to handle the conflict here (trivial either way)


madrob:
@davidkoski actions are enabled now, you might have to close and re-open the PR to get them to run the first time.

.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,56 @@
...
+          submodules: recursive
+     
+      - name: Verify MetalToolchain installed
+        shell: bash
+        run: xcodebuild -showComponent MetalToolchain
+      
+      - name: Setup uv
+        uses: astral-sh/setup-uv@v6
+        with:
+            python-version: ${{ inputs.python-version }}
```
madrob: This input is never defined. Do we use this whole setup just for pre-commit? Would it make sense to split the workflow into a `lint` job that runs precommit (maybe on ubuntu-latest?) and then unblocks a full macOS build?

.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,100 @@
...
+      - name: Clone swift-format
+        shell: sh
+        run: |
+          cd /tmp
+          LATEST_TAG=$(curl -s https://api.github.com/repos/swiftlang/swift-format/releases/latest | \
+            grep '"tag_name":' | \
+            sed -E 's/.*"([^"]+)".*/\1/')
+          echo "swift-format $LATEST_TAG"
+          git clone --branch $LATEST_TAG --depth 1 https://github.com/swiftlang/swift-format.git
+          echo "SWIFT_FORMAT_VERSION=$LATEST_TAG" >> $GITHUB_ENV
```
madrob: Slightly safer to use job outputs for this

.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,100 @@
...
+           
+      - name: Setup uv
+        uses: astral-sh/setup-uv@v6
+        with:
+            activate-environment: true
+     
+      - name: Setup pre-commit
+        shell: sh
+        run: |
+          uv pip install pre-commit
```
madrob: There’s a packaged precommit action, would that be easier?

davidkoski: Ugh, now it hits GLIBC issues :-)

https://github.com/actions/setup-python/issues/1224

It looks like `mlx` uses the action but they are not using a container.  I think I will go back to the `uv` path as I can't get the `runs-on` and `container` to agree on a GLIBC version and the `uv` method worked ok.

.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,100 @@
...
+        run: |
+          cd /tmp
+          LATEST_TAG=$(curl -s https://api.github.com/repos/swiftlang/swift-format/releases/latest | \
+            grep '"tag_name":' | \
+            sed -E 's/.*"([^"]+)".*/\1/')
+          echo "swift-format $LATEST_TAG"
+          git clone --branch $LATEST_TAG --depth 1 https://github.com/swiftlang/swift-format.git
+          echo "SWIFT_FORMAT_VERSION=$LATEST_TAG" >> $GITHUB_ENV
+
+      - name: Cache swift-format build
```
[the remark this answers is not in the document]
madrob: Oh, I misread the cache directory. I think we need an `if: steps.cache-action.outputs.cache-hit != 'true'` on the build step then? Or is the idea that `swift build` will be smart enough on its own?

.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,102 @@
...
+          cd /tmp/swift-format
+          swift build -c release
+
+      - name: Link swift-format to /usr/local/bin
+        shell: sh
+        run: |
+          cd /tmp/swift-format
+          ln -s "$(swift build --show-bin-path -c release)/swift-format" /usr/local/bin/swift-format
+     
+      - name: Configure safe directory for git
```
madrob: What issue is this working around?

davidkoski: `git diff` won't run because the owner of the directory is not the current user:

https://github.com/ml-explore/mlx-swift-examples/actions/runs/19842319335/job/56853222085

```
fatal: detected dubious ownership in repository at '/__w/mlx-swift-examples/mlx-swift-examples'
To add an exception for this directory, call:

	git config --global --add safe.directory /__w/mlx-swift-examples/mlx-swift-examples
```

I have found that the pre-commit message is a bit terse and the diff on the lint check is helpful.

.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,102 @@
...
+
+      - name: Clone swift-format
+        shell: sh
+        run: |
+          cd /tmp
+          LATEST_TAG=$(curl -s https://api.github.com/repos/swiftlang/swift-format/releases/latest | \
+            grep '"tag_name":' | \
+            sed -E 's/.*"([^"]+)".*/\1/')
+          echo "swift-format $LATEST_TAG"
+          git clone --branch $LATEST_TAG --depth 1 https://github.com/swiftlang/swift-format.git
```
madrob: Should we move this line to the build-swift-format step? If there's a cache hit then we don't need to clone the repo at all.

davidkoski: that worked perfectly

madrob (approved):

PR #459, opened 2026-01-05
improve handling of portrait display on iOS

### LINKED ISSUE #456: View Cutoff for LLMEval App on iOS
Half of the LLMEval app is cutoff when viewing in portrait mode on iOS. It seems that the content is shift left so the everything on the left side of the screen is cutoff and there is a gap/padding on the right side of the screen. I tried unchecking all the orientation except for portrait on LLMEval iOS target, but that did not fix the issue. However, when I build and run the app on my mac the view looks fine.

Image

davidkoski:
## Proposed changes

Fix #456 

This isn't perfect, but the UI is usable on iOS portrait:

Screenshot 2026-01-05 at 2 07 13 PM

Specifically on `horizontalSizeClass == .compact` this will use a `DisclosureGroup` for some of the controls and statistics.  It also uses a multi-line layout for the controls (this is specifically what broke the layout).

## Checklist

Put an `x` in the boxes that apply.

- [x] I have read the [CONTRIBUTING](https://github.com/ml-explore/mlx/blob/main/CONTRIBUTING.md) document
- [x] I have run `pre-commit run --all-files` to format my code / installed pre-commit prior to committing changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have updated the necessary documentation (if needed)

### DIFF
--- a/Applications/LLMEval/ViewModels/LLMEvaluator.swift
+++ b/Applications/LLMEval/ViewModels/LLMEvaluator.swift
@@ -79,17 +79,18 @@ class LLMEvaluator {
 
     /// Load and return the model. Can be called multiple times; subsequent calls return the cached model.
     func load() async throws -> ModelContainer {
-        switch loadState {
-        case .idle:
-            return try await performLoad()
+        while true {
+            switch loadState {
+            case .idle:
+                return try await performLoad()
 
-        case .loading:
-            // Already loading, wait and retry
-            try await Task.sleep(nanoseconds: 100_000_000)  // 100ms
-            return try await load()
+            case .loading:
+                // Already loading, wait and retry
+                try await Task.sleep(for: .milliseconds(100))
 
-        case .loaded(let modelContainer):
-            return modelContainer
+            case .loaded(let modelContainer):
+                return modelContainer
+            }
         }
     }
 
--- a/Applications/LLMEval/Views/HeaderView.swift
+++ b/Applications/LLMEval/Views/HeaderView.swift
@@ -6,78 +6,112 @@ struct HeaderView: View {
     @Bindable var llm: LLMEvaluator
     @Binding var selectedDisplayStyle: ContentView.DisplayStyle
 
-    var body: some View {
-        VStack(alignment: .leading, spacing: 12) {
-            // Model info with status
-            HStack {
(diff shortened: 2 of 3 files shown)


davidkoski:
Contributors are welcome to improve in future PRs!

Applications/LLMEval/ViewModels/LLMEvaluator.swift
```diff
@@ -79,17 +79,18 @@ class LLMEvaluator {
...
-            return try await performLoad()
+        while true {
+            switch loadState {
+            case .idle:
+                return try await performLoad()
 
-        case .loading:
-            // Already loading, wait and retry
-            try await Task.sleep(nanoseconds: 100_000_000)  // 100ms
-            return try await load()
```
davidkoski, line 89: Replaced the recursive call with a while loop.

Applications/LLMEval/Views/HeaderView.swift
```diff
@@ -6,78 +6,112 @@ struct HeaderView: View {
...
-        VStack(alignment: .leading, spacing: 12) {
-            // Model info with status
-            HStack {
-                VStack(alignment: .leading, spacing: 4) {
-                    Text("Model")
-                        .font(.caption)
-                        .foregroundStyle(.secondary)
+    @Environment(\.horizontalSizeClass) var horizontalSizeClass
+
+    var status: some View {
```
davidkoski, line 11: Break the view up into a couple pieces that we can reassemble as either a single line (original) or a couple lines (portrait iOS).

Applications/LLMEval/Views/MetricsView.swift
```diff
@@ -15,7 +15,20 @@ struct MetricsView: View {
...
     @State private var showMemoryDetails = false
 
+    @Environment(\.horizontalSizeClass) var horizontalSizeClass
+
     var body: some View {
+        if horizontalSizeClass == .compact {
+            DisclosureGroup("Statistics") {
+                stats
+                    .scaleEffect(0.8)
+            }
```
davidkoski, line 25: Same for the stats view -- it takes up a large potion of the screen on a smaller device.  Scale the text slightly.  The formatting is still poor, but usable.

awni (approved):
Thanks for fixing that!

PR #460, opened 2026-01-10
Better readability with `discardableResult`

alpaycli:
## Improve readability of the code

Added `@discardabeResult` attribute to functions those getting called with `_ =` to silence the compiler. This PR helps to get rid of them.

## Checklist

Put an `x` in the boxes that apply.

- [x] I have read the [CONTRIBUTING](https://github.com/ml-explore/mlx/blob/main/CONTRIBUTING.md) document
- [x] I have run `pre-commit run --all-files` to format my code / installed pre-commit prior to committing changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have updated the necessary documentation (if needed)

### DIFF
--- a/Applications/LLMEval/Views/ContentView.swift
+++ b/Applications/LLMEval/Views/ContentView.swift
@@ -83,7 +83,7 @@ struct ContentView: View {
         .task {
             do {
                 // pre-load the weights on launch to speed up the first generation
-                _ = try await llm.load()
+                try await llm.load()
             } catch {
                 llm.output = "Failed: \(error)"
             }
--- a/Applications/VLMEval/ContentView.swift
+++ b/Applications/VLMEval/ContentView.swift
@@ -250,7 +250,7 @@ struct ContentView: View {
             }
         }
         .task {
-            _ = try? await llm.load()
+            try? await llm.load()
         }
     }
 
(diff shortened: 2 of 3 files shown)


Applications/LLMEval/Views/ContentView.swift
```diff
@@ -83,7 +83,7 @@ struct ContentView: View {
         .task {
             do {
                 // pre-load the weights on launch to speed up the first generation
-                _ = try await llm.load()
+                try await llm.load()
```
davidkoski, line 86: I think the `_ =` is more correct -- we are explicitly discarding the result and are only interested in the side effect.

Applications/VLMEval/ContentView.swift
```diff
@@ -250,7 +250,7 @@ struct ContentView: View {
             }
         }
         .task {
-            _ = try? await llm.load()
+            try? await llm.load()
```
davidkoski, line 253: This example is removed in favor of MLXChatExample

davidkoski:
Thank you for the contribution but see the comment -- I believe that explicitly discarding the result in the case where we are just trying to get the side effect is more explicit and clear.

PR #472, opened 2026-04-30
MLXChatExample: fix VLM image handling on iOS (PhotosPicker, EXIF, empty assistant trim)

john-rocky:
## Summary

Three iOS-only fixes in \`MLXChatExample\` so VLM image input actually works on iPhone:

1. **PhotosPicker instead of \`fileImporter\`** — \`fileImporter\` shows the Files app on iOS, which doesn't expose the photo library. Replace with \`.photosPicker\` for image / video selection. Custom \`Transferable\` wrappers (\`PickedImage\` / \`PickedVideo\`) request the right content type explicitly so loading is reliable.

2. **Normalize EXIF orientation before saving** — \`UIImage.jpegData()\` writes the EXIF orientation tag but leaves pixel data in raw sensor orientation. \`CIImage(contentsOf:)\` (used downstream for the model input) does **not** apply that EXIF tag, so portrait / landscape photos arrive at the VLM rotated. Re-render through \`UIGraphicsImageRenderer\` before writing the JPEG so the file is upright.

3. **Trim trailing empty assistant message before applying chat template** — \`ChatViewModel\` appends an empty assistant message before generation starts (so the UI has a placeholder to stream into). When that vector is passed straight into the chat template, the template closes the assistant turn and the model emits an EOS immediately. Strip the trailing empty assistant entry, matching the implicit behavior of \`ChatSession\`.

macOS path unchanged — \`fileImporter\` still works there. The \`#if os(iOS)\` blocks are scoped to the new code only.

## Test plan

Tested on iPhone (Daisuke の iPhone, iOS 26):

- [x] qwen2.5-VL-3B-Instruct-4bit, photo selected from Photos → image preview shown.
- [ ] Re-test landscape / portrait image after merge: \"describe this image\" output should match the visible orientation.
- [ ] Video selection still working.
- [ ] Successive sends without restarting the chat.

### DIFF
--- a/Applications/MLXChatExample/ChatView.swift
+++ b/Applications/MLXChatExample/ChatView.swift
@@ -7,13 +7,50 @@
 
 import AVFoundation
 import AVKit
+import PhotosUI
 import SwiftUI
+import UniformTypeIdentifiers
+
+#if canImport(UIKit)
+import UIKit
+
+/// Transferable wrapper that explicitly requests image content type from PhotosPicker.
+private struct PickedImage: Transferable {
+    let data: Data
+
+    static var transferRepresentation: some TransferRepresentation {
+        DataRepresentation(importedContentType: .image) { data in
+            PickedImage(data: data)
+        }
+    }
+}
+
+/// Transferable wrapper for video content from PhotosPicker.
+private struct PickedVideo: Transferable {
+    let url: URL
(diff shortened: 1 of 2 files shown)


davidkoski (approved):
Looks good, thank you!

PR #474, opened 2026-05-01
Improve MLXChatExample app

DePasqualeOrg:
In preparation for a subsequent PR demonstrating the model output parsing and Open Responses streaming enabled by [Swift LM Response Parser](https://github.com/DePasqualeOrg/swift-lm-response-parser), I've fixed many issues in the MLXChatExample app. In order to keep the diff relatively small, I'm deliberately not solving every issue and not covering every edge case, since this is an example app and not meant for production. But these changes represent a significant improvement:

- Fixes for various bugs related to the toolbar buttons
- Download sizes are properly formatted, but download progress is still not shown due to a longstanding bug in Swift Hugging Face (this works properly in my fork, [Swift HF API](https://github.com/DePasqualeOrg/swift-hf-api), which is not used here).
- Chats can be continued after switching models.
- Many other quality-of-life improvements

### DIFF
--- a/Applications/MLXChatExample/Services/MLXService.swift
+++ b/Applications/MLXChatExample/Services/MLXService.swift
@@ -21,29 +21,60 @@ class MLXService {
     /// List of available models that can be used for generation.
     /// Includes both language models (LLM) and vision-language models (VLM).
     static let availableModels: [LMModel] = [
-        LMModel(name: "llama3.2:1b", configuration: LLMRegistry.llama3_2_1B_4bit, type: .llm),
-        LMModel(name: "qwen2.5:1.5b", configuration: LLMRegistry.qwen2_5_1_5b, type: .llm),
-        LMModel(name: "smolLM:135m", configuration: LLMRegistry.smolLM_135M_4bit, type: .llm),
-        LMModel(name: "qwen3:0.6b", configuration: LLMRegistry.qwen3_0_6b_4bit, type: .llm),
-        LMModel(name: "qwen3:1.7b", configuration: LLMRegistry.qwen3_1_7b_4bit, type: .llm),
-        LMModel(name: "qwen3:4b", configuration: LLMRegistry.qwen3_4b_4bit, type: .llm),
-        LMModel(name: "qwen3:8b", configuration: LLMRegistry.qwen3_8b_4bit, type: .llm),
         LMModel(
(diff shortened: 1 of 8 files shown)


davidkoski:
> this works properly in my fork, [Swift HF API](https://github.com/DePasqualeOrg/swift-hf-api), which is not used here

It is unfortunate that the examples has to pick only one provider.  Ugh.

Applications/MLXChatExample/Services/MLXService.swift
```diff
@@ -15,120 +15,195 @@ import MLXVLM
...
+    ) async throws -> ChatSession {
+        if let currentSession, currentModel?.id == model.id {
+            return currentSession
         }
+        let container = try await load(model: model)
+        let session = ChatSession(
+            container,
+            instructions: Self.systemPrompt,
+            history: history,
+            generateParameters: generateParameters,
```
davidkoski, line 171: Nice idea!

mlx-swift-examples.xcodeproj/project.pbxproj
```diff
@@ -1274,13 +1274,13 @@
 				INFOPLIST_KEY_UILaunchScreen_Generation = YES;
 				INFOPLIST_KEY_UISupportedInterfaceOrientations_iPad = "UIInterfaceOrientationPortrait UIInterfaceOrientationPortraitUpsideDown UIInterfaceOrientationLandscapeLeft UIInterfaceOrientationLandscapeRight";
 				INFOPLIST_KEY_UISupportedInterfaceOrientations_iPhone = "UIInterfaceOrientationPortrait UIInterfaceOrientationLandscapeLeft UIInterfaceOrientationLandscapeRight";
-				IPHONEOS_DEPLOYMENT_TARGET = 18.0;
+				IPHONEOS_DEPLOYMENT_TARGET = 26.0;
```
davidkoski, line 1277: I don't think it is a problem to update the targets for the examples, but is it required?

DePasqualeOrg: This is due to an issue with Liquid Glass: you need `.sharedBackgroundVisibility(.hidden)` to remove the glass background from indicators in the toolbar. It also allows for using Liquid Glass in the tool call/results view in the subsequent PR.

